Issue #12, opened 2018-04-03
high_entropy_strings plugin fails to capture non-string-like secrets


domanchi:
The [regex for high_entropy_strings](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/high_entropy_strings.py#L19) is too restrictive in the sense that it requires a string to be defined as `'string here'` or `"string here"`.

This means for file formats like `.ini` or `.yaml`, it fails to capture high entropy strings.

Eg.

```
[credentials]
admin_secret = superhighentropystringhere
```

domanchi:
The fix for `.ini` files doesn't capture the allowed format:

```
[section name]
key =
    value1
    value2
```

Reopening issue.

PR #13, opened 2018-04-05
Handling ini files

domanchi:
First part of fix for https://github.com/Yelp/detect-secrets/issues/12.

I currently just assume that config files in this format end with `.ini`. Technically, they can end with anything, but maybe we can just keep adding to `INI_FILE_EXTENSIONS`?


detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -22,6 +32,14 @@ def __init__(self, charset, limit, *args):
         # TODO: Update for not just python comments?
         self.ignore_regex = re.compile(r'# ?pragma: ?whitelist[ -]secret')
 
+    def analyze(self, file, filename):
+        # Heuristically determine whether file is an ini-formatted file.
+        for ext in INI_FILE_EXTENSIONS:
```
KevinHock: We could maybe run `file` on them and see if it comes back `ASCII text`, might get expensive.

KevinHock: I'm kind of ambivalent, right now it's easier to exclude files than include them in the tuple, which makes me lean towards including more by default/making the false-p/false-n rate customizable. Nothing is really set in stone though, we can always change it in the future.

domanchi: Good point. Let me jiggle it up a bit.

PR #16, opened 2018-04-09
Handling yaml files

domanchi:
Part 2/2 for the fix to: https://github.com/Yelp/detect-secrets/issues/12

### DIFF
--- a/detect_secrets/plugins/high_entropy_strings.py
+++ b/detect_secrets/plugins/high_entropy_strings.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import math
+import os
 import re
 import string
 from contextlib import contextmanager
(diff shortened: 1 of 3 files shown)


test_data/config.yaml
```diff
@@ -0,0 +1,9 @@
+credentials:
```
KevinHock: What are your thoughts on badly formatted YAML files? If I'm reading the code right the line-numbers won't be correct but maybe we decide it doesn't matter that much.

e.g.
```yaml
repos: [
    {repo: ..., sha: ..., hooks: [{id: ...}]},
]
```

detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -110,6 +128,44 @@ def _analyze_ini_file(self, file, filename):
 
         return potential_secrets
 
+    def _analyze_yaml_file(self, file, filename):
+        """
+        :returns: same format as super().analyze()
```
KevinHock: Nit: Maybe add e.g. `:raises: yaml.YAMLError` and the same w/ `configparser.Error` to `_analyze_ini_file`.

PR #17, opened 2018-04-13
better handling for ini parsing

domanchi:
Bug fix to https://github.com/Yelp/detect-secrets/issues/12.


detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -189,43 +170,156 @@ def _non_quoted_string_regex(self):
...
+            string.ascii_letters + string.digits + '+/=',
+            limit
+        )
+
+
+class IniFileParser(object):
+
+    def __init__(self, file):
+        self.parser = configparser.ConfigParser()
+        self.parser.optionxform = str
```
KevinHock: This is to keep case, I'm assuming?

domanchi: Yup.

PR #21, opened 2018-06-11
Update pre-commit example config

asottile:


KevinHock:
Thanks again for your contributions @asottile!

PR #27, opened 2018-06-15
Empty values in INI files are not supported

LouisTrezzini:


tests/plugins/high_entropy_strings_test.py
```diff
@@ -152,6 +152,14 @@ def test_yaml_file(self):
                 'Location:    test_data/config.yaml:5',
             )
 
+    def test_entropy_lower_limit(self):
+        with pytest.raises(ValueError):
+            Base64HighEntropyString(-1)
+
+    def test_entropy_upper_limit(self):
+        with pytest.raises(ValueError):
+            Base64HighEntropyString(15)
```
KevinHock: Super nit: We could make this 9

Issue #31, opened 2018-06-16
private_key.py could be considered a derivative work without proper license attribution


asottile:
https://github.com/Yelp/detect-secrets/blob/86e96ee9b90a6dbf3a66930b662bea4980624cee/detect_secrets/plugins/private_key.py#L20

https://github.com/pre-commit/pre-commit-hooks/blob/a193eab99ed99429b9b8e517be68bed9a7f6ec4f/LICENSE#L10-L11

Also noticed a few bugs in that file:

- [it only checks the first line of the file](https://github.com/Yelp/detect-secrets/blob/86e96ee9b90a6dbf3a66930b662bea4980624cee/detect_secrets/plugins/private_key.py#L37)
- [the copied code has drifted from the original source](https://github.com/pre-commit/pre-commit-hooks/blob/a193eab99ed99429b9b8e517be68bed9a7f6ec4f/pre_commit_hooks/detect_private_key.py#L6-L14)

KevinHock:
Hi @asottile, sorry about that, it was an honest mistake on my part and @domanchi's, does https://github.com/Yelp/detect-secrets/pull/35 look okay to you?

domanchi:
Fix merged.

PR #35, opened 2018-06-18
Updating private key plugin

domanchi:
Addressing https://github.com/Yelp/detect-secrets/issues/31, and upgrading the private key plugin.

### DIFF
--- a/detect_secrets/plugins/private_key.py
+++ b/detect_secrets/plugins/private_key.py
@@ -1,3 +1,29 @@
+"""
+This code was extracted in part from
+https://github.com/pre-commit/pre-commit-hooks. Using similar heuristic logic,
+we adapt it to fit our plugin infrastructure, to create an organized,
+concerted effort in detecting all type of secrets in code.
+
+Copyright (c) 2014 pre-commit dev team: Anthony Sottile, Ken Struys
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


detect_secrets/plugins/private_key.py
```diff
@@ -10,14 +36,14 @@
     'BEGIN EC PRIVATE KEY',
     'BEGIN OPENSSH PRIVATE KEY',
     'BEGIN PRIVATE KEY',
+    'PuTTY-User-Key-File-2',
+    'BEGIN SSH2 ENCRYPTED PRIVATE KEY',
 )
 
 
 class PrivateKeyDetector(BasePlugin):
```
KevinHock: GitHub will only let me comment on the diff, however so we don't forget what asottile had brought up, shall we do `s/file.readline/file.read/g` on https://github.com/Yelp/detect-secrets/pull/35/files#diff-e2fb9e2d9b26d21395825207c6fb0cd2R63

domanchi: It's by design that we only check the first line.

https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/private_key.py#L26

asottile: hmm, I'd be concerned about false-negatives here, someone could embed a private key in a string and not get caught?

domanchi: In investigations, looks like this issue is more prominent than I originally thought.

Will update.

Issue #37, opened 2018-06-20
baselines break with version bumps


domanchi:
Currently, baselines have no notion of which version of detect-secrets created it. This makes things slightly annoying, because with a major version bump, it could invalidate old baselines, requiring the user to recreate the baseline to be compliant once again.

At the very least, we should have baselines know which detect-secrets version created it, so we can be aware when this happens.

KevinHock:
0.8.2 are one version, 0.8.8 is a new, breaking version. Due to `detect-secrets` thinking the secrets that have the same hash but different types are different. (https://github.com/Yelp/detect-secrets/pull/26/files was the change you will notice if you update to 0.8.8)

KevinHock:
Note: If we don't update/change the plugin version types in the future, this is relatively preventable IMO

KevinHock:
A related aside: I didn't update the CHANGELOG when I bumped, in a rush. Feel free to not approve my next PR until I do so. ([May add emojis](https://github.com/python-security/pyt/blob/master/CHANGELOG.md)

PR #38, opened 2018-06-21
Adding version in baseline

domanchi:
Fixes https://github.com/Yelp/detect-secrets/issues/37.


detect_secrets/pre_commit_hook.py
```diff
@@ -98,15 +131,45 @@ def raise_exception_if_baseline_file_is_not_up_to_date(filename):
...
+        # would be outdated.
+        raise ValueError
+
+    baseline_version = version.split('.')
+    current_version = VERSION.split('.')
+
+    if int(current_version[0]) > int(baseline_version[0]):
+        raise ValueError
+    elif current_version[0] == baseline_version[0] and \
+            int(current_version[1]) > int(baseline_version[1]):
```
KevinHock: So e.g. we make a baseline on `0.8.8`, and we auto-update `detect-secrets` via `all-repos` and the next time somebody runs `detect-secrets` version `0.9.0` we throw a value error? That seems good.

What does the pre-commit hook say to the user when the error is thrown?

detect_secrets/pre_commit_hook.py
```diff
@@ -129,13 +192,18 @@ def pretty_print_diagnostics(secrets):
...
     """
-    log = CustomLog(formatter='%(message)s').getLogger()
+    log = _get_custom_log()
 
     _print_warning_header(log)
     _print_secrets_found(log, secrets)
     _print_mitigation_suggestions(log)
 
 
+@lru_cache(maxsize=1)
```
KevinHock: ++ Neat 👍

detect_secrets/pre_commit_hook.py
domanchi: Ah whoops. Forgot to add error message. Too excited about that `lru_cache` =)

Issue #32, opened 2018-06-18
pre-commit hook removes plugins from baseline (on version 0.8.8)


LouisTrezzini:
```
$ tox -e pre-commit -- run detect-secrets --all-files
[detect-secrets] Detect secrets..........................................Failed
hookid: detect-secrets

Files were modified by this hook. Additional output:

Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
`git add .secrets.baseline` to fix this.
Your baseline file (.secrets.baseline) is unstaged.
[...]

PR #40, opened 2018-06-22
Fix "pre-commit hook removes plugins from baseline"

KevinHock:
This fixes the issue @LouisTrezzini reported in https://github.com/Yelp/detect-secrets/issues/32 where we would not read in or output plugin information when reading a baseline from the pre-commit hook code.


detect_secrets/core/secrets_collection.py
```diff
@@ -85,9 +86,24 @@ def _load_baseline_from_dict(cls, data):
...
+            'results',
+            'plugins_used'
+        )):
             raise IOError
 
+        result.exclude_regex = data['exclude_regex']
+
+        plugins = []
+        for plugin in data['plugins_used']:
+            plugin_classname = plugin.pop('name')
```
KevinHock: I'm so glad I finally got to use `pop` (cc @omergunal)

detect_secrets/core/secrets_collection.py
```diff
@@ -85,9 +86,24 @@ def _load_baseline_from_dict(cls, data):
         :raises: IOError
         """
         result = SecretsCollection()
-        if 'results' not in data or 'exclude_regex' not in data:
+        if not all(key in data for key in (
+            'exclude_regex',
+            'results',
+            'plugins_used'
```
domanchi: This is gonna be a "breaking" change, because `plugins_used` was only introduced "recently". Then again, as long as we bundle this with https://github.com/Yelp/detect-secrets/pull/38 before wrapping it up as a version, it should be OK.

detect_secrets/core/usage.py
```diff
@@ -210,13 +210,13 @@ def _add_custom_limits(self):
         self.parser.add_argument(
             '--base64-limit',
             type=self._argparse_minmax_type,
-            nargs=1,
+            nargs='?',
```
domanchi: What does it mean to have `--base64-limit` without any limit?

detect_secrets/plugins/__init__.py
```diff
@@ -145,7 +141,7 @@ def _convert_sensitivity_values_to_class_tuple(sensitivity_values):
              This way, we can initialize the class with <plugin_class_name>(<value>)
 
     Example:
-        >>> [ ('HexHighEntropyString', 3), ('PrivateKeyDetector', true), ]
+        >>> [ ('HexHighEntropyString', {'hex_limit': 3}), ('PrivateKeyDetector', {'private_key_detector': true}), ]
```
domanchi: please shorten this line. =)

Maybe,

```
>>> [
...    ('HexHighEntropyString', {'hex_limit': 3}),
...    ('PrivateKeyDetector', {'private_key_detector': true}),
... ]
```

?

To make it easier, I will adjust E501 settings for the future.

detect_secrets/core/usage.py
KevinHock: I'm not sure I follow, do you mean if I do `python -m detect_secrets.main --base64-limit --scan test_secrets/`? It uses the default.

detect_secrets/core/secrets_collection.py
KevinHock: Added the requirement of `version` in my merge with master, which *definitely* breaks (3 months ago plugins_used was added vs. version just added today.) 👍

PR #44, opened 2018-06-26
Add audit baseline functionality

domanchi:
## Testing

```
detect-secrets$ pip install -e .
detect-secrets$ detect-secrets --scan test_data > .secrets.baseline
detect-secrets$ detect-secrets --audit .secrets.baseline
```

### Output

```
Secrets Left: 1/14
Filename:     test_data/baseline.file
----------
4:  "results": {
5:    "file_with_secrets.py": [
6:      {
7:        "type": "High Entropy String",
8:        "line_number": 3,
9:        "hashed_secret": "262547656a76a8a5d24ad428e7010e0fbb8d0413"
10:      }
11:    ],
12:    "sample.diff": [
13:      {
14:        "type": "High Entropy String",
----------
Is this a valid secret? (y)es, (n)o, (s)kip, (q)uit:
```

## Features

* Full colored, interactive auditor for secrets found through `--scan`.
* Ability to flag both false and true positives within the baseline.
* Handles case well where baseline's reported line number falls out of sync with actual code.
* `skip` allows you to revisit the secret at a later time.
* Audit sessions can be "saved" by quitting pre-maturely: secrets already audited will not need to be reviewed again.

## Backwards Compatibility Considerations

**This is a breaking change!**

* Baselines will need to be [created in sorted order](https://github.com/Yelp/detect-secrets/pull/25) as well as [requiring the `plugins_used` list with proper initialization args](https://github.com/Yelp/detect-secrets/pull/40).
* Private Key Detector now creates a hash of the blacklisted line, rather than the whole line itself. This will require all older baselines with this plugin to be rescanned, to get the new hash.
* `SensitivityValues` has been dropped. `detect-secrets-server` will need to be pinned on an older version, until fixed.


detect_secrets/core/audit.py
```diff
@@ -0,0 +1,330 @@
...
+    return '\n'.join(
+        map(
+            lambda x: '{}:{}'.format(
+                BashColor.color(
+                    str(int(x[0]) + start_line),
+                    Color.LIGHT_GREEN,
+                ),
+                x[1],
+            ),
+            enumerate(output),
```
KevinHock: Wow, woah, this is reel nice 🍰 ⚡️ 💡

PR #48, opened 2018-07-03
audit bug fix: handles short files better

domanchi:
Our default `lines_of_context` when printing the interactive auditor is 5. However, when there isn't five lines to print, we can find ourselves in a strange case: the auditor doesn't know how to handle it so it either:

a) throws an IndexError exception (if `abs(-lines_of_context -1 ) > len(output)`), OR
b) displays a `ERROR: Secret not found on specified line number!`

So let's handle this better.

### DIFF
--- a/tests/core/audit_test.py
+++ b/tests/core/audit_test.py
@@ -11,6 +11,7 @@ import pytest
 from detect_secrets.core import audit
 from detect_secrets.core.color import BashColor
 from testing.factories import potential_secret_factory
+from testing.mocks import mock_printer as mock_printer_base
 
 
 class TestAuditBaseline(object):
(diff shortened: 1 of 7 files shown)


tests/core/audit_test.py
```diff
@@ -441,15 +442,7 @@ def test_input_message(self, prompt_secret_decision, expected_output):
 
 @pytest.fixture
 def mock_printer():
```
KevinHock: Super nit: Maybe consider renaming to avoid the name conflict/alias of the `mock_printer` import, e.g. `mock_audit_printer` or something.

tests/main_test.py
```diff
@@ -94,6 +99,83 @@ def test_reads_old_baseline_from_file(self, mock_merge_baseline):
...
+                    5:    0123456789a
+                """)[1:-1],
+            ),
+        ],
+    )
+    def test_audit_first_line(self, filename, expected_output):
+        BashColor.disable_color()
+
+        with mock_stdin(), mock_printer(
+            # To extract the baseline output
```
KevinHock: Great `mock.patch` comments!

PR #51, opened 2018-07-05
Clearer usage commands

domanchi:
Previously, scanning and auditing were command line flags. This meant that *technically*, they could be used in the following manner:

```
$ detect-secrets --audit .secrets.baseline --scan test_data
```

which doesn't really make any sense.

This change clarifies the API a little, so we avoid this situation.

### DIFF
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -3,6 +3,8 @@ from __future__ import absolute_import
 import argparse
 from collections import namedtuple
 
+from detect_secrets import VERSION
+
 
 class ParserBuilder(object):
 
@@ -34,7 +36,8 @@ class ParserBuilder(object):
     def _add_version_argument(self):
         self.parser.add_argument(
             '--version',
-            action='store_true',
+            action='version',
+            version=VERSION,
             help='Display version information.',
         )
         return self
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -10,26 +10,36 @@ class ParserBuilder(object):
 
     def __init__(self):
         self.parser = argparse.ArgumentParser()
-        self.plugins_parser = PluginOptions(self.parser)
 
         self.add_default_arguments()
 
     def add_default_arguments(self):
-        self.plugins_parser.add_arguments()
         self._add_verbosity_argument()\
             ._add_version_argument()
 
+        return self
+
     def add_pre_commit_arguments(self):
-        return self._add_filenames_argument()\
+        self._add_filenames_argument()\
             ._add_set_baseline_argument()
 
+        PluginOptions(self.parser).add_arguments()
+
+        return self
+
     def add_console_use_arguments(self):
-        return self._add_initialize_baseline_argument()\
-            ._add_audit_baseline_argument()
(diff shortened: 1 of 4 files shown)


detect_secrets/core/usage.py
```diff
@@ -194,6 +230,11 @@ def consolidate_args(args):
 
         :param args: output of `argparse.ArgumentParser.parse_args`
         """
+        # Using `--hex-limit` as a canary to identify whether this
```
KevinHock: Nit: This seems a little hacky.

detect_secrets/main.py
```diff
@@ -73,10 +68,10 @@ def _perform_scan(args):
     return new_baseline
 
 
-def _get_existing_baseline(args):
+def _get_existing_baseline(import_filename):
     # Favors --import argument over stdin.
-    if getattr(args, 'import'):
```
KevinHock: ++ Much better :D

domanchi:
```
$ detect-secrets --help
usage: detect-secrets [-h] [-v] [--version] {scan,audit} ...

positional arguments:
  {scan,audit}

optional arguments:
  -h, --help     show this help message and exit
  -v, --verbose  Verbose mode.
  --version      Display version information.
$
$ detect-secrets scan -h
usage: detect-secrets scan [-h] [--exclude EXCLUDE]
                           [--import OLD_BASELINE_FILE] [--string [STRING]]
                           [--base64-limit [BASE64_LIMIT]]
                           [--hex-limit [HEX_LIMIT]] [--no-hex-string-scan]
                           [--no-base64-string-scan] [--no-private-key-scan]
                           [path]

positional arguments:
  path                  Scans the entire codebase and outputs a snapshot of
                        currently identified secrets.

optional arguments:
  -h, --help            show this help message and exit
  --exclude EXCLUDE     Pass in regex to specify ignored paths during
                        initialization scan.
  --import OLD_BASELINE_FILE
                        Import settings from previous existing baseline.

plugins:
  Configure settings for each secret scanning ruleset. By default, all
  plugins are enabled unless explicitly disabled.

  --base64-limit [BASE64_LIMIT]
                        Sets the entropy limit for high entropy strings. Value
                        must be between 0.0 and 8.0.
  --hex-limit [HEX_LIMIT]
                        Sets the entropy limit for high entropy strings. Value
                        must be between 0.0 and 8.0.
  --no-hex-string-scan  Disables scanning for hex high entropy strings
  --no-base64-string-scan
                        Disables scanning for base64 high entropy strings
  --no-private-key-scan
                        Disables scanning for private keys.
$
$ detect-secrets audit --help
usage: detect-secrets audit [-h] filename

positional arguments:
  filename    Audit a given baseline file to distinguish the difference
              between false and true positives.

optional arguments:
  -h, --help  show this help message and exit
```

KevinHock:
Random: What is your opinion on adding that to the README? To much/ugly or helpful? e.g. https://github.com/python-security/pyt#usage

domanchi:
`¯\_(ツ)_/¯`

I've seen it some places; no real preference. I figure, examples to get going are better, and if people want the "advanced" options, they can read `-h` themselves.

The README needs to be updated with this API change, but I'll do that with the version bump.

PR #52, opened 2018-07-05
Adding cli functionality to check strings in an adhoc manner

domanchi:
This long desired functionality should help usage when playing around with the tool.

## Testing

```
$ detect-secrets scan --string 0123456789a
Base64HighEntropyString: False (3.459)
HexHighEntropyString   : True  (3.459)
PrivateKeyDetector     : False
$
$ echo "some secret, but base64 encoded here" | base64 | detect-secrets scan --string
Base64HighEntropyString: True (4.676)
HexHighEntropyString   : False
PrivateKeyDetector     : False
```

### DIFF
--- a/detect_secrets/main.py
+++ b/detect_secrets/main.py
@@ -26,9 +26,21 @@ def main(argv=None):
         log.set_debug_level(args.verbose)
 
     if args.action == 'scan':
+        # Plugins are *always* rescanned with fresh settings, because
+        # we want to get the latest updates.
+        plugins = initialize.from_parser_builder(args.plugins)
+
+        if args.string:
+            line = args.string
+            if isinstance(args.string, bool):
+                line = sys.stdin.read().splitlines()[0]
+
+            _scan_string(line, plugins)
+            return 0
+
         print(
             json.dumps(
-                _perform_scan(args),
+                _perform_scan(args, plugins),
                 indent=2,
                 sort_keys=True,
             ),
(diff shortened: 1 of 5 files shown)


detect_secrets/main.py
```diff
@@ -26,9 +26,21 @@ def main(argv=None):
         log.set_debug_level(args.verbose)
 
     if args.action == 'scan':
+        # Plugins are *always* rescanned with fresh settings, because
+        # we want to get the latest updates.
+        plugins = initialize.from_parser_builder(args.plugins)
+
+        if args.string:
+            line = args.string
```
KevinHock: Super nit: it's not a line if it is a `bool`, maybe `else:`?

detect_secrets/plugins/base.py
```diff
@@ -50,6 +50,29 @@ def secret_generator(self, string):  # pragma: no cover
...
+        PotentialSecret type). e.g. HighEntropyStrings adds their Shannon
+        entropy in which they made their decision.
+
+        :type string: str
+        :param string: the string to analyze
+        :rtype: str
+        :returns: descriptive string that fits the format
+            <classname>: <returned-value>
+        """
+        # TODO: Handle multiple secrets on single line.
```
KevinHock: Nit in general: Maybe github issue instead of `todo`'s in the code.

KevinHock: (To bring attention of short-comings to reviewers, I normally call it out in github PR comment.)

domanchi: https://github.com/Yelp/detect-secrets/issues/53

PR #56, opened 2018-07-10
Fix --audit Crash on Missing File

KevinHock:


detect_secrets/core/audit.py
```diff
@@ -173,6 +175,9 @@ def _secret_generator(baseline):
     )
 
     for filename, secrets in baseline['results'].items():
+        if not os.path.exists(filename):
+            print('{} does not exist.'.format(filename))
```
domanchi: I'm debating the use of this case.

1. `print` statements shouldn't be in the `_secret_generator`.
2. If the file is gone, there's nothing to audit (and we should probably just delete its entry in the updated baseline). After all, we already exclude it from the count (from your change above).

detect_secrets/core/baseline.py
```diff
@@ -160,7 +160,6 @@ def merge_baseline(old_baseline, new_baseline):
     :type new_baseline: dict
     :param new_baseline: most recent scan
     """
-    # TODO: merge_results does not take into account `is_secret`
```
domanchi: 👍

tests/core/audit_test.py
```diff
@@ -24,7 +24,12 @@ def test_no_baseline(self, mock_printer):
             assert mock_printer.message == ''
 
     def test_quit_before_making_decision(self, mock_printer):
-        with self.mock_env(['q']) as m:
+        with mock.patch(
```
domanchi: Let's throw this in `mock_env`. After all, we're already patching a bunch of things there.

detect_secrets/core/audit.py
KevinHock: I was ambivalent too, wasn't sure if we wanted `audit` to only have `is_secret` diffs, but yeah, removing all dead files in `audit_baseline` before calling `_secret_generator` etc. is cleaner, I'll change.

detect_secrets/core/audit.py
```diff
@@ -19,18 +20,30 @@ class SecretNotFoundOnSpecifiedLineError(Exception):
     pass
 
 
+def _clean_baseline_of_nonexistent_files(baseline):
```
domanchi: `_remove_nonexistent_files_from_baseline`? And it should return `True` if modified -- just so it's less negatives, so we don't get a case where:

```
if not baseline_wasnt_modified:
    do_something()
```

nit: let's throw this below `audit_baseline` function declaration, so it reads from top-down.

KevinHock: Hmm, I think I originally made it that but I changed it to `wasnt_modified` was to avoid the `and not baseline_was_modified`, which I read as slightly more confusing. This may be a weird preference of mine though, I'll change it ;)

KevinHock: Hmm `and not files_removed:` looks nice 👍

detect_secrets/core/audit.py
```diff
@@ -66,6 +70,45 @@ def audit_baseline(baseline_filename):
     _save_baseline_to_file(baseline_filename, original_baseline)
 
 
+def _get_baseline_from_file(filename):  # pragma: no cover
```
KevinHock: Raised you one further and re-org'd them based on the order that they are called.

domanchi: <3

PR #57, opened 2018-07-11
adding --all-files flag

domanchi:
## Testing

```
detect-secrets scan --all-files
```

This allows the extended usage of this tool, so we can scan non-git tracked files as well on an adhoc manner.

### DIFF
--- a/detect_secrets/core/baseline.py
+++ b/detect_secrets/core/baseline.py
@@ -7,7 +7,7 @@ import subprocess
 from detect_secrets.core.secrets_collection import SecretsCollection
 
 
-def initialize(plugins, exclude_regex=None, rootdir='.'):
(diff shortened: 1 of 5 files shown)


detect_secrets/core/baseline.py
```diff
@@ -256,3 +258,15 @@ def _get_git_tracked_files(rootdir='.'):
...
     except subprocess.CalledProcessError:
         return None
+
+
+def _get_files_recursively(rootdir):
+    """Sometimes, we want to use this tool with non-git repositories.
+    This function allows us to do so.
+    """
+    output = []
+    for root, dirs, files in os.walk(rootdir):
```
KevinHock: Super nit: you can [use _ for dirs](https://github.com/python-security/pyt/blob/master/pyt/__main__.py#L39)

PR #64, opened 2018-07-31
Reduce sequential string false positives

KevinHock:


detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -17,6 +17,12 @@
 from detect_secrets.plugins.core.yaml_file_parser import YamlFileParser
 
 
+IGNORED_SEQUENTIAL_STRINGS = (
```
domanchi: Let's use the `string` library for this.

e.g.

```
import string

IGNORED_SEQUENTIAL_STRINGS = (
    string.ascii_uppercase + string.digits + string.ascii_uppercase,
    string.hexdigits + string.hexdigits,
)
```

Issue #60, opened 2018-07-13
pre-commit hook removes audited secrets


domanchi:
### Steps to Reproduce

```
$ detect-secrets scan test_data/short_files/first_line.py > .secrets.baseline
$ echo "delete the secret from test_data/short_files/first_line.py"
$ PYTHONPATH=`pwd` detect_secrets/pre_commit_hook.py --baseline .secrets.baseline test_data/short_files/first_line.py
$ git diff .secrets.baseline
```

KevinHock:
s/"delete the secret from test_data/short_files/first_line.py"/"move the secret to a new line in test_data/short_files/first_line.py"/g

You'll see something like

```diff
   "exclude_regex": null,
-  "generated_at": "2018-07-31T17:55:15Z",
+  "generated_at": "2018-07-31T18:06:31Z",
   "plugins_used": [
     {
       "base64_limit": 4.5,
@@ -18,8 +18,7 @@
     "test_data/short_files/first_line.py": [
       {
         "hashed_secret": "4c8e08f94044d9b59913f66873af8eac4299266e",
-        "is_secret": true,
-        "line_number": 1,
+        "line_number": 2,
```

Issue #61, opened 2018-07-18
-v Verbosity overflow, 3+ v's cause a KeyError


KevinHock:
Repro: `detect-secrets --scan . -vvv`

Here lies the relevant stack trace:

```python
Traceback (most recent call last):
  File "/hey/three_six/bin/detect-secrets", line 11, in <module>
    sys.exit(main())
  File "/hey/three_six/lib/python3.6/site-packages/detect_secrets/main.py", line 27, in main
    log.set_debug_level(args.verbose)
  File "/hey/three_six/lib/python3.6/site-packages/detect_secrets/core/log.py", line 46, in _set_debug_level
    self.setLevel(mapping[debug_level])
KeyError: 3
```

KevinHock:
Fixed by https://github.com/Yelp/detect-secrets/pull/65/commits/0b2700251f03d96757349a87fc07425dc0a4973b

PR #65, opened 2018-07-31
Fix "pre-commit hook removes audited secrets" and verbosity overflow

KevinHock:
Fixes https://github.com/Yelp/detect-secrets/issues/60 and https://github.com/Yelp/detect-secrets/issues/61

Aspiring to be more like @bcaller, I've made my git commit history squeaky clean 😁

### DIFF
--- a/detect_secrets/core/log.py
+++ b/detect_secrets/core/log.py
@@ -43,7 +43,9 @@ def _set_debug_level(self, debug_level):
         2: logging.DEBUG,
     }
 
-    self.setLevel(mapping[debug_level])
+    self.setLevel(
+        mapping[min(debug_level, 2)],
+    )
 
 
 log = get_logger()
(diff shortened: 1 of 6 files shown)


KevinHock:
Sorry I was a little [tucked away in my commit message](https://github.com/Yelp/detect-secrets/pull/65/commits/710d96eafe1ce18090a4db75e5488a1b07874940), but 
[`TestPreCommitHook.test_writes_new_baseline_if_modified`](https://github.com/Yelp/detect-secrets/blob/master/tests/pre_commit_hook_test.py#L133-L136) covers the change, since [line 136](https://github.com/Yelp/detect-secrets/blob/master/tests/pre_commit_hook_test.py#L136) modifies the line to be 0, the pre-commit code is given a baseline with an out-of-date lineno and `is_secret=True`, and it outputs the [correct lineno with `is_secret=True`](https://github.com/Yelp/detect-secrets/blob/master/tests/pre_commit_hook_test.py#L152).

KevinHock:
Note to self: what seems like the cleanest way of doing the above is to add a secret to `_create_baseline` w/out the `is_secret` attribute, however I'm a little lazy to do that at the moment, as it may (or may not) require changing a lot of tests.

PR #69, opened 2018-08-09
Only load json from stdin if it exists

guykisel:
I was trying to warp `detect-secrets` in an automated code review tool I'm working on that shells out to various static analysis tools. However, when it runs `detect-secrets`, `detect-secrets` appears to expect JSON on stdin when run in non-interactive mode. This PR adds a small workaround to only load the JSON if stdin input actually exists.

An alternative approach might be a try/except block around the JSON load.

### DIFF
--- a/detect_secrets/main.py
+++ b/detect_secrets/main.py
@@ -82,7 +82,9 @@ def _get_existing_baseline(import_filename):
     if import_filename:
         return _read_from_file(import_filename[0])
     if not sys.stdin.isatty():
-        return json.loads(sys.stdin.read())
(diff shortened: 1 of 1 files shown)


KevinHock:
Feel free to checkout [`pyt`](https://github.com/python-security/pyt) if you like static analysis 😉

guykisel:
Thanks! I will check it out. Currently for Python security linting I run Bandit :)

guykisel:
Nice! I like to keep an eye on https://github.com/mre/awesome-static-analysis and check recent commits occasionally to see if any new tools have showed up.

Issue #63, opened 2018-07-30
Add a (b)ack option to 'Is this a valid secret?'


KevinHock:
User may be spamming `s` `Enter`, and so we will want to go backwards.

KevinHock:
Thanks @cleborys!

PR #72, opened 2018-08-30
Add a (b)ack option to 'Is this a valid secret?' Closes Issue #63

cleborys:
Add going backwards option to 'Is this a valid secret?', closing Issue #63 

This would be my first contribution to open source ever - thank you for the "good first issue" flag!

I have not yet figured out how to run the tests properly, else I would try to be test-driven.

The two directions I could see this go in are currently (preferences welcome):
  1) (the "obvious" one) Make `_secret_generator` in `core/audit.py` a list instead of a generator and handle indices in `audit_baseline`
  2) Cast `_secret_generator` to a "bidirectional iterator" object (still goes through `list`). That adds some overhead, but keeps the `for ... in _secret_generator` of `audit_baseline` and might make `audit.py` more readable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 *.egg-info
 *.py[co]
 *.sw[op]
-.secrets.baseline
 
 /.coverage
 /.pytest_cache
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.egg-info
 *.py[co]
 *.sw[op]
+.secrets.baseline
 
 /.coverage
 /.pytest_cache
--- a/detect_secrets/core/bidirectional_iterator.py
+++ b/detect_secrets/core/bidirectional_iterator.py
@@ -1,25 +1,25 @@
 class BidirectionalIterator(object):
     def __init__(self, collection):
         self.collection = collection
-        self.index = -1 #starts on -1, as index is increased _before_ getting result
-        self.step_back_once = False 
+        self.index = -1  # starts on -1, as index is increased _before_ getting result
+        self.step_back_once = False
 
     def __next__(self):
+        if self.step_back_once:
+            self.index -= 1
+            self.step_back_once = False
+        else:
+            self.index += 1
+        if self.index < 0:
+            raise StopIteration
         try:
-            if self.step_back_once:
-                self.index -= 1
-                self.step_back_once = False 
-                if self.index < 0:
-                    raise StopIteration
-            else:
-                self.index += 1
             result = self.collection[self.index]
         except IndexError:
             raise StopIteration
         return result
 
     def next(self):
-        self.__next__()
+        return self.__next__()
 
     def step_back_on_next_iteration(self):
         self.step_back_once = True
--- a/detect_secrets/core/bidirectional_iterator.py
+++ b/detect_secrets/core/bidirectional_iterator.py
@@ -18,6 +18,9 @@ class BidirectionalIterator(object):
             raise StopIteration
         return result
 
+    def next(self):
+        self.__next__()
+
     def step_back_on_next_iteration(self):
         self.step_back_once = True
 
--- a/detect_secrets/core/bidirectional_iterator.py
+++ b/detect_secrets/core/bidirectional_iterator.py
@@ -1,0 +1,28 @@
+class BidirectionalIterator(object):
+    def __init__(self, collection):
+        self.collection = collection
+        self.index = -1 #starts on -1, as index is increased _before_ getting result
+        self.step_back_once = False 
+
+    def __next__(self):
+        try:
+            if self.step_back_once:
+                self.index -= 1
+                self.step_back_once = False 
+                if self.index < 0:
+                    raise StopIteration
+            else:
+                self.index += 1
+            result = self.collection[self.index]
+        except IndexError:
+            raise StopIteration
+        return result
+
+    def step_back_on_next_iteration(self):
+        self.step_back_once = True
+
+    def can_step_back(self):
+        return self.index > 0
+
+    def __iter__(self):
+        return self
--- a/detect_secrets/core/audit.py
+++ b/detect_secrets/core/audit.py
@@ -14,6 +14,7 @@ from .baseline import merge_results
 from .color import BashColor
 from .color import Color
 from .potential_secret import PotentialSecret
+from .bidirectional_iterator import BidirectionalIterator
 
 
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 *.egg-info
 *.py[co]
 *.sw[op]
+.secrets.baseline
```
cleborys: `.secrets.baseline` felt like it should only be in a local copy. I might be mistaken.

KevinHock: Ahh, that is confusing, sorry about that, it is purposefully meant to be included in the Git repo though.

KevinHock:
> `This would be my first contribution to open source ever`

That's really awesome :D   🎈 🍰 🎉 

I'll get back to you on Tuesday or Wednesday with possible implementation preferences.

domanchi:
Bidirectional iterator looks good! Looking forward to tests!

detect_secrets/core/bidirectional_iterator.py
```diff
@@ -0,0 +1,31 @@
...
+    def __next__(self):
+        try:
+            if self.step_back_once:
+                self.index -= 1
+                self.step_back_once = False 
+                if self.index < 0:
+                    raise StopIteration
+            else:
+                self.index += 1
+            result = self.collection[self.index]
```
KevinHock: Nit: Generally people recommend that only the code that throws the relevant exception should be inside of the `try: except:`, in this case `result = self.collection[self.index]` (This also cuts down on [`Hadouken code`](http://i.imgur.com/BtjZedW.jpg), as there will be 2 indentation levels instead of 3.)

```python
    def __next__(self):
        if self.step_back_once:
            self.index -= 1
            self.step_back_once = False 
            if self.index < 0:
                raise StopIteration
        else:
            self.index += 1

        try:
            result = self.collection[self.index]
        except IndexError:
            raise StopIteration

        return result
```

p.s. 'Nit' means nitpick, i.e. something that doesn't really matter that much.

detect_secrets/core/bidirectional_iterator.py
```diff
@@ -0,0 +1,31 @@
...
+        except IndexError:
+            raise StopIteration
+        return result
+
+    def next(self):
```
KevinHock: Nit: Looping should call the `__next__` method directly, so no need for a `next` method 👍

cleborys: I added `next` to be python2 compatible (it was renamed form `next` to `__next__` from python2 to 3).
However, this does not feel very clean - perhaps you know of a better way? :)

.gitignore
cleborys: Thanks, I re-added it :)

detect_secrets/core/bidirectional_iterator.py
KevinHock: Aha! Very good point, I only tested on Python 3. My bad. I am impressed by how thorough you are 👍

cleborys:
Is there an assumption that secrets in the results of the baseline are in a fixed order? (I mean the lists `baseline['results'][filename]') 
If not I would re-write `merge_results` in `baseline.py` to remove this assumption so that it works with updates to the baseline that are out-of-order.
That would simplify some choices in `audit_baseline` of `audit.py` a lot :)

domanchi:
@cleborys: currently, no there isn't. It's a Python dictionary for O(1) access by filename (and because the JSON dump is human readable), but that means key iteration cannot be depended on for order.

It's a good point though -- and would simplify testing and other logic. Maybe you would like to work on it as a next PR? =D If we sort the keys before iterating through it, that should fix stuff.

cleborys:
@domanchi Sorry, I am a bit confused :sweat_smile: 
I meant the list that is stored under the `filename` key and contains the secrets in that file. The `merge_results` function assumes that updates to that list are in order so that it is easier to update, but that now means I have some problems when stepping back on choices in the audit:

Currently the audit iterates over all secrets and automatically skips these that already had the `is_secret` flag set. But that means that when I skip back one secret after I previously chose y or n, it will be automatically skipped again, hence the back option currently only works correctly when the last option chosen was skip.
Instead I would like to change the loop to only iterate over secrets without the `is_secret` flag set and then append all previously known secrets that had there flag set already. But that would shuffle the list of secrets (the lists that are stored by filename) and I'm not sure if it is ok to shuffle this (after removing the "order" assumption of `merge_results`.

I am not completely confident that that would work as stated, because simply removing the "has `is_secret`-flag already set, then skip"-condition seems to produce lots of identical secrets to be shown, so there seems to be an extra effect that I don#t understand yet :wink:

cleborys:
@KevinHock  Thanks :) But it is not yet functional! Only the framework with the new iterator works yet, so don't merge :wink: 

(I don't really know the etiquette yet - I made a pull request already so that you can see whats happening, but I wouldn't remove the [WIP] tag in the title until it is functional and I think I'm done. So I hope your approval doesn't mean that you intend to merge)

domanchi:
@cleborys, ah, I understand you better now.

Yes, the secrets in the list are ordered. One benefit of this is comparing side by side diffs of baselines - and it's very easy to see the changes between iterations. You can quite easily see whether a secret has been removed, added, or appropriately labelled. By reordering the contents, it becomes that much harder to see differences, if applicable.

For example, if you start the audit process with pre-existing audited secrets, but not perform any additional labelling, you would expect no changes to the baseline.

If the goal is to allow the `back` option to iterate over previously labelled secrets, perhaps you can just append that to the if statement?

e.g. something like

```
if 'is_secret' not in secret or decision == 'b':
    blah
```

cleborys:
@domanchi Cool, I will not mess around with the order then :sweat_smile: 
Yes, I will try something like this. it is not completely straightforward, because you don't want to have the user step back into secrets that were previously autoskipped and that he is not aware of. 
I think I will split the list into secrets shown to the user and secrets not shown before the loop and then interlace them in the correct order after the loop, again :)

cleborys:
It seems I had forgotten how Python works ("everything is a pointer") and in the end the necessary changes were much easier.

The user decision loop now only loops over secrets which don't have the `is_secret` flag set and decisions modify them in-place. So the order is kept and there is no need to backtrack changes when they are overridden. The `results` dict by filename is constructed after the user is done and then merged.

I expect that everything works as it should now, but I'll add some tests to make sure :smile:

cleborys:
Tests written and passing, some manual sanity checks also successful! :tada: :balloon: 
Thanks, guys! Ready for final review.

PR #76, opened 2018-09-17
Added a KeywordDetector plugin

KevinHock:
Our version of https://github.com/PyCQA/bandit/blob/master/bandit/plugins/general_hardcoded_password.py#L24

### DIFF
--- a/detect_secrets/core/secrets_collection.py
+++ b/detect_secrets/core/secrets_collection.py
@@ -88,8 +88,8 @@ class SecretsCollection(object):
                 secret = PotentialSecret(
                     item['type'],
                     filename,
-                    item['line_number'],
                     secret='will be replaced',
+                    lineno=item['line_number'],
                     is_secret=item.get('is_secret'),
                 )
                 secret.secret_hash = item['hashed_secret']
@@ -204,7 +204,7 @@ class SecretsCollection(object):
         if type_:
             # Optimized lookup, because we know the type of secret
             # (and therefore, its hash)
-            tmp_secret = PotentialSecret(type_, filename, 0, 'will be overriden')
+            tmp_secret = PotentialSecret(type_, filename, secret='will be overriden')
             tmp_secret.secret_hash = secret
 
             if tmp_secret in self.data[filename]:
--- a/detect_secrets/core/secrets_collection.py
+++ b/detect_secrets/core/secrets_collection.py
@@ -251,18 +251,48 @@ class SecretsCollection(object):
                  Caller is responsible for updating the dictionary with
                  results of plugin analysis.
         """
-        results = {}
+        file_results = {}
 
         for plugin in self.plugins:
-            yield results, plugin
+            yield file_results, plugin
 
-        if not results:
+        if not file_results:
             return
 
+        self._remove_password_secrets_if_line_reported_already(
+            file_results,
+        )
+
         if filename not in self.data:
-            self.data[filename] = results
+            self.data[filename] = file_results
         else:
-            self.data[filename].update(results)
+            self.data[filename].update(file_results)
+
+    def _remove_password_secrets_if_line_reported_already(
+        self,
+        file_results,
+    ):
+        """
+        It is often the case that e.g.
+            SUPER_SECRET_VALUE = 'c3VwZXIgbG9uZyBzdHJ'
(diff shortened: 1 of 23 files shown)


detect_secrets/plugins/password.py
```diff
@@ -0,0 +1,68 @@
+"""
+This code was extracted in part from
+https://github.com/PyCQA/bandit. Using similar heuristic logic,
```
KevinHock: Greeting @ericwb and @viraptor, you both seem like the primary maintainers of Bandit.

This code is largely borrowed from [plugins/general_hardcoded_password.py](https://github.com/PyCQA/bandit/blob/master/bandit/plugins/general_hardcoded_password.py#L24) and I've tried my best to reference it as such per the APACHE license of Bandit. Can you please give this your blessing?

detect_secrets/plugins/password.py
```diff
@@ -0,0 +1,68 @@
...
+    'passwd',
+    'pwd',
+    'secret',
+    'secrete',
+    'token',
+)
+
+
+class PasswordDetector(BasePlugin):
+    """This checks for private keys by determining whether the blacklisted
```
domanchi: whoops. copy pasta?

detect_secrets/core/secrets_collection.py
```diff
@@ -251,18 +251,48 @@ def _results_accumulator(self, filename):
...
+        )
+
         if filename not in self.data:
-            self.data[filename] = results
+            self.data[filename] = file_results
         else:
-            self.data[filename].update(results)
+            self.data[filename].update(file_results)
+
+    def _remove_password_secrets_if_line_reported_already(
```
domanchi: Instead of removing, why don't we change our baseline schema to display types in list form? This way, we can preserve all the intelligence that we gather on a particular flagged secret.

detect_secrets/plugins/password.py
KevinHock: Indeed

detect_secrets/core/secrets_collection.py
KevinHock: Well, it would break all the existing baseline's we made ([like when we changed the high entropy string types](https://github.com/Yelp/detect-secrets/pull/26/files)). I'm not sure how hard that would be to implement though, I can look into that.

domanchi: Hmm, good point. Though, arguably a step in the right direction?

Thankfully, we built things for doing better version bumps, since the last time. We would need to update the [merge_baseline](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/baseline.py#L151) function, and the pre-commit hook should get all *active* clients [slowly upgrading](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/pre_commit_hook.py#L162).

The only gotchas I see are:
* Server needs to be updated (and handle stale clients through detected baseline versions)
* We need to handle this case in [audit](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/audit.py#L315).

Perhaps the fix is merely to implement the dedup logic in auditing (as it's the only place that I can find that uses the `type` output from the baseline)?

e.g.

```
def to_list(value):
    try:
        iter(value)
        return value
    except ValueError:
        return [value]

...

plugin = initialize.from_secret_type(
    to_list(secret['type']),
    plugin_settings,
)
```

KevinHock: When you said `Perhaps the fix is merely to implement the dedup logic in auditing`, did you mean you don't mind having all the `Password` ones separate from the others in the baseline? We might double our baseline sizes but I think I'm leaning towards that a little :D

KevinHock: (Stating for the record, discussed in person yesterday, that we will just take the baseline size increase, and punt of de-duping in the audit functionality until later.)

OiCMudkips:
Maybe add a note to the README that this at least attempts to block these now? Otherwise looks good to me.

KevinHock:
Good point!

detect_secrets/plugins/password.py
KevinHock: Thanks @viraptor! Feel free to checkout [PyT](https://github.com/python-security/pyt) sometime ;)

PR #78, opened 2018-09-28
Nicer JSON output + exclude regex bug fix

domanchi:
\[ https://github.com/Yelp/detect-secrets/commit/658ad0d14b00bb7d595cd1653a67febc4536693c ]
It always annoyed me that the JSON output from `scan` had trailing whitespace. However, looks like this is an [old, known issue, that only happens in < Python 3.4](https://bugs.python.org/issue16333), and the [fix](https://docs.python.org/2/library/json.html#json.dump) is merely to add that line to it.

\[ https://github.com/Yelp/detect-secrets/commit/0a6f767f7961c356e937c886575ec69cb7e8e3ae ]
Fix to minor bug I discovered:
```
$ cat .secrets.baseline
{
  "exclude_regex": "^.secrets.baseline$",
  ...
}
$ detect-secrets scan --update .secrets.baseline
$ cat .secrets.baseline
{
  "exclude_regex": "^.secrets.baseline$|^.secrets.baseline",
  ...
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 /.coverage
 /.pytest_cache
 /.tox
-/venv
+/venv**
 /tmp
 
 .*ignore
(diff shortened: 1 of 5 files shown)


KevinHock:
Squeaky clean commit history!

Issue #79, opened 2018-10-02
BasicAuth regex run all night long


KevinHock:
Regex in question
https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/basic_auth.py#L10

Maybe we can run truffle hog regex, instead

Issue #77, opened 2018-09-21
Performance issues


kytwb:
I've been trying to run this tool against the [discourse](https://github.com/discourse/discourse) repository. 
It seems to get stuck when encountering [this file](https://github.com/discourse/discourse/blob/master//plugins/discourse-narrative-bot/lib/discourse_narrative_bot/certificate_generator.rb):

```
[secrets_collection]    INFO    Checking file: ./plugins/discourse-narrative-bot/lib/discourse_narrative_bot/certificate_generator.rb
```

Any workaround/suggestion?

KevinHock:
Hi @kytwb, sorry for the late reply, I strongly suspect that it is due to https://github.com/Yelp/detect-secrets/issues/79, we'll be working on fixing that soon.

domanchi:
Hi @kytwb,

We've merged an update to the BasicAuth plugin, which significantly improves the performance of it. That will go out in the next release, and will work with the file you specified.

If you want, you can pull master and run it against repo, or you can try running the scan with `--no-basic-auth-plugin`.

KevinHock:
(Bumped to version `0.10.3`)

kytwb:
Thanks for the reactivity guys 🎉

streichsbaer:
> Hi @kytwb,
> 
> We've merged an update to the BasicAuth plugin, which significantly improves the performance of it. That will go out in the next release, and will work with the file you specified.
> 
> If you want, you can pull master and run it against repo, or you can try running the scan with `--no-basic-auth-plugin`.

Just tried the cmd line switch and it didn't work. To disable plugins use `--no-{plugin-name}` e.g `--no-basic-auth`, or `--no-keyword.

domanchi:
@streichsbaer : ah, whoops! Thanks for the correction. If interested, the disable command line flags are found here: https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/usage.py#L225

PR #80, opened 2018-10-03
More performant BasicAuth plugin

domanchi:
### Summary

This should make the BasicAuth plugin much more performant, and fixes https://github.com/Yelp/detect-secrets/issues/79 and https://github.com/Yelp/detect-secrets/issues/77.

### Testing

```
$ time python detect_secrets/main.py scan --no-hex-string-scan --no-base64-string-scan --no-keyword-scan --no-private-key-scan <large-file>
```

It actually finishes this time! Furthermore, I modified the large file by inserting a secret that I'd expect it to find, and it finds it as well.

### Rationale

We're already doing `BASIC_AUTH_REGEX.findall(line)` which will run the regex on all non-overlapping matches of the pattern in the string. This means that we can ignore the prefix catch-all expression (`.*?`) because we don't care about that, and we can let the python library handle it for us.

### DIFF
--- a/detect_secrets/plugins/basic_auth.py
+++ b/detect_secrets/plugins/basic_auth.py
@@ -7,7 +7,7 @@ from detect_secrets.core.potential_secret import PotentialSecret
 
 
 BASIC_AUTH_REGEX = re.compile(
-    r'.*?://[^:]+:([^@]+)@',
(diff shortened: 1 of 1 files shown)


JoshuaRLi:
Nice one @domanchi!

PR #83, opened 2018-10-08
Fixes for audit

jkozera:
This PR fixes two issues I've noticed in the `audit` mode:

1. `keyword` plugin failing with `SecretNotFoundOnSpecifiedLineError` if the secret is different in terms of lowercase vs uppercase. (before this change the `keyword` plugin is case-insensitive at https://github.com/Yelp/detect-secrets/blob/d912ced2e7a7607f06794e3d916b10e52536bd5c/detect_secrets/plugins/keyword.py#L58 but case-sensitive at https://github.com/Yelp/detect-secrets/blob/d912ced2e7a7607f06794e3d916b10e52536bd5c/detect_secrets/plugins/keyword.py#L70 — and the latter is used by the `audit` mode at https://github.com/Yelp/detect-secrets/blob/cc6bddbd9d2bd84a3fe32fb82db7c750460f812b/detect_secrets/core/audit.py#L313)
2. ` _get_secret_with_context` failing for small files if they don't have `\n` at the end of the file.

### DIFF
--- a/detect_secrets/plugins/keyword.py
+++ b/detect_secrets/plugins/keyword.py
@@ -55,7 +55,7 @@ class KeywordDetector(BasePlugin):
         if WHITELIST_REGEX.search(string):
             return output
 
-        for identifier in self.secret_generator(string.lower()):
+        for identifier in self.secret_generator(string):
             secret = PotentialSecret(
                 self.secret_type,
                 filename,
--- a/detect_secrets/plugins/keyword.py
+++ b/detect_secrets/plugins/keyword.py
@@ -68,5 +68,5 @@ class KeywordDetector(BasePlugin):
 
     def secret_generator(self, string):
         for line in BLACKLIST:
-            if line in string:
+            if line.lower() in string.lower():
                 yield line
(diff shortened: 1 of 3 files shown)


KevinHock:
This looks great to me! Thank you so much for making this, squeaky clean commits too. :)

detect_secrets/plugins/keyword.py
```diff
@@ -55,7 +55,7 @@ def analyze_string(self, string, line_num, filename):
         if WHITELIST_REGEX.search(string):
             return output
 
-        for identifier in self.secret_generator(string.lower()):
+        for identifier in self.secret_generator(string):
```
domanchi: It's probably more performant to make the string lowered once here, and passed into `secret_generator`, than to make it `.lower()` every time.

detect_secrets/plugins/keyword.py
```diff
@@ -68,5 +68,5 @@ def analyze_string(self, string, line_num, filename):
 
     def secret_generator(self, string):
         for line in BLACKLIST:
-            if line in string:
+            if line.lower() in string.lower():
```
domanchi: Why not just change `PASS = ` in BLACKLIST to lowercase?

jkozera:
@domanchi Good points :) I will update the PR shortly.

PR #86, opened 2018-10-22
Use different regexes in KeywordDetector to improve accuracy

KevinHock:


detect_secrets/plugins/keyword.py
```diff
@@ -26,22 +26,54 @@
...
     'password',
     'passwd',
-    'pwd',
+    'private_key',
     'secret',
     'secrete',
     'token',
 )
+# Uses lazy quantifiers
+BLACKLIST_REGEX = re.compile(
```
KevinHock: Ends up being 
`'(.*?)apikey(.*?)"(.*?)";|(.*?)api_key(.*?)"(.*?)";|(.*?)pass(.*?)"(.*?)";|(.*?)password(.*?)"(.*?)";|(.*?)passwd(.*?)"(.*?)";|(.*?)private_key(.*?)"(.*?)";|(.*?)secret(.*?)"(.*?)";|(.*?)secrete(.*?)"(.*?)";|(.*?)token(.*?)"(.*?)";|(.*?)apikey[:](.*?)|(.*?)api_key[:](.*?)|(.*?)pass[:](.*?)|(.*?)password[:](.*?)|(.*?)passwd[:](.*?)|(.*?)private_key[:](.*?)|(.*?)secret[:](.*?)|(.*?)secrete[:](.*?)|(.*?)token[:](.*?)|(.*?)apikey(.*?)=(.*?)|(.*?)api_key(.*?)=(.*?)|(.*?)pass(.*?)=(.*?)|(.*?)password(.*?)=(.*?)|(.*?)passwd(.*?)=(.*?)|(.*?)private_key(.*?)=(.*?)|(.*?)secret(.*?)=(.*?)|(.*?)secrete(.*?)=(.*?)|(.*?)token(.*?)=(.*?)|^pwd(.*?)=(.*?)'`

detect_secrets/plugins/keyword.py
```diff
@@ -26,22 +26,54 @@
...
+    r'|'.join(
+        '(.*?){}(.*?)"(.*?)";'.format(
+            value,
+        )
+        for value in BLACKLIST
+    ) + '|' +
+    # Followed by a :
+    # e.g. token:
+    r'|'.join(
+        '(.*?){}[:](.*?)'.format(
```
domanchi: Why do we have a capturing class with only one element? Could also just do `{}:`

detect_secrets/plugins/keyword.py
```diff
@@ -49,30 +49,31 @@
...
+        '{}(.*?)("|\')(\\S*?)(\'|");'.format(
             value,
         )
         for value in BLACKLIST
     ) + '|' +
     # Followed by a :
     # e.g. token:
     r'|'.join(
-        '(.*?){}[:](.*?)'.format(
+        '{}:'.format(
```
domanchi: We need to add a capturing group here. Otherwise, we're not going to get anything.

detect_secrets/plugins/keyword.py
```diff
@@ -49,30 +49,31 @@
...
             value,
         )
         for value in BLACKLIST
     ) + '|' +
     # Follwed by an = sign
     # e.g. my_password =
     r'|'.join(
-        '(.*?){}(.*?)=(.*?)'.format(
+        '{}(.*?)='.format(
```
domanchi: Same for this: we need to capture the secret here, instead of the content in between the keyword and the equal sign.

detect_secrets/plugins/keyword.py
```diff
@@ -49,30 +49,31 @@
...
             value,
         )
         for value in BLACKLIST
     ) +
-    # pwd has to start with pwd, it is too common
-    '|^pwd(.*?)=(.*?)',
+    # For `pwd` it has to start with pwd after whitespace, it is too common
+    '|\\spwd(.*?)=',
```
domanchi: This too.

detect_secrets/plugins/keyword.py
```diff
@@ -49,30 +49,31 @@
 BLACKLIST_REGEX = re.compile(
     # Followed by double-quotes and a semi-colon
     # e.g. private_key "something";
+    # e.g. private_key 'something';
     r'|'.join(
-        '(.*?){}(.*?)"(.*?)";'.format(
+        '{}(.*?)("|\')(\\S*?)(\'|");'.format(
```
domanchi: We've got a bit too many capturing groups here, which will make `secret_generator` produce unexpected results.

You can use `(?:<>)` to indicate a non-capturing group.

tests/plugins/keyword_test.py
```diff
@@ -64,7 +64,7 @@ def test_analyze_positives(self, file_content):
         'file_content',
         [
             (
-                'private_key \'something\';'  # Single-quotes not double-quotes
+                'private_key \'"no spaces\';'  # Has whitespace in-between
```
domanchi: Is the `"` expected?

KevinHock: @domanchi Do we want to have a regex for single-quotes, and a regex for double-quotes? I can't think of either way I'm happy with.
e.g. `private_key_for_blah "grammar' messes thing's up bad"` with this regex only returns `grammar`, whereas if we had 2 there would be 2 matches. One for the whole string (the double-quotes, which is good), and one for the single-quotes (which is ` messes thing`, and bad).
Unless we could somehow have a stack, and only match, the 2nd time, the first type of quote that was matched. I'll see how hard that is.

We have only seen it followed by double-quotes though, so maybe we should stick to that?

tests/plugins/keyword_test.py
```diff
@@ -12,24 +13,137 @@ class TestKeywordDetector(object):
...
+                'private_key "";'  # Nothing in the quotes
+            ),
+            (
+                'private_key \'"no spaces\';'  # Has whitespace in the secret
+            ),
+            (
+                'private_key "fake";'  # 'fake' in the secret
+            ),
+            (
+                'private_key "hopenobodyfindsthisone\';'  # Double-quote does not match single-quote
```
KevinHock: Alright, so, I ought to fix this one so that it catches it.

It's a `Nit:` but 🤷‍♂️

KevinHock:
Just stating for memories sake, that we had a discussion that I'll paraphrase here, I'll maybe make a separate issue:

Currently, the assumption is that "no secret is repeated within a single file" *however*, with this plugin it is more likely that this assumption is gonna break e.g. if you have 2 different assignments of `password = "ehehehwew"` in the same file, the hash of `ehehehwew` will be in 1 place in the baseline, but 2 places in the file. This is currently true for e.g. high-entropy secrets right now, for instance.

We can either (a) incorporate line numbers in https://github.com/Yelp/detect-secrets/blob/5f4a05523827654cfbc6a7d2d176b48b4d3a79c7/detect_secrets/core/potential_secret.py#L53, or (b) put a count of how many secrets are in the file, and put the count in baseline.

The downsides of doing nothing, and merging the PR as is:
   1. It looks unintelligent to users, as in "This tool doesn't even detect the same secret on the next line."
   2. When removing a secret, you suddenly get alerted to a new one that we did not complain about before.
   3. Less concerning: Adding a new secret can be missed, if it is already in that file.
   4. Less concerning: The audit command does not show the secret more than once.

We agree changing that assumption is a larger task than what's at hand.

tests/plugins/keyword_test.py
KevinHock: Decided not to, b/c it's not valid syntax in php anyhow, apparently

detect_secrets/core/audit.py
```diff
@@ -572,10 +583,14 @@ def _highlight_secret(secret_line, secret_lineno, secret, filename, plugin_setti
     )
 
 
-def _raw_secret_generator(plugin, secret_line):
+def _raw_secret_generator(plugin, secret_line, is_php_file):
     """Generates raw secrets by re-scanning the line, with the specified plugin"""
-    for raw_secret in plugin.secret_generator(secret_line):
-        yield raw_secret
+    if isinstance(plugin, KeywordDetector):
```
KevinHock: I felt :/ about writing it like this, but didn't see a better way.

Just wanted to put neon lights on it for the review 😅

detect_secrets/plugins/keyword.py
```diff
@@ -170,17 +189,23 @@ def secret_generator(self, string, filetype):
...
+            )
+        ) or (  # If it is a .php file, do not report $variables
             filetype == FileType.PHP
             and lowered_secret[0] == '$'
         )
     ):
         return True
 
-    # No function calls
+    # Hueristic for no function calls
```
calvinli: nit: it's `heuristic`

KevinHock: Great catch! I can `git commit --amend` it like it never happened.

detect_secrets/plugins/keyword.py
```diff
@@ -26,34 +26,116 @@
...
 
 
+# Note: All values here should be lowercase
 BLACKLIST = (
-    # NOTE all values here should be lowercase,
-    # otherwise _secret_generator can fail to match them
-    'pass =',
+    'apikey',
+    'api_key',
+    'aws_secret_access_key',
```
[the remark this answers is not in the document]
KevinHock: Mostly wanted to point out that this will catch AWS secret keys now. 👍

PR #89, opened 2018-10-23
Add a "git add baseline" message, disable Keyword detector for now

KevinHock:


detect_secrets/pre_commit_hook.py
```diff
@@ -63,8 +63,11 @@ def main(argv=None):
             baseline_collection.format_for_baseline_output(),
         )
 
-        # The pre-commit framework should automatically detect a file change
-        # and print a relevant error message.
+        log.error(
+            'The baseline file was updated.\n'
+            'Probably to keep line numbers of secrets up-to-date.\n'
+            'Please `git add {}`, thank you.\n\n'.format(args.baseline[0]),
```
domanchi: This case doesn't matter, but something to keep in mind for future: https://timber.io/blog/the-pythonic-guide-to-logging/#performance

PR #90, opened 2018-10-24
Update merge results to only propagate is_secret of new secrets

KevinHock:
When I do `detect-secrets -vvv scan --update .secrets.baseline` it'll run with new plugins, so I changed some of the assumptions in `merge_results`

### DIFF
--- a/detect_secrets/core/baseline.py
+++ b/detect_secrets/core/baseline.py
@@ -188,17 +188,21 @@ def merge_results(old_results, new_results):
         if filename not in new_results:
             continue
 
-        for new_secret in new_results[filename]:
-            for old_secret in old_secrets:
-                if old_secret['hashed_secret'] != new_secret['hashed_secret']:
-                    # We don't join the two secret sets, because if the later
-                    # result set did not discover an old secret, it's probably
-                    # moved.
-                    continue
+        old_secrets_mapping = dict()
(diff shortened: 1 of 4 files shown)


detect_secrets/core/baseline.py
```diff
@@ -190,42 +184,21 @@ def merge_results(old_results, new_results):
...
     :rtype: dict
     """
-    for filename, secrets in old_results.items():
+    for filename, old_secrets in old_results.items():
         if filename not in new_results:
-            new_results[filename] = secrets
             continue
 
-        if len(secrets) == len(new_results[filename]):
-            # Assuming that secrets remain in order.
```
domanchi: Secrets are still ordered, because of https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/baseline.py#L233. So we may not need to use this O(n^2) approach.

KevinHock: Amazing catch, the solution would have been clear if it was `merge 234 and 156`, as you could easily tell 1 < 2 and increment the list with the lower element, but since we were talking `hashed_secrets`'s and I didn't want to do anything based off line numbers, I made it a `dict` instead `old_secrets_mapping[old_secret['hashed_secret']] = old_secret` 👍

PR #91, opened 2018-10-26
Improve 'Unable to open baseline file' message

KevinHock:
fix `--import`->`--update` comment,
bump to 0.10.5,
hardcode flake8==3.5.0


detect_secrets/pre_commit_hook.py
```diff
@@ -122,9 +122,11 @@ def _get_baseline_string_from_file(filename):   # pragma: no cover
 
     except IOError:
         log.error(
-            'Unable to open baseline file: %s.', filename,
+            'Unable to open baseline file: {}\n'
+            'Please create it via\n'
+            '   `detect-secrets scan > .secrets.baseline`\n'
```
domanchi:
```suggestion
            '   `detect-secrets scan > {}`\n'
```

Remember that we never explicitly enforce a certain baseline filename to be used.

PR #95, opened 2018-11-07
Adding `audit --diff` functionality

domanchi:
## Testing

```
detect-secrets$ pip install -e .
detect-secrets$ detect-secrets scan test_data > .secrets.baseline
detect-secrets$ detect-secrets scan --base64-limit 5.5 --hex-limit 2 > .secrets.baseline.new
detect-secrets$ detect-secrets audit --diff .secrets.baseline .secrets.baseline.new
```

### Output

```
Secret:      1 of 10
Filename:    test_data/config.ini
Secret Type: Hex High Entropy String
Status:      >> ADDED <<
----------
1:[credentials]                                    
2:password = 12345678901234                        
3:                                                 
4:[parent]                                         
5:    [child]                                      
6:    keyA = 678912345                             
7:    keyB = value1 
----------
What would you like to do? (s)kip, (q)uit:   
```

## Features

* Leverages the full colored, interactive auditor to be able to see which secrets have been added and removed from the scan, based on plugin configurations.

Hopefully this will better support plugin developers, to make the process easier to configure limits and plugin regexes!

## Backwards Compatibility Considerations

* This requires 0.10.4 at a minimum, because it expects baselines to be [*fully* sorted](https://github.com/Yelp/detect-secrets/commit/ad1de5d3dad066d00da1310802ba83580a87fe13)


test_data/each_secret.py
```diff
@@ -0,0 +1,5 @@
+red_herring = 'DEADBEEF'
+
+base64_secret = 'c2VjcmV0IG1lc3NhZ2Ugc28geW91J2xsIG5ldmVyIGd1ZXNzIG15IHBhc3N3b3Jk'
+hex_secret = '8b1118b376c313ed420e5133ba91307817ed52c2'
+basic_auth = 'http://username:whywouldyouusehttpforpasswords@example.com'
```
KevinHock: :D

detect_secrets/core/audit.py
```diff
@@ -109,11 +194,128 @@ def _secret_generator(baseline):
...
+    def _check_secret(a, b):
+        if a == b:
+            return 0
+
+        if a['line_number'] < b['line_number']:
+            return -1
+        elif a['line_number'] > b['line_number']:
+            return 1
+
+        return _check_string(a['hashed_secret'], b['hashed_secret'])
```
KevinHock: As a reminder of context: We [punted off](https://github.com/Yelp/detect-secrets/pull/76) changing the audit functionality to de-duplicate secrets with the same `hashed_secret`, and accepted larger baselines.

However this brings up something interesting, if I merged in the [current Keyword plugin PR](https://github.com/Yelp/detect-secrets/pull/86), and the Keyword detector found something of the same hash (new baseline) as a Hex High Entropy plugin in the old baseline, either (a) we would say that the Keyword Detector secret got added (and that wouldn't mean we found a new secret) or (b) we would say that the Hex High Entropy plugin got added, (assuming the Hex High Entropy secret was in both baseline) which is definitely not True.

detect_secrets/__init__.py
```diff
@@ -1 +1 @@
-VERSION = '0.10.5'
+VERSION = '0.11.0'
```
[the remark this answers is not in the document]
KevinHock: I should update the changelog, but https://github.com/Yelp/detect-secrets/pull/90 was a bug fix for the update functionality, before running the temp bandaid fixer.

PR #100, opened 2018-12-10
add plugin to look for AWS key IDs

calvinli:
AWS key IDs are not strictly speaking secrets, although we've found that files containing one almost always contain the corresponding private key. (The only notable false positives I could find in an internal search are AWS S3 pre-signed URLs, which include the signing key's ID in the URL. These are arguably potentially secret anyway.)

This was implemented by adding a new plugin base class (`RegexBasedDetector`) and adding the AKIA regex as a subclass; this is a bit verbose, but necessary if we want the ability to turn individual regex-based checks on and off. It looks like potentially `KeywordDetector`, `PrivateKeyDetector`, and `BasicAuthDetector` could also be implemented using`RegexBasedDetector` too, if we wanted to consolidate code a bit.

Change is test-coverage neutral.


detect_secrets/plugins/core/initialize.py
```diff
@@ -10,6 +10,7 @@
 from ..high_entropy_strings import HexHighEntropyString     # noqa: F401
 from ..keyword import KeywordDetector                       # noqa: F401
 from ..private_key import PrivateKeyDetector                # noqa: F401
+from ..aws import AWSKeyDetector                            # noqa: F401
```
KevinHock: Nit: We normally do alphabetical imports

detect_secrets/plugins/regexes.py
```diff
@@ -0,0 +1,36 @@
+"""
+This plugin searches each line with a set of regexes.
+"""
+from __future__ import absolute_import
+
+from .base import BasePlugin
+from detect_secrets.core.potential_secret import PotentialSecret
+
+
+class RegexBasedDetector(BasePlugin):
```
KevinHock: Maybe make this class abstract as well, `e.g. __metaclass__ = ABCMeta` etc.

detect_secrets/plugins/core/initialize.py
```diff
@@ -10,6 +10,7 @@
 from ..high_entropy_strings import HexHighEntropyString     # noqa: F401
```
KevinHock: Below in this file is

https://github.com/Yelp/detect-secrets/blob/c0bea0b9b0dad74edf5d197eada36dfa64dd3e5a/detect_secrets/plugins/core/initialize.py#L92

you'll probably want to do this with `BasePlugin` as well.

cc @domanchi for confirmation

detect_secrets/plugins/regexes.py
```diff
@@ -0,0 +1,36 @@
+"""
```
domanchi: If this is a base class, probably belongs in `plugins/base.py`.

detect_secrets/plugins/regexes.py
```diff
@@ -0,0 +1,36 @@
...
+from __future__ import absolute_import
+
+from .base import BasePlugin
+from detect_secrets.core.potential_secret import PotentialSecret
+
+
+class RegexBasedDetector(BasePlugin):
+    """Base class for regular-expressed based detectors.
+
+    Replace `secret_type` with a description and `blacklist`
```
domanchi: We can enforce this better with `abc.abstractproperty`.

detect_secrets/plugins/core/initialize.py
calvinli: `RegexBasedDetector` isn't imported here, so it won't be included anyway

domanchi: > You'll probably want to do this with `BasePlugin` as well.

Accurate, but for `RegexBasedDetector`.

```
if issubclass(value, BasePlugin) and value not in [BasePlugin, RegexBasedDetector]:
```

domanchi:
Mostly LGTM. We can keep things DRYer by refactoring `plugins/basic_auth.py` and `plugins/private_key.py` too! I'd probably stay away from touching `plugins/keyword.py` though, because of the [improvements still sitting in queue](https://github.com/Yelp/detect-secrets/pull/86).

calvinli:
> Mostly LGTM. We can keep things DRYer by refactoring `plugins/basic_auth.py` and `plugins/private_key.py` too! I'd probably stay away from touching `plugins/keyword.py` though, because of the [improvements still sitting in queue](https://github.com/Yelp/detect-secrets/pull/86).

let's do that in a different PR...

PR #103, opened 2018-12-18
refactor various detectors to use RegexBasedDetector

JoshuaRLi:
Follow up to the `RegexBasedDetector` implemented in https://github.com/Yelp/detect-secrets/pull/100.

### DIFF
--- a/detect_secrets/plugins/base.py
+++ b/detect_secrets/plugins/base.py
@@ -126,5 +126,5 @@ class RegexBasedDetector(BasePlugin):
 
     def secret_generator(self, string):
         for regex in self.blacklist:
-            if regex.search(string):
-                yield regex.pattern
+            for match in regex.findall(string):
+                yield match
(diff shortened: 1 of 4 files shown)


KevinHock:
Let's merge in https://github.com/Yelp/detect-secrets/pull/86 before this one

JoshuaRLi:
If that's done, then the changes to `KeywordDetector` in this branch should be discarded.

detect_secrets/plugins/base.py
```diff
@@ -126,5 +126,5 @@ def analyze_string(self, string, line_num, filename):
 
     def secret_generator(self, string):
         for regex in self.blacklist:
-            if regex.search(string):
-                yield regex.pattern
+            for match in regex.findall(string):
+                yield match
```
JoshuaRLi: This is the desired behavior.

detect_secrets/plugins/private_key.py
```diff
@@ -1,68 +1,25 @@
-"""
-This code was extracted in part from
```
KevinHock: I'd recommend we keep this, better safe than sorry IMO https://github.com/Yelp/detect-secrets/issues/31

JoshuaRLi: Alright.

PR #105, opened 2018-12-21
add whitelist directive regexes to match against inline comment syntaxes in more languages

JoshuaRLi:
Also:

- improved the base regex to account for leading whitespace before the inline comment start character, and trailing whitespace until EOL.
- some docs tweaks

The only suitable and pre-existing place I could find was in `tests/plugins/high_entropy_strings_test.py`. Ideally, tests should be moved to (and should only exist exclusively in) `tests/plugins/base_test.py`. The redundant logic to detect whether or not to ignore a file should also be removed from `YamlFileParser` and the `HighEntropyStringsPlugin`. The latter should be refactored so that the logic in `analyze_string` is run by `BasePlugin`'s `analyze`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ either the client-side pre-commit hook, or the server-side secret scanner.
 3. **Secrets Baseline**, to whitelist pre-existing secrets in the repository,
    so that they won't be continuously caught through scan iterations.
 
-### Client-side Pre-commit Hook
+### Client-side `pre-commit` Hook
 
 See [pre-commit](https://github.com/pre-commit/pre-commit) for instructions
 to install the pre-commit framework. The example usage above has a sample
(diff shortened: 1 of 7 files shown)


JoshuaRLi:
Ah, actually the logic was being run twice - once in `HighEntropyStringsPlugin.analyze_string` and another in `HighEntropyStringsPlugin.analyze`'s `self._analyze_yaml_file`. The former can be removed.

JoshuaRLi:
So what remains here is that the `IniFileParser` should implement something like `YamlFileParser`'s `get_ignored_lines`. And a follow-up to that would be to make a base parser class for those to consolidate that logic. Then more thought should be put into language-specific parsers, and then I think the catch-all list of whitelist directive regexes should instead be split up into those parsers. That would be a bit of initial boilerplate, but it would make tests neater and make language support more explicit and modular.

PR #107, opened 2018-12-21
small improvements to existing FileParsers

JoshuaRLi:
- moved some stuff around in `IniFileParser`; things needed in smaller scopes were lazily created and constants were moved to object-level scope.
- `YamlFileParser`: small follow up to #105; only the whitelist directive regex for YAML inline comment syntax should be used.

### DIFF
--- a/detect_secrets/plugins/core/constants.py
+++ b/detect_secrets/plugins/core/constants.py
@@ -11,7 +11,18 @@ WHITELIST_REGEXES = [
             ('\'', ''),             # e.g. visual basic .net
             ('--', ''),             # e.g. sql
             # many other inline comment syntaxes are not included,
-            # because we want to be performant for the common case
+            # because we want to be performant for
+            # any(regex.search(line) for regex in WHITELIST_REGEXES)
+            # calls. of course, this won't be a concern if detect-secrets
+            # switches over to implementing file plugins for each supported
+            # filetype.
         )
     ]
 ]
+
+# add to this mapping (and WHITELIST_REGEXES if applicable) lazily,
+# as more language specific file parsers are implemented.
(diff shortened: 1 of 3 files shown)


detect_secrets/plugins/core/ini_file_parser.py
```diff
@@ -4,14 +4,16 @@
...
+    _comment_regex = re.compile(r'\s*[;#]')
+
     def __init__(self, file):
         self.parser = configparser.ConfigParser()
         self.parser.optionxform = str
         self.parser.read_file(file)
 
         # Hacky way to keep track of line location
         file.seek(0)
-        self.lines = list(map(lambda x: x.strip(), file.readlines()))
```
KevinHock: ++

detect_secrets/plugins/core/constants.py
```diff
@@ -11,7 +11,18 @@
...
+            # filetype.
         )
     ]
 ]
+
+# add to this mapping (and WHITELIST_REGEXES if applicable) lazily,
+# as more language specific file parsers are implemented.
+# discussion: https://github.com/Yelp/detect-secrets/pull/105
+WHITELIST_REGEX = {
+    'yaml': WHITELIST_REGEXES[0],
```
KevinHock: Super Nit: Maybe add to the regex comment `# e.g. python` -> `# e.g. python or yaml`

JoshuaRLi: I don't think that'll be necessary if a python file parser is implemented. Then we can `'py': WHITELIST_REGEXES[0]`. More language/specification specific file parsers is what I'm pushing for, or at least suggesting, architecturally. See #105.

Issue #106, opened 2018-12-21
ini-like file not parsing


domanchi:
```
$ cat environment.variables
PASSWORD=d1bc8d3ba4afc7e109612cb73acbdddac052c93025aa1f82942edabb7deb82a1
$ detect-secrets scan environment.variables
{
  "exclude_regex": null,
  "generated_at": "2018-12-21T20:02:03Z",
  "plugins_used": [
    {
      "base64_limit": 4.5,
      "name": "Base64HighEntropyString"
    },
    {
      "name": "BasicAuthDetector"
    },
    {
      "hex_limit": 3,
      "name": "HexHighEntropyString"
    },
    {
      "name": "PrivateKeyDetector"
    }
  ],
  "results": {},
  "version": "0.11.0"
}
```

Environment variables look like `.ini` files, but without a header. We should be able to capture this using this fact.

PR #108, opened 2018-12-21
Capturing env vars in files

domanchi:
Fixes: https://github.com/Yelp/detect-secrets/issues/106

### DIFF
--- a/detect_secrets/plugins/high_entropy_strings.py
+++ b/detect_secrets/plugins/high_entropy_strings.py
@@ -52,17 +52,23 @@ class HighEntropyStringsPlugin(BasePlugin):
 
     def analyze(self, file, filename):
         file_type_analyzers = (
-            (self._analyze_ini_file, configparser.Error,),
+            (self._analyze_ini_file(), configparser.Error,),
             (self._analyze_yaml_file, yaml.YAMLError,),
+            (super(HighEntropyStringsPlugin, self).analyze, Exception,),
+            (self._analyze_ini_file(True), configparser.Error,),
         )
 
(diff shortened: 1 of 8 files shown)


detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -52,17 +52,23 @@ def __init__(self, charset, limit, *args):
 
     def analyze(self, file, filename):
         file_type_analyzers = (
-            (self._analyze_ini_file, configparser.Error,),
+            (self._analyze_ini_file(), configparser.Error,),
             (self._analyze_yaml_file, yaml.YAMLError,),
+            (super(HighEntropyStringsPlugin, self).analyze, Exception,),
+            (self._analyze_ini_file(True), configparser.Error,),
```
KevinHock: Nit: `add_header=True` (so it's clear what we're setting.)

tests/plugins/high_entropy_strings_test.py
```diff
@@ -108,6 +108,25 @@ def test_ignored_lines(self, content_to_format):
...
+        with pytest.raises(ValueError):
+            Base64HighEntropyString(15)
+
+
+class TestBase64HighEntropyStrings(HighEntropyStringsTest):
+
+    def setup(self):
+        super(TestBase64HighEntropyStrings, self).setup(
+            # Testing default limit, as suggested by truffleHog.
+            Base64HighEntropyString(4.5),
```
KevinHock: Super nit: Might be more DRY if we make this a hard-coded value, and export it from `usage.py`

PR #109, opened 2018-12-21
replace BashColor singleton with colorize function

JoshuaRLi:
AFAIK the BashColor singleton exists to toggle color enable/disable for easier testing, but it can indeed be replaced with a more lightweight and reusable `colorize` function.

Ideally, tests should be written to compare colors so the UX can be consistent - I won't be including that in this branch, and instead opted to just strip the ANSI escape sequences away for now since that was essentially what was being done before.

### DIFF
--- a/testing/util.py
+++ b/testing/util.py
@@ -1,0 +1,8 @@
+import re
+
+# https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python
+_ansi_escape = re.compile(r'\x1B\[[0-?]*[ -/]*[@-~]')
+
+
+def uncolor(text):
+    return _ansi_escape.sub('', text)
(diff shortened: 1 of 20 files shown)


testing/util.py
```diff
@@ -0,0 +1,8 @@
+import re
+
+# https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python
```
KevinHock: Wow, Martijn Pieters strikes again.

JoshuaRLi: lol

tests/main_test.py
```diff
@@ -253,7 +252,6 @@ def test_audit_same_file(self):
                 'No difference, because it\'s the same file!'
             )
 
-
```
KevinHock: Still there, right?

JoshuaRLi:
LOOOOL i added `test_audit_diff_not_enough_files` twice during the merge fuckery.

JoshuaRLi:
Friday is a great day!

PR #111, opened 2019-01-04
:bug: Prevent end-of-file fixer complaining about extra newline after `ds scan > baseline`

KevinHock:
By not adding a newline in format_baseline_for_output
And adding the newline in a commonly used write_baseline_to_file function


detect_secrets/main.py
```diff
@@ -45,7 +45,7 @@ def main(argv=None):
             if args.import_filename:
                 _write_to_file(args.import_filename[0], output)
             else:
-                print(output)
+                print(output, end='')
```
domanchi: Let's make the change in `format_baseline_for_output` instead, and write the new line in `_write_to_file`.

That seems less of an anti-pattern.

KevinHock: I'm on the fence about adding `+ '\n'` in 3 places, but I don't have strong feelings about it.

```diff 
 def _save_baseline_to_file(filename, data):  # pragma: no cover
     with open(filename, 'w') as f:
-        f.write(format_baseline_for_output(data))
+        f.write(format_baseline_for_output(data) + '\n')
...
-                _write_to_file(args.import_filename[0], output)
+                _write_to_file(
+                    args.import_filename[0],
+                    output + '\n'
+                )
...
     with open(filename, 'w') as f:
-        f.write(format_baseline_for_output(payload))
+        f.write(format_baseline_for_output(payload) + '\n')
```

domanchi: Maybe something along the lines of:

```
$ cat detect_secrets/core/common.py
def write_to_file(filename, data):
    with open(filename, 'w') as f:
        f.write(format_baseline_for_output(data) + '\n')
```

KevinHock: ✅

PR #114, opened 2019-01-07
Travis CI and tox.ini: Add Python 3.7 to the testing

cclauss:
One warning to be aware of for Python 3.8...
```
=============================== warnings summary ===============================
tests/main_test.py::TestMain::test_reads_from_stdin
  /home/travis/build/Yelp/detect-secrets/.tox/py37/lib/python3.7/site-packages/yaml/constructor.py:126:
  DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    if not isinstance(key, collections.Hashable):
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 project = detect_secrets
 # These should match the travis env list
-envlist = py27,py35,py36,pypy
+envlist = py27,py35,py36,py37,pypy
 tox_pip_extensions_ext_pip_custom_platform = true
 tox_pip_extensions_ext_venv_update = true
 
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,7 @@
 project = detect_secrets
 # These should match the travis env list
 envlist = py27,py35,py36,py37,pypy
+skip_missing_interpreters = true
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 project = detect_secrets
 # These should match the travis env list
-envlist = py27,py35,py36,pypy
+envlist = py27,py35,py36,py37,pypy
```
KevinHock: Could you add a `skip_missing_interpreters = true` under this, in case some developer env's don't have `3.7`

KevinHock:
LGTM, merging now :)

JoshuaRLi:
@KevinHock @cclauss Xenial is now [generally available](https://docs.travis-ci.com/user/reference/xenial/) on Travis, you should be able to have a top-level `dist: xenial` declaration instead of putting it in the matrix.

cclauss:
I can confirm that both solutions work.  While __xenial__ is generally available, it is not the Travis _default_ so I usually take the more conservative route of only using it where it is needed.

JoshuaRLi:
Good point, it's nice to be explicit about nondefaults.

PR #116, opened 2019-01-07
:bug: Fix format_baseline_for_output TypeError

KevinHock:
Where we accidentally passed a str where we were expecting a dict

:snake: Made sure to use kwargs when calling `write_baseline_to_file`
Added some docstrings
Added `ipdb` to `requirements-dev.txt`

### DIFF
--- a/detect_secrets/pre_commit_hook.py
+++ b/detect_secrets/pre_commit_hook.py
@@ -65,8 +65,8 @@ def main(argv=None):
 
     if baseline_modified:
         write_baseline_to_file(
-            args.baseline[0],
-            baseline_collection.format_for_baseline_output(),
+            filename=args.baseline[0],
+            data=baseline_collection.format_for_baseline_output(),
         )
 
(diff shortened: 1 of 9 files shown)


detect_secrets/pre_commit_hook.py
```diff
@@ -65,8 +65,8 @@ def main(argv=None):
 
     if baseline_modified:
         write_baseline_to_file(
-            args.baseline[0],
-            baseline_collection.format_for_baseline_output(),
+            filename=args.baseline[0],
+            data=baseline_collection.format_for_baseline_output(),
```
KevinHock: 🚨 Note that here we have to call `format_for_baseline_output` even though it'll be called again in `write_baseline_to_file ` because there is no other existing good way to create a dict from a `SecretsCollection` object, b/c `.json` only worries about `.data` (`['results']`) and not other attributes. Creating a new one didn't seem worth it.

PR #118, opened 2019-01-29
:telescope: [Keyword Plugin] Add more to FALSE_POSITIVES

KevinHock:
Just a few additions to the `FALSE_POSITIVES` tuple.

The only notable one is that we will *not* alert on `password = 'password'`, I felt comfortable with this because I have seen nothing but false-positives so far.


detect_secrets/plugins/keyword.py
```diff
@@ -62,21 +62,29 @@
     'dummy_secret',
     'false',
     'false):',
+    'false,',
+    'false;',
```
domanchi: At some point, I wonder whether it would be better to just do:

```
FALSE_POSITIVES = []
for key in ['false', 'none', ...]:
    for suffix in [',', ';', '']:
        FALSE_POSITIVES.append(key + suffix)
```

KevinHock: I agree, I think we will at some point, a lot fp's don't have a trailing char so I'm gonna hold off until it is more battle-tested.

Also been meaning to make it a `dict` instead of a tuple.

PR #119, opened 2019-01-30
Add high entropy exclude regex option for each high entropy plugin

KevinHock:
The emojis say it all :sparkles::cake::sparkles:

Primary commit:
🎉 Add high entropy exclude options  …
This is so you can specifically ignore lines with false-positives
that often trigger the high entropy plugins, e.g.
```
"AWS": [
    "CanonicalUser: 79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be"
]
```
can be ignored with e.g.
```
detect-secrets scan test.py --hex-high-entropy-exclude "(CanonicalUser)"
```

### DIFF
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -219,6 +219,7 @@ class PluginOptions(object):
             disable_help_text='Disables scanning for hex high entropy strings',
             related_args=[
                 ('--hex-limit', 3,),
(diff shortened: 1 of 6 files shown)


KevinHock:
(This was reverted in https://github.com/Yelp/detect-secrets/pull/127 btw)

PR #120, opened 2019-01-31
Use python instead of sed/wc during audit

killuazhu:
Update audit code to read files using python instead of sed/wc to align the end of line character handling.

If the audit is performed on a file with non-posix end of line characters then it would incorrectly extract the line range from the file and miscalculate the index of the line the secret is on.  This would result in an error during the audit "Secret not found on line....".

The root cause of this issue was due to audit using "sed" and "wc" to extract line ranges from the file and count total lines.  When these tools encountered the non-posix end of line characters they would calculate different lines then via python which uses a universal or cross-platform end of line characters.

To ensure that it would calculate the same lines as the scan the audit code has been updated to use python to read the original file and extract the specified lines.

CC @jribm

### DIFF
--- a/detect_secrets/core/audit.py
+++ b/detect_secrets/core/audit.py
@@ -1,5 +1,7 @@
 from __future__ import print_function
 
+import codecs
+import itertools
 import json
 import os
 import subprocess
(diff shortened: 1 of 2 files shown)


killuazhu:
@KevinHock Travis build has passed now. Ready for a look.

detect_secrets/core/audit.py
```diff
@@ -451,63 +453,33 @@ def _get_secret_with_context(
...
 
-    try:
-        output[index_of_secret_in_output] = _highlight_secret(
-            output[index_of_secret_in_output],
-            secret_lineno,
-            secret,
-            filename,
-            plugin_settings,
-        )
-    except SecretNotFoundOnSpecifiedLineError:
```
KevinHock: I believe `_highlight_secret` can still throw a `SecretNotFoundOnSpecifiedLineError`, in the case that the baseline is out-of-date.

So I think we do need to keep the `try:` `exception:` code here https://github.com/Yelp/detect-secrets/blob/f0ece437079854910a97394451b3ab66f99ae515/detect_secrets/core/audit.py#L486-L503

killuazhu: It's updated in the latest commit https://github.com/Yelp/detect-secrets/pull/120/commits/4d5ab73816d23081ef1e8b865739db8c22f99c1f

killuazhu:
@KevinHock thanks for accepting the PR. I'm glad it could be accepted so quick! 👍

Issue #121, opened 2019-01-31
Respect plugin list from baseline


killuazhu:
Some of the plugins, in particular, entropy-based and keyword plugins, can generate a relatively high number of false positives. When some of our teams are using detect-secrets, they choose to exclude certain plugins (with or without the combination of excluding some files). Currently, if you run a scan with `--no-xxx-scan` option, the used plugin list would be persisted in the baseline file.

If some developer or automation system picks up the repo, have no pre-commit hook setup and also unaware of the exclude list, they could run into the issue that they issue `detect-secrets --update baseline`, then the baseline file is regenerated with all plugins used.

Would the community entertain the idea that `detect-secrets --update baseline` scan use the plugin list from baseline instead of all plugins (default setting)? Some additional options can be added if you want to use more plugins than baseline ones to scan the repo.

We have something implemented in our fork (offline in our GHE), we'd like to hear some feedback on the problem before submitting a big PR.

killuazhu:
CC @jribm

KevinHock:
That certainly sounds reasonable @killuazhu, let me chat with @domanchi about it IRL and I'll update this afterwards. 

The only "Pro" for having `--update` use all the plugins is that if a new plugin was added since the baseline was created, the new plugins are used automatically. We can probably solve this with an `--update-with-all-plugins` and `--update-with-new-plugins` options like you mentioned.

domanchi:
Hey @killuazhu,

That's a really good idea. I think `--update` adding back all the plugins was partially a design choice (so people will get the new plugins), but also partially a bug (because if people already specify the plugins they care about, then we should respect that). After all, that is one of the reasons why we decided to publish a plugins list with the baseline.

We would definitely welcome the addition of a new `--update-all` flag, where this new entry will get all latest plugins, while `--update` only gets the current plugins that are in the baseline!

killuazhu:
Hey @KevinHock , @domanchi thanks for looking into the issue.

So instead of using `--update-all`, how about using a new option called `--use-all-plugins`. It would be easier to align with `detect-secrets-hook`. Somehow the hook command is using `--baseline <baseline>` to update the baseline, instead of `--update` 😕

### The updated behavior would be
* created a new baseline with all plugins
	* `detect-secrets scan`
* created a new baseline with fewer plugins
	* `detect-secrets scan --no-xxx-scan`
* update baseline with detect-secrets with exact plugins in baseline
	* `detect-secrets scan --update baseline`
[...]

PR #124, opened 2019-02-01
Respect options from baseline

killuazhu:
Fully implements the proposal in https://github.com/Yelp/detect-secrets/issues/121#issuecomment-459575817. More than half of the changes are new test cases.

Some logic can be implemented in different places, I chose the current way based on my understanding. Let me know if you have any suggetsions or comments.

FYI @KevinHock @domanchi
CC @jribm

### DIFF
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -291,6 +291,14 @@ class PluginOptions(object):
         return self
 
     @staticmethod
+    def get_disabled_plugins(args):
+        return [
+            plugin.classname
+            for plugin in PluginOptions.all_plugins
+            if plugin.classname not in args.plugins
+        ]
+
+    @staticmethod
     def consolidate_args(args):
         """There are many argument fields related to configuring plugins.
         This function consolidates all of them, and saves the consolidated
@@ -308,8 +316,7 @@ class PluginOptions(object):
             return
 
         active_plugins = {}
-        disabled_plugins = {}
-        param_from_default = {}
+        is_using_default_value = {}
 
         for plugin in PluginOptions.all_plugins:
             arg_name = PluginOptions._convert_flag_text_to_argument_name(
@@ -320,9 +327,6 @@ class PluginOptions(object):
             is_disabled = getattr(args, arg_name, False)
             delattr(args, arg_name)
             if is_disabled:
-                disabled_plugins.update({
-                    plugin.classname: {},
-                })
                 continue
 
             # Consolidate related args
@@ -343,15 +347,14 @@ class PluginOptions(object):
 
                 if default_value and related_args[arg_name] is None:
                     related_args[arg_name] = default_value
-                    param_from_default[arg_name] = True
+                    is_using_default_value[arg_name] = True
 
             active_plugins.update({
                 plugin.classname: related_args,
             })
 
         args.plugins = active_plugins
-        args.disabled_plugins = disabled_plugins
-        args.param_from_default = param_from_default
+        args.is_using_default_value = is_using_default_value
 
     def _add_custom_limits(self):
         high_entropy_help_text = (
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -10,7 +10,7 @@ def add_use_all_plugins_argument(parser):
     parser.add_argument(
         '--use-all-plugins',
         action='store_true',
-        help='Use all available plugins to scan files',
+        help='Use all available plugins to scan files.',
     )
 
 
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -6,6 +6,14 @@ from collections import namedtuple
 from detect_secrets import VERSION
 
 
+def add_use_all_plugins_argument(parser):
+    parser.add_argument(
+        '--use-all-plugins',
+        action='store_true',
+        help='Use all available plugins to scan files',
+    )
+
+
 class ParserBuilder(object):
 
     def __init__(self):
@@ -21,7 +29,8 @@ class ParserBuilder(object):
 
     def add_pre_commit_arguments(self):
         self._add_filenames_argument()\
-            ._add_set_baseline_argument()
+            ._add_set_baseline_argument()\
+            ._add_use_all_plugins_argument()
 
         PluginOptions(self.parser).add_arguments()
 
@@ -78,6 +87,9 @@ class ParserBuilder(object):
         )
         return self
 
+    def _add_use_all_plugins_argument(self):
+        add_use_all_plugins_argument(self.parser)
+
 
 class ScanOptions(object):
 
@@ -121,6 +133,9 @@ class ScanOptions(object):
             help='Update existing baseline by importing settings from it.',
             dest='import_filename',
         )
+
+        # Pairing `--update` with `--use-all-plugins` to overwrite plugins list from baseline
+        add_use_all_plugins_argument(self.parser)
 
         self.parser.add_argument(
             '--all-files',
@@ -293,6 +308,7 @@ class PluginOptions(object):
             return
 
         active_plugins = {}
+        disabled_plugins = {}
 
         for plugin in PluginOptions.all_plugins:
             arg_name = PluginOptions._convert_flag_text_to_argument_name(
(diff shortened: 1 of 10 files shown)


detect_secrets/main.py
```diff
@@ -39,10 +40,12 @@ def main(argv=None):
...
 
         else:
-            baseline_dict = _perform_scan(
-                args,
-                plugins,
-            )
+            if args.import_filename:
+                plugins = initialize.merge_plugin_from_baseline(
+                    _get_plugin_from_baseline(args.import_filename), args,
+                )
```
domanchi: Let's move this block inside `_perform_scan`, right after we get the old_baseline.

e.g.

```
def _perform_scan(args, plugins):
    old_baseline = _get_existing_baseline(args.import_filename)
    if old_baseline:
        plugins = initialize.merge_plugin_from_baseline(...)
```

This way, we don't need to incur two file reads. Also, I'm not entirely sure that reading from `stdin` would work twice.

tests/main_test.py
```diff
@@ -163,10 +164,186 @@ def test_old_baseline_ignored_with_update_flag(
...
+            assert main(
+                shlex.split(
+                    'scan --update old_baseline_file {}'.format(
+                        plugins_overwriten,
+                    ),
+                ),
+            ) == 0
+
+            print("Used:", file_writer.call_args[1]['data']['plugins_used'])
+            print("Wrote:", plugins_wrote)
```
domanchi: Let's remove these `print` statements, before merging.

tests/pre_commit_hook_test.py
```diff
@@ -44,17 +44,53 @@ def test_file_with_secrets(self, mock_log):
     def test_file_no_secrets(self):
         assert_commit_succeeds('test_data/files/file_with_no_secrets.py')
 
-    def test_baseline(self):
+    @pytest.mark.parametrize(
+        ' has_result, use_private_key_scan,, hook_command, commit_result',
```
domanchi: extra comma? maybe `s/commit_result/commit_succeeds`? (also prefixed space, but less fussy)

tests/main_test.py
```diff
@@ -163,10 +164,186 @@ def test_old_baseline_ignored_with_update_flag(
...
+                    {
+                        "base64_limit": 3.5,
+                        "name": "Base64HighEntropyString",
+                    },
+                    {
+                        "name": "PrivateKeyDetector",
+                    },
+                ],
+            ),
+            (  # ignore overwriten option from CLI when not using --use-all-plugins
```
domanchi: Hmm. If you **explicitly** override it by providing the CLI argument, even without `--use-all-plugins`, wouldn't you expect CLI to take precedence over a configured baseline?

Example use case: user wants to see if a decrease in sensitivity results in a large drop in noise.

```
$ detect-secrets scan --update .secrets.baseline --base64-limit 5
$ git diff .secrets.baseline
```

detect_secrets/plugins/common/initialize.py
```diff
@@ -30,6 +30,31 @@ def from_parser_builder(plugins_dict):
...
+    if args.use_all_plugins:
+        plugins = from_parser_builder(args.plugins)
+    elif args.disabled_plugins:  # strip some plugins off baseline
+        plugins = _merge_plugin_from_baseline(baseline_plugins, args)
+    else:
+        plugins = baseline_plugins
+    return plugins
+
+
+def _merge_plugin_from_baseline(baseline_plugins, args):
```
KevinHock: Maybe rename to `_trim_disabled_plugins_from_baseline`, it would make the make the comments in `merge_plugin_from_baseline` less needed.

detect_secrets/plugins/common/initialize.py
```diff
@@ -30,6 +30,31 @@ def from_parser_builder(plugins_dict):
...
+    merged_plugins_dict = {vars(plugin)['name']: plugin for plugin in baseline_plugins}
```
KevinHock: Great use of `vars` 👍

tests/main_test.py
killuazhu: I debated about that, it might have some edge cases. For example, if the baseline does not have the base64 scan, do we ignore the limit? or add the base64 scan in? So I chose to let the user on purposely use the plugin, then allow adjusting. How would you suggest to adjust?

tests/pre_commit_hook_test.py
killuazhu: yep, extra comma. The rename sounds good. I will update in a new commit.

detect_secrets/plugins/common/initialize.py
killuazhu: Credit to @meneal

tests/main_test.py
killuazhu: It's updated to support these cases now. New test cases added.

killuazhu:
@domanchi ready for another round.

detect_secrets/core/usage.py
```diff
@@ -309,6 +309,7 @@ def consolidate_args(args):
 
         active_plugins = {}
         disabled_plugins = {}
```
domanchi: Now that I think about it, what **is** our purpose of returning a set of disabled_plugins? Since `args.plugins` is a collection of all active plugins (combining default values and specifically set values), wouldn't any plugin that isn't in this dictionary be disabled?

detect_secrets/core/usage.py
```diff
@@ -309,6 +309,7 @@ def consolidate_args(args):
...
+        param_from_default = {}
```
domanchi: Maybe `is_default_parameter`, so it's clearer that it contains boolean values?

detect_secrets/core/usage.py
```diff
@@ -309,6 +309,7 @@ def consolidate_args(args):
 
         active_plugins = {}
```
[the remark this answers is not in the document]
killuazhu: `disabled_plugins` is mainly used to filter off some disabled plugins when reading the plugin list from baseline. Relates to the comment below, it's possible to generate the `disabled_plugins` list on-demand by calculating the difference between all plugins and active plugins. I can add a helper function to the `usage.py`.

I was using `disabled_plugins` earlier to avoid re-iterating on the all plugins to get such list.

detect_secrets/core/usage.py
killuazhu: See comment above.

detect_secrets/core/usage.py
killuazhu: That's a good idea. I would make a suggestion to rename it to `is_using_default_value`

killuazhu:
> Thanks for writing that hairy logic of plugin prioritization.

It definitely makes the job more challenge when I have to convert the BasePlugin tuple into a dictionary, then compares and sets values. I noticed the BasePlugin class has made `__dict__` field itself a property instead of the exact properties such as `name`, `base64_limit` and etc. This prevents the `{set|has|get}attr` from reading and writing values. May I ask what was the rationale of overwriting `__dict__` instead of directly using property on each field? @domanchi

detect_secrets/plugins/common/initialize.py
```diff
@@ -32,28 +32,68 @@ def from_parser_builder(plugins_dict):
...
+        for param_name, param_value in list(plugin_params.items()):
+            from_default = args.param_from_default.get(param_name, False)
+            if from_default is False:
+                try:
+                    plugins_dict[plugin_name][param_name] = param_value
+                except KeyError:
+                    log.warning(
+                        '%s specified, but %s not configured! Ignoring...'
+                        % ("".join(["--", param_name.replace("_", "-")]), plugin_name),
+                    )
```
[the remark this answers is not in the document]
killuazhu: It's been updated now. Although having more lines, but I do think the logic is a little bit more clear.

killuazhu:
@domanchi all comments addressed.

detect_secrets/core/usage.py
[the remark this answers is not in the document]
domanchi: Gotcha. Makes sense.

Thanks for adding the helper function anyway!

detect_secrets/plugins/common/initialize.py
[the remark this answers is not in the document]
domanchi: Clearer logic >> more lines, especially when it comes to long term maintenance 😸

PR #127, opened 2019-02-07
Add per line exclude regex via --exclude-line

KevinHock:
So this reverts https://github.com/Yelp/detect-secrets/commit/15a6e6a0fe4884925c15f4378667b2d8f2e16cab in favor of a line exclude arg for all plugins. If at a later date we would like a regex specifically for e.g. `KeywordDetector`, then we can add that feature then.

💥 This removes the `--exclude` CLI arg

:tada: Replace `--exclude with` `--exclude-files` and `--exclude-lines`
    
    Change baseline format to replace `exclude_regex` with
    ```
    "exclude":{
      "files": "foo",
      "lines": "bar"
    }
    ```
    
    Pass `exclude_lines_re` to all plugins in `from_plugin_classname`, and use 'em

### DIFF
--- a/detect_secrets/core/baseline.py
+++ b/detect_secrets/core/baseline.py
@@ -10,8 +10,8 @@ from detect_secrets.core.secrets_collection import SecretsCollection
 
 def initialize(
     plugins,
-    exclude_files_re=None,
-    exclude_lines_re=None,
+    exclude_files_regex=None,
+    exclude_lines_regex=None,
     path='.',
     scan_all_files=False,
 ):
@@ -21,14 +21,18 @@ def initialize(
     :type plugins: tuple of detect_secrets.plugins.base.BasePlugin
     :param plugins: rules to initialize the SecretsCollection with.
 
-    :type exclude_files_re: str|None
-    :type exclude_lines_re: str|None
+    :type exclude_files_regex: str|None
+    :type exclude_lines_regex: str|None
     :type path: str
     :type scan_all_files: bool
 
     :rtype: SecretsCollection
     """
-    output = SecretsCollection(plugins, exclude_files_re)
+    output = SecretsCollection(
+        plugins,
+        exclude_files=exclude_files_regex,
+        exclude_lines=exclude_lines_regex,
+    )
 
     if os.path.isfile(path):
         # This option allows for much easier adhoc usage.
@@ -41,11 +45,11 @@ def initialize(
     if not files_to_scan:
         return output
 
-    if exclude_files_re:
-        exclude_files_re = re.compile(exclude_files_re, re.IGNORECASE)
+    if exclude_files_regex:
+        exclude_files_regex = re.compile(exclude_files_regex, re.IGNORECASE)
         files_to_scan = filter(
             lambda file: (
-                not exclude_files_re.search(file)
+                not exclude_files_regex.search(file)
             ),
             files_to_scan,
         )
@@ -68,13 +72,13 @@ def get_secrets_not_in_baseline(results, baseline):
     :rtype: SecretsCollection
     :returns: SecretsCollection of new results (filtering out baseline)
     """
-    exclude_files_re = None
+    exclude_files_regex = None
     if baseline.exclude_files:
-        exclude_files_re = re.compile(baseline.exclude_files, re.IGNORECASE)
+        exclude_files_regex = re.compile(baseline.exclude_files, re.IGNORECASE)
 
     new_secrets = SecretsCollection()
     for filename in results.data:
-        if exclude_files_re and exclude_files_re.search(filename):
+        if exclude_files_regex and exclude_files_regex.search(filename):
             continue
 
         if filename not in baseline.data:
--- a/detect_secrets/core/baseline.py
+++ b/detect_secrets/core/baseline.py
@@ -10,7 +10,8 @@ from detect_secrets.core.secrets_collection import SecretsCollection
 
 def initialize(
     plugins,
-    exclude_regex=None,
+    exclude_files_re=None,
+    exclude_lines_re=None,
     path='.',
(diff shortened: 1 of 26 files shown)


tests/core/secrets_collection_test.py
```diff
@@ -276,41 +276,80 @@ def setup(self):
...
             ),
+            exclude_files_re='foo',
         )
 
     def test_output(self, mock_gmtime):
-        assert self.logic.format_for_baseline_output() == self.get_baseline_dict(mock_gmtime)
+        assert self.logic.format_for_baseline_output() == self.get_new_baseline_dict(mock_gmtime)
 
-    def test_load_baseline_from_string(self, mock_gmtime):
+    def test_old_load_baseline_from_string(self, mock_gmtime):
```
domanchi: Let's change `old` to something else more version specific.

tests/core/secrets_collection_test.py
```diff
@@ -276,41 +276,80 @@ def setup(self):
...
+        secrets = SecretsCollection.load_baseline_from_string(
+            json.dumps(old_original),
+        ).format_for_baseline_output()
+
+        # exclude_regex got updated to exclude: files
+        assert old_original['exclude_regex'] == secrets['exclude']['files']
+        assert secrets['exclude']['lines'] is None
+        assert old_original['results'] == secrets['results']
+
+    def test_new_load_baseline_from_string(self, mock_gmtime):
```
domanchi: Same thing with `new`.

detect_secrets/core/baseline.py
```diff
@@ -20,13 +21,14 @@ def initialize(
...
 
-    :type exclude_regex: str|None
+    :type exclude_files_re: str|None
+    :type exclude_lines_re: str|None
     :type path: str
     :type scan_all_files: bool
 
     :rtype: SecretsCollection
     """
-    output = SecretsCollection(plugins, exclude_regex)
```
domanchi: Did you also mean to pass in `exclude_lines_re`?

detect_secrets/main.py
```diff
@@ -164,5 +171,30 @@ def _read_from_file(filename):  # pragma: no cover
...
+        return old_baseline['exclude']['files']
+    if old_baseline.get('exclude_regex'):
+        return old_baseline['exclude_regex']
+
+
+def _add_baseline_to_exclude_files(args):
+    """
+    Modifies args.exclude_files in-place.
+    """
+    baseline_name_re = r'^{}$'.format(args.import_filename[0])
```
domanchi: Please change naming convention from `[\w+]_re` to `[\w+]_regex`. It's easier to read, and unless there's a specific reason why the abbreviation is used, we can spare the extra three characters. :)

detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -48,6 +48,11 @@ def __init__(self, charset, limit, *args):
         self.charset = charset
         self.entropy_limit = limit
         self.regex = re.compile(r'([\'"])([%s]+)(\1)' % charset)
+        self.exclude_lines_re = None
+        if exclude_lines_re:
+            self.exclude_lines_re = re.compile(
+                exclude_lines_re,
+            )
```
domanchi: This is a great opportunity to use subclassing and `super()`.

detect_secrets/plugins/common/yaml_file_parser.py
```diff
@@ -38,10 +38,11 @@ class YamlFileParser(object):
     This parsing method is inspired by https://stackoverflow.com/a/13319530.
     """
 
-    def __init__(self, file):
+    def __init__(self, file, exclude_lines_re=None):
```
domanchi: Please specify the type you expect for this. It seems like a regex object, but that can get confusing since we pass in a string in `BasePlugin.__init__`.

tests/plugins/high_entropy_strings_test.py
```diff
@@ -165,7 +173,10 @@ def test_ini_file(self):
         assert count == 7
 
     def test_yaml_file(self):
-        plugin = Base64HighEntropyString(3)
+        plugin = Base64HighEntropyString(
+            base64_limit=3,
+            exclude_lines_re='(CanonicalUser)',
```
domanchi: Huh. Are `()` required?

detect_secrets/main.py
KevinHock: I saw Anthony Sottile use `_re` on a live stream and have been doing it since lol, I'll be sure to change it not here but other places too.

KevinHock:
(took feedback into account and fixed a 🐛  that detect-secrets-server triggers 👍 )

detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -39,7 +39,9 @@ class HighEntropyStringsPlugin(BasePlugin):
 
     __metaclass__ = ABCMeta
 
-    def __init__(self, charset, limit, exclude_lines_re, *args):
+    secret_type = 'High Entropy String'
```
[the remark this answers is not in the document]
KevinHock: @domanchi  I am wondering, why has this test not required us to implement this before?

>`this class cannot (and should not) be constructed directly.`
https://github.com/Yelp/detect-secrets/blob/8e4190fe3603bfdb22147d855a1fd8bbe1c16646/tests/plugins/high_entropy_strings_test.py#L221

from https://github.com/Yelp/detect-secrets/commit/cc2d0411121840dec3b5d34915ad66d8122a148f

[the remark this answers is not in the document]
domanchi: Ah fair. Probably cause we didn't do a `super` call.

You can drop the issue. Probably easier this way, than creating a mock base class.

Issue #71, opened 2018-08-27
[audit functionality] Handle Unicode better


KevinHock:
Someone reported to us the following redacted stacktrace:

```python
...detect_secrets/core/audit.py", line 347, in _highlight_secret
    secret_line[index_of_secret + len(raw_secret):],
UnicodeEncodeError: 'ascii' codec can't encode character u'\xef' in position 51: ordinal not in range(128)
```

KevinHock:
This should have been fixed by @killuazhu's wonderful linked #129 PR :D 🎉 🎂 🎈

PR #129, opened 2019-02-12
fix: unicode error during audit

killuazhu:
Fix an issue when it is Unicode in a file containing secret, the audit would fail with UnicodeEncodeError.

### DIFF
--- a/detect_secrets/core/audit.py
+++ b/detect_secrets/core/audit.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from __future__ import unicode_literals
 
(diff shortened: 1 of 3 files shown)


killuazhu:
I don't know the exact source file which fails https://github.com/Yelp/detect-secrets/issues/71, but I think the fix here would also fix https://github.com/Yelp/detect-secrets/issues/71

KevinHock:
(just waiting for travis, which seems to be having some timeout issues)

KevinHock:
(Bumped version)

Issue #131, opened 2019-02-12
Line exclude regex just for keyword detector


KevinHock:
In https://github.com/Yelp/detect-secrets/pull/127 we added a line exclude for _all_ plugins and reverted https://github.com/Yelp/detect-secrets/commit/15a6e6a0fe4884925c15f4378667b2d8f2e16cab, which was a line exclude regex for just the high-entropy plugins, this was partially due to imagined use-cases of things someone may want to exclude from all plugins e.g. lines whose first non-whitespace character is a`#`, and general cleanliness (DRY).

The keyword detector however, is a young plugin with kind of an ambitious goal, and users shouldn't have to add to the [`FALSE_POSITIVES`](https://github.com/Yelp/detect-secrets/blob/50e2edf46453eacbc1e42144cf51bb0f5ce2618c/detect_secrets/plugins/keyword.py#L49) dict things that may be specific to their codebases, and version bumps won't need to happen to trim false positives.

This is a good-first-issue, b/c it was done very similarly in https://github.com/Yelp/detect-secrets/commit/15a6e6a0fe4884925c15f4378667b2d8f2e16cab

hpandeycodeit:
@KevinHock I would like to work on this.

KevinHock:
Awesome @hpandeycodeit, let me know if you need any help 👍

hpandeycodeit:
@KevinHock Are we looking to change the same files as in this commit [15a6e6a](https://github.com/Yelp/detect-secrets/commit/15a6e6a0fe4884925c15f4378667b2d8f2e16cab)

KevinHock:
@hpandeycodeit  We will still change `detect_secrets/core/usage.py` and `tests/core/usage_test.py`, except instead of changing `HexHighEntropyString` and `Base64HighEntropyString` we can change `KeywordDetector` to have the option. Instead of an `_add_high_entropy_excludes` function we can add something like a `_add_keyword_exclude` function that adds a `--keyword-exclude` option to the `KeywordDetector`.


Instead of the `detect_secrets/plugins/high_entropy_strings.py` and `tests/plugins/high_entropy_strings_test.py` files. We can change `detect_secrets/plugins/keyword.py` and `tests/plugins/keyword_test.py`
e.g. in `keyword.py` we can add an `__init__` method to take in a new arg, then call super's init method. We can use the new arg to skip strings in the `analyze_string_content` method.

Hope this helps 👍 Let me know if you need me to clarify anything

[...]

PR #132, opened 2019-02-20
Changes for Line Exclude regex - Keyword detector

hpandeycodeit:
Issue #131

### DIFF
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -276,7 +276,7 @@ class PluginOptions(object):
             classname='KeywordDetector',
             disable_flag_text='--no-keyword-scan',
             disable_help_text='Disables scanning for secret keywords.',
-            relate_args=[
+            related_args=[
                 ('--keyword-exclude', None),
             ],
         ),
@@ -427,5 +427,5 @@ class PluginOptions(object):
         self.parser.add_argument(
             '--keyword-exclude',
             type=str,
-            help='Pass in regex to exclude false positives found by keyword exclude',
+            help='Pass in regex to exclude false positives found by keyword detector',
         )
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -133,14 +133,16 @@ class ScanOptions(object):
         add_exclude_lines_argument(self.parser)
 
         # Pairing `--exclude-files` with `--scan` because it's only used for the initialization.
-        # The pre-commit hook framework already has an `exclude` option that can be used instead.
+        # The pre-commit hook framework already has an `exclude` option that can
+        # be used instead.
         self.parser.add_argument(
             '--exclude-files',
             type=str,
             help='Pass in regex to specify ignored paths during initialization scan.',
         )
 
-        # Pairing `--update` with `--scan` because it's only used for initialization.
+        # Pairing `--update` with `--scan` because it's only used for
+        # initialization.
         self.parser.add_argument(
             '--update',
             nargs=1,
@@ -149,7 +151,8 @@ class ScanOptions(object):
             dest='import_filename',
         )
 
-        # Pairing `--update` with `--use-all-plugins` to overwrite plugins list from baseline
+        # Pairing `--update` with `--use-all-plugins` to overwrite plugins list
+        # from baseline
         add_use_all_plugins_argument(self.parser)
 
         self.parser.add_argument(
@@ -228,6 +231,7 @@ class PluginDescriptor(namedtuple(
         'related_args',
     ],
 )):
+
     def __new__(cls, related_args=None, **kwargs):
         if not related_args:
             related_args = []
@@ -272,6 +276,9 @@ class PluginOptions(object):
             classname='KeywordDetector',
             disable_flag_text='--no-keyword-scan',
             disable_help_text='Disables scanning for secret keywords.',
+            relate_args=[
+                ('--keyword-exclude', None),
+            ],
         ),
         PluginDescriptor(
             classname='AWSKeyDetector',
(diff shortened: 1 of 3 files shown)


detect_secrets/core/usage.py
```diff
@@ -414,3 +422,10 @@ def _convert_flag_text_to_argument_name(flag_text):
         :return: `no_hex_string_scan`
         """
         return flag_text[2:].replace('-', '_')
+
+    def _add_keyword_exclude(self):
+        self.parser.add_argument(
+            '--keyword-exclude',
+            type=str,
+            help='Pass in regex to exclude false positives found by keyword exclude',
```
KevinHock: Maybe `found by keyword detector.`

detect_secrets/core/usage.py
```diff
@@ -272,6 +276,9 @@ class PluginOptions(object):
             classname='KeywordDetector',
             disable_flag_text='--no-keyword-scan',
             disable_help_text='Disables scanning for secret keywords.',
+            relate_args=[
```
KevinHock: `relate_args` ->`related_args`

detect_secrets/plugins/keyword.py
```diff
@@ -139,19 +139,24 @@ class KeywordDetector(BasePlugin):
...
-            string,
-            filetype=determine_file_type(filename),
-        ):
-            secret = PotentialSecret(
-                self.secret_type,
-                filename,
-                identifier,
-                line_num,
-            )
+        if keyword_exclude and keyword_exclude != string:
```
KevinHock: Before the `for identifier in self.secret_generator` loop we can do
```python
    if (
        self.keyword_exclude
        and self.keyword_exclude.search(string)
    ):
        return output
```
as I think string should always be a single-line.

detect_secrets/plugins/keyword.py
```diff
@@ -139,19 +139,24 @@ class KeywordDetector(BasePlugin):
 
     secret_type = 'Secret Keyword'
 
+    def __init__(self, keyword_exclude=None, **kwargs):
```
KevinHock: I'm sorry I didn't explain this well, we can do

```python
    def __init__(self, keyword_exclude=None, exclude_lines_regex=None, **kwargs):
        super(KeywordDetector, self).__init__(
            exclude_lines_regex,
        )

        self.keyword_exclude = None
        if keyword_exclude:
            self.keyword_exclude = re.compile(
                keyword_exclude,
                re.IGNORECASE,
            )
```

Since `super` will call the base plugin
https://github.com/Yelp/detect-secrets/blob/50e2edf46453eacbc1e42144cf51bb0f5ce2618c/detect_secrets/plugins/base.py#L17
we'll only pass the "all plugins" `exclude_lines_regex` to it.

detect_secrets/core/usage.py
hpandeycodeit: typo :)

detect_secrets/plugins/keyword.py
hpandeycodeit: np. This looks much better. :) 
So does it mean every plugin depends on this base plugin?

detect_secrets/plugins/keyword.py
hpandeycodeit: done

detect_secrets/core/usage.py
hpandeycodeit: done

detect_secrets/core/usage.py
hpandeycodeit: done

detect_secrets/plugins/keyword.py
```diff
@@ -139,19 +139,35 @@ class KeywordDetector(BasePlugin):
...
-            string,
-            filetype=determine_file_type(filename),
-        ):
-            secret = PotentialSecret(
-                self.secret_type,
-                filename,
-                identifier,
-                line_num,
-            )
+            if (
```
KevinHock: I think we can de-dent the added lines.

detect_secrets/plugins/keyword.py
KevinHock: Yup, that's correct, every plugin inherits from the `BasePlugin` e.g. some place will loop through plugins and call `plugin.analyze` and which is `analyze` in the base plugin that calls `analyze_string` in the base plugin that will call `analyze_string_content` in the child plugin (`KeywordDetector` in this case)  https://github.com/Yelp/detect-secrets/blob/50e2edf46453eacbc1e42144cf51bb0f5ce2618c/detect_secrets/plugins/base.py#L31-L72

detect_secrets/plugins/keyword.py
```diff
@@ -153,22 +153,22 @@ def __init__(self, keyword_exclude=None,
...
+            string,
+            filetype=determine_file_type(filename),
+        ):
+            secret = PotentialSecret(
+                self.secret_type,
+                filename,
+                identifier,
+                line_num,
+            )
+        output[secret] = secret
```
KevinHock: `output[secret] = secret` was on the right indentation level, my mistake

hpandeycodeit: Okay. Reverted it.

detect_secrets/plugins/keyword.py
hpandeycodeit: okay. That's much clear now. I am interested if there is more work in this project :)

KevinHock:
https://travis-ci.org/Yelp/detect-secrets/jobs/496360278 shows that a pre-commit hook is failing, I think you can do
pip install pre-commit
pre-commit install
pre-commit run —all-files

To fix the file

KevinHock:
Awesome, I’ll merge in the morning. Thank you 👍

hpandeycodeit:
Thanks @KevinHock! :)

PR #143, opened 2019-03-20
Fix unicode on RHS in ini scan

KevinHock:
#### Changes
- 🐛 Handle unicode in ini_file_parser.py

This bug was trigged only when Unicode was on the RHS e.g.
```
foo=bår
```

&
- merged in master to get coverage back up to the minimum (98%)

#### Why this works
To clarify why `unicode_literals` fixes this, we format things into a string literal when doing `re.compile`, so this [makes sure it's unicode and not bytes](https://stackoverflow.com/questions/23370025/what-is-unicode-literals-used-for).

### DIFF
--- a/detect_secrets/plugins/common/ini_file_parser.py
+++ b/detect_secrets/plugins/common/ini_file_parser.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 import configparser
 import re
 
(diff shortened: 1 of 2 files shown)


calvinli:
To clarify, the problem this fixes is that a Python 2.7 `str` cannot be `.format()`-ed with unicode:

```py
>>> x = u'helló'
>>> r'foo{}'.format(x)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf3' in position 4: ordinal not in range(128)
>>>
```

The pattern must be unicode in order to do this:
```py
>>> x = u'helló'
>>> ur'foo{}'.format(x)
u'foohell\xf3'
>>>
```

PR #147, opened 2019-03-21
Patch backports configparser too

KevinHock:
:bug: Patch backports configparser too

We previously 🐒  patched configparser with `EfficientParsingError`, however this did not patch the backports version present on Python 2.

The [configparser README](https://github.com/jaraco/configparser/#configparser) says `If you’d like to use the backport on Python 2 and the built-in version on Python 3`, then use the invocation that we were using, i.e. `import configparser`. However this doesn't work when monkey-patching, so I did the more explicit `from backports import configparser`.

### DIFF
--- a/detect_secrets/plugins/common/ini_file_parser.py
+++ b/detect_secrets/plugins/common/ini_file_parser.py
@@ -1,6 +1,9 @@
 from __future__ import unicode_literals
 
-import configparser
+try:
+    from backports import configparser
+except ImportError:  # pragma: no cover
+    import configparser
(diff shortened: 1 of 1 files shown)


KevinHock:
Hi @jaraco, just curious, why is it necessary to do a `from` style import here?

jaraco:
It’s just that the backport exists as backports.configparser and that’s the only way to reliably get the backport, because if you import configparser, you’ll get the stdlib version of present or the backport otherwise.
Did that answer the question?

PR #152, opened 2019-04-03
feature: support whitelist for xml

killuazhu:
Better support whitelist for XML. Currently user needs to move the XML comment closing tag --> to 2nd line to properly whitelist. This PR would allow same line whitelisting for xml.

### DIFF
--- a/detect_secrets/plugins/common/constants.py
+++ b/detect_secrets/plugins/common/constants.py
@@ -6,11 +6,12 @@ WHITELIST_REGEXES = [
     for r in [
         r'[ \t]+{} *pragma: ?whitelist[ -]secret{}[ \t]*$'.format(start, end)
         for start, end in (
-            ('#', ''),              # e.g. python or yaml
-            ('//', ''),             # e.g. golang
-            (r'/\*', r' *\*/'),     # e.g. c
-            ('\'', ''),             # e.g. visual basic .net
-            ('--', ''),             # e.g. sql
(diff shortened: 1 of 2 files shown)


detect_secrets/plugins/common/constants.py
```diff
@@ -6,11 +6,12 @@
...
-            ('//', ''),             # e.g. golang
-            (r'/\*', r' *\*/'),     # e.g. c
-            ('\'', ''),             # e.g. visual basic .net
-            ('--', ''),             # e.g. sql
+            ('#', ''),                 # e.g. python or yaml
+            ('//', ''),                # e.g. golang
+            (r'/\*', r' *\*/'),        # e.g. c
+            ('\'', ''),                # e.g. visual basic .net
+            ('--', ''),                # e.g. sql
+            ('<!--[# \t]*', ' *-->'),  # e.g. xml
```
KevinHock: Can you make this regex [lazy](https://stackoverflow.com/questions/2301285/what-do-lazy-and-greedy-mean-in-the-context-of-regular-expressions)  for performance reasons, I think `*` to `*?` in this case, and use the `r'` prefix just to denote that it is also a regex.

killuazhu: @KevinHock sorry for the delay. The comment should be addressed now.

KevinHock: No worries, looks great to me 👍

Issue #156, opened 2019-04-08
Add plugin for an Artifactory credential detector


justineyster:
Artifactory credentials are a common type of secret that is leaked. We developed a plugin for `detect-secrets` to address this.

justineyster:
Closed via #157.

PR #157, opened 2019-04-08
Adds artifactory credential detector to plugins

justineyster:
Supports #156

### DIFF
--- a/detect_secrets/plugins/common/initialize.py
+++ b/detect_secrets/plugins/common/initialize.py
@@ -4,6 +4,7 @@ try:
 except ImportError:  # pragma: no cover
     from functools32 import lru_cache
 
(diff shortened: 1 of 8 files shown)


detect_secrets/plugins/common/initialize.py
```diff
@@ -12,6 +12,7 @@
 from ..keyword import KeywordDetector                       # noqa: F401
 from ..private_key import PrivateKeyDetector                # noqa: F401
 from ..slack import SlackDetector                           # noqa: F401
+from ..artifactory import ArtifactoryDetector               # noqa: F401
```
KevinHock: Nit: can you put this on line 7, so they're alphabetical

justineyster: @KevinHock nit addressed!

PR #160, opened 2019-04-11
More readable regex in keyword.py

baboateng:
Simplify the regex by using named string replacement


KevinHock:
(Not merging due to merge conflict, but code looks great to me.)

baboateng:
Thanks @KevinHock, happy to contribute to such a great project 👍, fixed the conflicts

KevinHock:
re: Keyword detector, if you've used it, I'm curious to hear your thoughts about it. I was planning on making `quotes required` the default, and adding exceptions for `no quotes required`, rather than how it is now.

Issue #117, opened 2019-01-11
`pragma: whitelist secret` doesn't support additional text


domanchi:
`detect-secrets` doesn't complain about:

```
whitelisted_api_key: "DEADBEEF1234" # pragma: whitelist secret
```

but complains about:

```
whitelisted_api_key: "DEADBEEF1234" # pragma: whitelist secret blah
```

If it's a comment, it should support text that follows it.

asdrubalos:
Help me please.
how write # pragma: whitelist secret in java?

I have 2 false positive:
1)
this.clientSecret = clientSecret; //# pragma: whitelist secret
2)
String x =", clientSecret='" + this.clientSecret +"'"; //# pragma: whitelist secret

KevinHock:
Hi @asdrubalos, I believe you can put a space after the `//`

When I scan
```java
$ cat test.java 
API_KEY = "blahwww"
API_KEY = "blah-blah-but-actually-not-secret" # pragma: whitelist secret
1)
this.clientSecret = clientSecret; // # pragma: whitelist secret
2)
String x =", clientSecret='" + this.clientSecret +"'"; //# pragma: whitelist secret
```
with the latest detect-secrets (0.12.0) `this.clientSecret = clientSecret;` was reported if I did not have a space between `//` and `#`

KevinHock:
Also @asdrubalos , I will be sure to make quotes required for keyword secrets in `.java` files https://github.com/Yelp/detect-secrets/pull/133 

Thank you for the report 👍

KevinHock:
Actually just `//` should work as well, e.g. `// pragma: whitelist secret`, this is due to this PR https://github.com/Yelp/detect-secrets/pull/105/files

We should document that better though 👍  We'll be sure to do that too

KevinHock:
This was solved by the wonderful @dgzlopes

PR #168, opened 2019-05-09
Add support for text after whitelist

dgzlopes:

Minor fix for #117. Added a couple of testing cases too!

### DIFF
--- a/detect_secrets/plugins/common/constants.py
+++ b/detect_secrets/plugins/common/constants.py
@@ -4,7 +4,7 @@ import re
 WHITELIST_REGEXES = [
     re.compile(r)
     for r in [
-        r'[ \t]+{} *pragma: ?whitelist[ -]secret{}[ \t]*$'.format(start, end)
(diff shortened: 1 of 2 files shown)


detect_secrets/plugins/common/constants.py
```diff
@@ -4,7 +4,7 @@
 WHITELIST_REGEXES = [
     re.compile(r)
     for r in [
-        r'[ \t]+{} *pragma: ?whitelist[ -]secret{}[ \t]*$'.format(start, end)
+        r'[ \t]+{} *pragma: ?whitelist[ -]secret.*{}[ \t]*$'.format(start, end)
```
KevinHock: For performance reasons, we can make this regex [lazy](https://stackoverflow.com/questions/2301285/what-do-lazy-and-greedy-mean-in-the-context-of-regular-expressions), e.g. `.*?` in this case, (I think.)

dgzlopes: Yep! It's better to search for the shortest match. Fixing it now.

Issue #159, opened 2019-04-10
Review and improve regex rules


domanchi:
There was a recent white paper released ([summary](https://blog.acolyer.org/2019/04/08/how-bad-can-it-git-characterizing-secret-leakage-in-public-github-repositories/), [source](https://www.ndss-symposium.org/wp-content/uploads/2019/02/ndss2019_04B-3_Meli_paper.pdf)).

What's most interesting is on page 15, they list a variety of explicit regexes that we may be able to incorporate into our scanning. I think we already cover like 80% (mostly with the high entropy scanner), but there are some interesting ones to extract from that. e.g.:

* finance related tokens
* Facebook access tokens

We should go through this list and create new plugins for the ones that we're missing.

killuazhu:
I love the idea. Be able to more deterministically identify the type of the token can also support #153

domanchi:
A couple of notes from this paper worth mentioning (for posterity):

- **Section III, Part E**: talks about some interesting ideas on how to better filter out junk keys (e.g. `XXXX`, has `EXAMPLE` in the text)

- **Section V, Part D**: notes that multi-factor secrets (e.g. username and password) has an 80% chance that they both can often be found within 5 lines of context, before and after a secret.

- **Section VII, Part D**: entropy checks still catch more than just regex rules. This is good to know, and allows users to decide how conservative they want to be (accuracy v recall trade-off).

KevinHock:
I thought this part was another cool thing to experiment with:

**Section III, Part D**:

>Note that each regular expression was prefixed with negative lookbehind `(?<![\w])` and suffixed with negative lookahead `(?![\w])` to ensure that no word characters appeared before or after the regular expression match and improve accuracy.

PR #169, opened 2019-05-10
Add Stripe detector to plugins

dgzlopes:
From paper cited on #159. Supports both standard and restricted keys.

### DIFF
--- a/detect_secrets/plugins/stripe.py
+++ b/detect_secrets/plugins/stripe.py
@@ -13,6 +13,6 @@ class StripeDetector(RegexBasedDetector):
     secret_type = 'Stripe Access Key'
 
     blacklist = (
-            # stripe standard keys begin with sk_live and restricted with rk_live
-            re.compile(r'(r|s)k_live_[0-9a-zA-Z]{24}'),
+        # stripe standard keys begin with sk_live and restricted with rk_live
+        re.compile(r'(r|s)k_live_[0-9a-zA-Z]{24}'),
     )
--- a/detect_secrets/plugins/stripe.py
+++ b/detect_secrets/plugins/stripe.py
@@ -1,0 +1,18 @@
+"""
+This plugin searches for Stripe keys
+"""
+from __future__ import absolute_import
+
+import re
+
+from .base import RegexBasedDetector
+
+
+class StripeDetector(RegexBasedDetector):
+
+    secret_type = 'Stripe Access Key'
+
+    blacklist = (
+            # stripe standard keys begin with sk_live and restricted with rk_live
+            re.compile(r'(r|s)k_live_[0-9a-zA-Z]{24}'),
+    )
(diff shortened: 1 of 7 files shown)


detect_secrets/plugins/stripe.py
```diff
@@ -0,0 +1,18 @@
...
+
+from .base import RegexBasedDetector
+
+
+class StripeDetector(RegexBasedDetector):
+
+    secret_type = 'Stripe Access Key'
+
+    blacklist = (
+        # stripe standard keys begin with sk_live and restricted with rk_live
```
KevinHock: This looks good to me from reading online [[1](https://wordpress.org/support/topic/restricted-api-keys-arent-accepted/)][[2](https://docs.woocommerce.com/document/stripe/#section-42)][[3](https://sportical.zendesk.com/hc/en-us/articles/217384808-Setting-up-Credit-Card-Processing-with-Stripe)].

Hi @andrew-d @richo, can you confirm this is accurate?

tests/plugins/stripe_key_test.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import pytest
+
+from detect_secrets.plugins.stripe import StripeDetector
+from testing.mocks import mock_file_object
+
+
+class TestAWSKeyDetector(object):
```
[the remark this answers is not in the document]
dgzlopes: Missed that one! Should be fixed now.

detect_secrets/plugins/stripe.py
richo: I left Stripe in December, so feel free to ignore me ;)

This is accurate in the abstract, depending on what your goals here are (find exploitable credentials vs detect organisational deviation from best practice) you plausibly want to look for `sk_test` and `rk_test` as well though. And while I'm opining and since my opinions can no longer be pinned on Stripe, I'm still 100% willing to die on the hill of prefering denylist/allowlist over blacklist/whitelist. lmk if you'd be interested in a PR for that :)

KevinHock: Thanks for the info, that's a great point! I'd happily accept a PR for that 👍

richo:
Thanks! I'll put something together.

detect_secrets/plugins/stripe.py
```diff
@@ -0,0 +1,18 @@
...
+import re
+
+from .base import RegexBasedDetector
+
+
+class StripeDetector(RegexBasedDetector):
+
+    secret_type = 'Stripe Access Key'
+
+    blacklist = (
```
KevinHock: `blacklist` -> `denylist`

KevinHock: Resolving since that will require a large refactor we have in the pipeline

Issue #151, opened 2019-03-26
Use --exclude-lines from baseline, if there


KevinHock:
We [somewhat recently](https://github.com/Yelp/detect-secrets/blob/master/CHANGELOG.md#tada-new-features-1) added an `--exclude-lines` option when scanning, that will end up in the baseline here:
```
  "exclude": {
    "files": "...",
    "lines": "(foo)"
  },
```

We might as well not make people do e.g.

```
    -   id: detect-secrets
        args: ['--baseline', '.secrets.baseline', '--exclude-lines', '(foo)']
```
in their `.pre-commit-config.yaml`, if they already have a line exclude in their baseline.

domanchi:
This is handled in the upcoming v1 release through filters (https://github.com/Yelp/detect-secrets/blob/pre-v1-launch/docs/filters.md). Once the baseline is loaded, all settings are also configured, which carry over any line/file exclusions, as well as other filters (e.g. wordlist, verifiability, etc.)

Closing as done.

Issue #148, opened 2019-03-22
Should 'secret_key' be added to the keyword plugin?


mdscruggs:
While trying out detect-secrets for the first time today, I noticed the term 'secret_key' is not present in the keyword plugin (see https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/keyword.py#L38). This means it doesn't find the very common Django `SECRET_KEY` variable. I was wondering if folks think 'secret_key' should be added to the keyword plugin's blacklist...but as a newcomer to this library I wasn't sure if that would cause consternation, since it would basically point out `SECRET_KEY` for any and all Django projects (if I understand the plugin correctly).

Just looking to start the conversation! Thanks!

KevinHock:
Hi @mdscruggs, that's a good idea 👍 

I'd have to do a decent amount of internal testing first, before adding that to the keyword blacklist, but it definitely sounds reasonable. I'll leave this open until I add it/do the internal testing.

Tioborto:
Interessed about it 👍

lorenzodb1:
We're going to close this issue as it hasn't received any update in a very long time. Feel free to re-open it if you think it's still relevant.

Issue #146, opened 2019-03-21
Add keyword exclude to baseline


KevinHock:
In https://github.com/Yelp/detect-secrets/pull/132 a keyword exclude option was added, but we didn't write what it was to the baseline. This is problematic because e.g. `detect-secrets scan --update .secrets.baseline` will re-scan and never use the keyword exclude.

I realized this before releasing [0.12.1](https://github.com/Yelp/detect-secrets/blob/master/CHANGELOG.md#0121) but thought it was pretty low priority, as it is mostly used to enable us to not have to bump `detect-secrets` and `detect-secrets-server` versions to make additions to the [`FALSE_POSITIVES` dictionary](https://github.com/Yelp/detect-secrets/blob/d2c8290dac0282c0fb4aaa96dfbd88d988076bea/detect_secrets/plugins/keyword.py#L49-L109) in the keyword plugin, and sync it upstream asynchronously.

PR #174, opened 2019-05-13
feat: add one more keyword

killuazhu:
Add one more keyword `token` to the keyword list.

### DIFF
--- a/detect_secrets/plugins/keyword.py
+++ b/detect_secrets/plugins/keyword.py
@@ -45,6 +45,7 @@ BLACKLIST = (
     'private_key',
     'secret',
     'secrete',
+    'token',
 )
 FALSE_POSITIVES = {
     '""',
--- a/tests/plugins/keyword_test.py
+++ b/tests/plugins/keyword_test.py
@@ -18,6 +18,7 @@ FOLLOWED_BY_COLON_RE = {
         "quotes_required": [
             'theapikey: ""',  # Nothing in the quotes
             'theapikey: "somefakekey"',  # 'fake' in the secret
+            'token: "somefaketoken"',  # 'fake' in the secret
         ],
         "quotes_not_required": [
             'theapikeyforfoo:hopenobodyfindsthisone',  # Characters between apikey and :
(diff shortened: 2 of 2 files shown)


killuazhu:
Looks like we haven't released a new version for awhile. Once we have recent PRs being merged, would we be able to generate a new release?

KevinHock:
Sure thing, been meaning too :)

killuazhu:
@KevinHock another thing, I sent you an email at your Github profile email kevin.hock.opentoall@gmail.com, once you get chance would you please take a look? It's coming from my Github profile email kyleezhu@gmail.com. Thanks.

KevinHock:
Will do 👍

detect_secrets/plugins/keyword.py
```diff
@@ -45,6 +45,7 @@
     'private_key',
     'secret',
     'secrete',
+    'token',
```
KevinHock: I am kind of ambivalent about this, when we [first wrote the keyword detector, `token` was a keyword](https://github.com/Yelp/detect-secrets/commit/d912ced2e7a7607f06794e3d916b10e52536bd5c#diff-1b8ec3b1cef86dbae9081fdea381e7fbL40), however the signal-to-noise ratio wasn't as good as the other keywords, it was too noisy.

Inflexibility is an anti-pattern though, which is one of the reasons we added [a keyword exclude regex](https://github.com/Yelp/detect-secrets/pull/132), so the static `FALSE_POSITIVES` dictionary could be added to and  customized for things specific to each users' codebase.

We should probably do the same thing for keyword detector tokens, though it is tricky work, because we will have to write and read it from the baseline file. #146 and #151 are examples of issues stemming from work like this.

This would also not require other keyword additions like #148 to be blocked on us doing a ton of internal testing.

killuazhu: I'm wondering what kind of internal testing do you guys carry to accept / reject a new keyword? We haven't built up a large test code base yet, which we probably should. I'm curious how big is your test code base, and what's process for you to evaluate false positive ratio?

KevinHock: At least for the keyword detector, we made the regexes super loose so it would be noisy, found all the true positives stemming from it, and trimmed the regexes down best we could to catch as little false-positives as possible. A decent amount of manual work was involved with this.

It's still a little loud now, I plan on working on it a little more, but have been doing some other projects recently.

For testing, we ran it on all of our largest codebases, I'm not sure of the exact LoC, but probably a few hundred thousand or more.

p.s. [Bumped ✊ the version today, btw 👍](https://github.com/Yelp/detect-secrets/commit/4e9555a60e9444d5153b90fe9d0d6b688e14e414?short_path=4ac32a7#diff-4ac32a78649ca5bdd8e0ba38b7006a1e)

killuazhu:
I will close this PR since `token` is considered as a noisy keyword.

Issue #141, opened 2019-03-19
Handle un-scannable files more gracefully


KevinHock:
Having a bunch of
`INFO: Checking file: some_image.png`
`WARNING: some_image.png failed to load.`
from
https://github.com/Yelp/detect-secrets/blob/9f3d9ee1d186ba8f5a9972540abde580fc3c570a/detect_secrets/core/secrets_collection.py#L310
is not ideal. We know we cannot scan certain files e.g. images, so we should behave more gracefully.

In [`detect-secrets-server`](https://github.com/Yelp/detect-secrets-server) we already have the [`IGNORED_FILE_EXTENSIONS`](https://github.com/Yelp/detect-secrets-server/blob/c7cee5b2814d589d6643376fcfd6023cad1f2039/detect_secrets_server/constants.py#L7 ) tuple we made [to skip files](https://github.com/Yelp/detect-secrets-server/blob/1905e80acf8cdb6318ee5ad416a0e3724755f1c7/detect_secrets_server/storage/core/git.py#L148) like this
```
IGNORED_FILE_EXTENSIONS = (
    '7z',
    'bmp',
    'bz2',
    'dmg',
    'exe',
    'gif',
    'gz',
    'ico',
    'jar',
    'jpg',
    'jpeg',
    'png',
    'rar',
    'realm',
    's7z',
    'tar',
    'tif',
    'tiff',
    'webp',
    'zip',
)
```
maybe we should move it to `detect-secrets`, change it to a `dict`, and use it.

dgzlopes:
Hello! I don't understand why the IGNORED_FILE_EXTENSIONS structure should change from a tuple to a dict. Could you offer some more insight on this?

On the other side, I made this quick fix https://github.com/dgzlopes/detect-secrets/commit/3828be91030c2243b11427649b875912470f2638 on my fork (handling the check inside scan_file() instead of _extract_secrets_from_file()) and seems to work ok. Any idea on where to handle it more gently?

KevinHock:
The reason why I would like to change it to a `dict` is because the only use is an `in` check, which is faster in a hash table (dictionary). It's less for performance reasons and more for knowing how it is used, from seeing the type.

dgzlopes:
Oh! I see. On the other hand (and sorry if this question is stupid) why we can't use a Set()? 

[...]

Issue #167, opened 2019-05-09
tests/core/audit_test.py is flaky on py35


domanchi:
See this [travis job](https://travis-ci.org/Yelp/detect-secrets/jobs/530362328) for more details.

Looks like this test is non-deterministic, and we should fix it -- but the simple fix now is just to kick off that test again.

domanchi:
v1 deprecates the use of py35. Closing as "Won't Fix".

PR #176, opened 2019-05-14
Improve handle of un-scannable files

dgzlopes:

Now using Set data type as said on #141.

### DIFF
--- a/detect_secrets/core/constants.py
+++ b/detect_secrets/core/constants.py
@@ -1,0 +1,28 @@
+# We don't scan files with these extensions.
+# NOTE: We might be able to do this better with
+#       `subprocess.check_output(['file', filename])`
+#       and look for "ASCII text", but that might be more expensive.
+#
+#       Definitely something to look into, if this list gets unruly long.
+IGNORED_FILE_EXTENSIONS = {
+    '7z',
+    'bmp',
+    'bz2',
+    'dmg',
+    'exe',
+    'gif',
+    'gz',
+    'ico',
+    'jar',
+    'jpg',
+    'jpeg',
+    'png',
+    'rar',
+    'realm',
+    's7z',
+    'tar',
+    'tif',
+    'tiff',
+    'webp',
+    'zip',
+}
--- a/detect_secrets/core/secrets_collection.py
+++ b/detect_secrets/core/secrets_collection.py
@@ -8,6 +8,7 @@ from time import gmtime
 from time import strftime
 
 from detect_secrets import VERSION
+from detect_secrets.core.constants import IGNORED_FILE_EXTENSIONS
 from detect_secrets.core.log import log
 from detect_secrets.core.potential_secret import PotentialSecret
 from detect_secrets.plugins.common import initialize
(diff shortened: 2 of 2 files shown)


detect_secrets/core/constants.py
```diff
@@ -0,0 +1,28 @@
...
+    'png',
+    'rar',
+    'realm',
+    's7z',
+    'tar',
+    'tif',
+    'tiff',
+    'webp',
+    'zip',
+}
```
KevinHock: Nit: Can you add a newline to the end of this file? It is causing the `end-of-file-fixer` pre-commit hook to fail

detect_secrets/core/secrets_collection.py
```diff
@@ -200,7 +201,8 @@ def scan_file(self, filename, filename_key=None):
 
         if os.path.islink(filename):
             return False
-
+        if (os.path.splitext(filename)[1] in IGNORED_FILE_EXTENSIONS):
```
KevinHock: (Just writing this down for posterity.)

We could fill two needs with one deed if we put this in `_results_accumulator` 
i.e.
```python
if os.path.splitext(filename)[1] in IGNORED_FILE_EXTENSIONS:
    return
```
This is because for regular `detect-secrets`, the control flow goes `scan_file`->`_extract_secrets_from_file`->`_results_accumulator`. And for `detect-secrets-server` the control flow goes `scan_diff`->`_results_accumulator`.

However, **let's leave the location as-is for this PR, it's better** 😀  👍 . The current location where we [filter out ignored file extensions](https://github.com/Yelp/detect-secrets-server/blob/1905e80acf8cdb6318ee5ad416a0e3724755f1c7/detect_secrets_server/storage/core/git.py#L96) in `detect-secrets-server`, is a lot earlier on, (before `scan_diff` ever gets called) and better performance-wise. So it's not worth it to make it super DRY at the cost of performance IMO.

detect_secrets/core/constants.py
dgzlopes: Fixed!

detect_secrets/core/secrets_collection.py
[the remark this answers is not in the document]
dgzlopes: Fixed!

dgzlopes:
Sorry for missing the style! Squashed my commits for cleaning the PR. Also, the test failing seems to be the flaky one #167.

KevinHock:
++, thanks :)

Re-running Travis now, will merge when it passes

detect_secrets/core/secrets_collection.py
```diff
@@ -200,7 +201,8 @@ def scan_file(self, filename, filename_key=None):
 
         if os.path.islink(filename):
             return False
-
+        if os.path.splitext(filename)[1] in IGNORED_FILE_EXTENSIONS:
```
KevinHock: This turned out to be a sneaky bug I found today, `splitext` will always have `.extension` in it, and so we were checking if e.g. `.svg` == `svg`. I'll push a fix 👍

KevinHock: Hmm, we did the same thing [in -server](https://github.com/Yelp/detect-secrets-server/blob/618b251919f6834deab9c78ffeb4a506516e79cc/detect_secrets_server/storage/core/git.py#L156) too

PR #178, opened 2019-05-16
Use allowlist and denylist

richo:
Following on from #169 I updated the project to use allowlist and denylist, and got the tests passing.

However, this will basically break every pragma in the wild, so I would guess that you want to support them as a legacy situation for a while? Let me know what you'd like to do, and I'll see about implementing something.

Thanks!


KevinHock:
Thanks for making this! It looks great.

re: pragma's
Good catch, we can support them for now (in `constants.py`) for legacy reasons, I can remove support in a future major version bump.

richo:
Woo! Thanks for getting this across the line!

Issue #130, opened 2019-02-12
Create performance testing benchmarks


domanchi:
This tool is meant to be fast. Not blazingly fast, but fast nonetheless. Before we can optimize on speed though, we need to create some testing frameworks to accurately measure performance of the engine.

This way, we can make improvements, and note the speed differential for different regexes / features.

KevinHock:
It would also be really nice to maybe have an option of running the tool and having it say which files took the longest to scan.

domanchi:
This is finalized in: https://github.com/Yelp/detect-secrets/pull/187.

Issue #136, opened 2019-02-28
Slow performance when scanning a non ini file with millions of lines


killuazhu:
When scanning a non-ini file with more than 1 million lines, it would hang at [line below](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/high_entropy_strings.py#L63). 

```python
(self._analyze_ini_file(add_header=True), configparser.Error,),
```

I'm able to trace back to configParse and found the [following line](https://github.com/python/cpython/blob/3.5/Lib/configparser.py#L335) is extremely inefficient to add all offending lines (essentially all the lines in the file) into the error message with string concatenation.

```python
self.message += '\n\t[line %2d]: %s' % (lineno, line)
```

I did not have the patience to wait for the scan to finish, on my laptop it did hang for at least more than 10 minutes.

We need a more efficient way to scan large non-ini file.

KevinHock:
I ran into this today as well, with a file that was ~250k lines.

domanchi:
I spoke with @KevinHock today about this, and decided to record conversation down, for posterity.

### Historical Context

Initially, the ini parser was written in order to try and catch secrets that did not need quote marks around them -- namely, config files.

```
$ cat config.ini
[private]
key=secret
```

The issue is that there's no easy way to identify whether a file is a config file. File extensions don't work, because config files don't have a typical set of extensions that they correspond to. And, there's no special header file that identifies that a file is a config file. e.g. It's not like you could do:

```
$ file config.ini
```

Therefore, the only way to really identify whether a file is a config file is to try and parse it, and handle errors appropriately.

### Issue

It seems that this approach runs into two performance hits:

1. Needing to parse the entire file, with `configparser`, before having usable results.
2. Error traceback construction for large files takes a long time (as @killuazhu pointed out)

### Possible Solutions

#### 1. Use the first N lines to try and determine whether a file is actually a config file

Credit to @KevinHock for this idea. Essentially, if the following conditions hold true, we may be able to identify whether a file is a config file by reading the first few lines.

a. The first N lines are a representative sample for the entire file, **and**
b. The first N lines are independently parseable as a config file by themselves.

If we're able to do this, then we would be able to optimize on both issues listed above, since you don't need to parse the entire file to determine whether a given file is suitable for ini file parsing.

Our issue is that we don't have a large enough sample set of config files to test out this method.

#### 2. Try to use a different library for config file parsing

[...]

PR #181, opened 2019-05-20
Add performance benchmark

dgzlopes:

Added performance benchmark for the scan stage as said on #130. Is the startTimeMeasurement() on the proper place?

On the other side, I had to use an external monotonic [[0](https://github.com/atdt/monotonic)] package as monotonic it's only avaliable on Python 3.3+ [[1](https://docs.python.org/3/library/time.html)].

[[0](https://github.com/atdt/monotonic)] https://github.com/atdt/monotonic
[[1](https://docs.python.org/3/library/time.html)] https://docs.python.org/3/library/time.html

### DIFF
--- a/scripts/bench.py
+++ b/scripts/bench.py
@@ -1,0 +1,52 @@
+#!/usr/bin/python
+import argparse
+import subprocess
+
+from monotonic import monotonic
+
+from detect_secrets.core.usage import PluginOptions
+
+
+def get_arguments():
+    parser = argparse.ArgumentParser(description='Run some benchmarks.')
+    parser.add_argument(
+        'rest_args',
+        nargs=argparse.REMAINDER,
+        help='filenames to check or detect-secrets compatible arguments',
+    )
+    args = parser.parse_args()
+    return ['../.'] if len(args.rest_args) == 0 else args.rest_args
+
+
+def run_bench(flags, arguments):
+    start_time = monotonic()
+    subprocess.check_output('detect-secrets scan'.split() + arguments + flags)
+    return monotonic() - start_time
+
+
+def print_line(name, time):
+    print('{:<20s}{:>20s}s'.format(name, str(time)))
+
+
+def main():
+    flag_list = list()
+    additional_arguments = get_arguments()
+
+    # Fill flags list with PluginOptions flag text
+    for flag_number, flag in enumerate(PluginOptions.all_plugins):
+        flag_list.append(flag.disable_flag_text)
+
+    # Run bechmarks for all the cases
+    for flag_number, flag in enumerate(flag_list):
+        temp = list(flag_list)
+        temp.pop(flag_number)
+        if flag_number == 0:
+            print('Scanning: ' + str(additional_arguments))
+            print('-' * 42)
+            print('{:<20s}{:>20s}'.format('benchmark', 'time'))
+            print('-' * 42)
+            print_line('all-plugins', run_bench([], additional_arguments))
+        print_line(flag[len('--no-'):], run_bench(temp, additional_arguments))
+
+
+main()
--- a/detect_secrets/main.py
+++ b/detect_secrets/main.py
@@ -115,6 +115,7 @@ def _perform_scan(args, plugins):
 
     :rtype: dict
     """
+
     old_baseline = _get_existing_baseline(args.import_filename)
     if old_baseline:
         plugins = initialize.merge_plugin_from_baseline(
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,7 @@
 coverage
 flake8==3.5.0
 mock
+monotonic
 pre-commit==1.11.2
 pytest
 pyyaml


domanchi:
@dgzlopes Is there a reason why this needs to be integrated into the main package?

I'm thinking an adhoc script should suffice, because this is more for package testing, rather than end user consumption.

KevinHock:
Makes sense, I'd agree with that.

dgzlopes:
@domanchi I thought that maybe integrating the base benchmarking functionality, would make easier to add in the future the per detector benchmarks proposed in https://github.com/Yelp/detect-secrets/issues/130#issuecomment-473468049. Also for the users, may be interesting to know the time elapsed if they run detect-secrets on an automated way.

On the other hand, following the ad-hoc way I understand that It would be a small script that starts the timer, runs detect-secrets and then returns the elapsed time. But if we just want this, maybe it is better to follow the Unix principles and rely on the time command [[0](https://linux.die.net/man/1/time)] (or in windows the Measure-Command) so there is no need for some project-specific script.

[[0](https://linux.die.net/man/1/time)] https://linux.die.net/man/1/time

domanchi:
We could do that too.

When I think of performance bench-marking, I think of the following questions:

- Which file takes the longest to run?
- Which plugin is the slowest?
- Have these timings changed drastically after my changes?
- What does a "good" timing look like?

There's no caching, so speed should remain relatively consistent between consecutive runs. Furthermore, because this tool is somewhat fast, I **doubt** there will be much information if we do a complete file speed breakdown (because the majority of files are small, and therefore, won't produce any reasonable timings to work with).

With these questions in mind, these are the objectives I envision for minimal performance bench-marking:

- [ ] Able to measure how long a single run takes, per plugin
- [ ] Able to customize which files to test (so we can easily test for [known edge cases](https://github.com/Yelp/detect-secrets/issues/136))
- [ ] Able to integrate in "manual" testing flow for other developers (automated tests won't work due to its heuristic nature)

`scripts/` would be a good place for putting this.

With these objectives, we could do the `time` command (and write shell script to iterate over [all plugins](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/usage.py#L248-L303), and disable all of them in turn), or we can still use this library in a more adhoc manner (without introducing it into the core code).

Apologies for not being clearer in the original issue. Hopefully this provides more direction for this added testing capabilities. Thanks for helping empower the community to be able to make better changes by having a scientific way to measure the impact on performance!

dgzlopes:
First, thanks for the time and the nice writeup @domanchi ! Now I know I was missing the point with the first PR. I have updated the PR today adding a new benchmark script (and cleaned the past additions) and would love If you have some time for giving it a spin. I want to note that this is just an update and It's still a WIP (missing some testing, comments and I'm sure that more things!) and It has just some basic functionality included:

- Running just `python bech.py` runs the benchmark inside de detect-secrets project root. Any detect-secrets compatible argument is accepted (from files to flags) too e.g. `python bech.py ../../powerfulseal --all-files` or `python bech.py myfilename.js`.
- It's able to benchmark with all plugins at the same time. Also, It's able to benchmark one plugin at a time (with the other ones excluded).
- Takes the PluginOptions disable_flag_text value from core.usage so we don't have to maintain another list of plugins.
- Uses monotonic for the elapsed time.
- Works with all the Python versions included in tox.ini! But as I note I don't really like the subprocess32 dependency  (even if it's just on the developer environment) so this might change.
- As an example, this is the output for `python bench.py`:

```
vagrant@ubuntu-xenial:~/detect-secrets/scripts$ python bench.py
Scanning: ['../.']
------------------------------------------
benchmark                           time
------------------------------------------
all-plugins                  3.779828449s
hex-string-scan              1.293300869s
base64-string-scan           1.832755461s
private-key-scan          0.987097253001s
basic-auth-scan              0.772286352s
keyword-scan                 1.026593331s
aws-key-scan                 0.800705507s
slack-scan                   0.721169404s
```

scripts/bench.py
```diff
@@ -0,0 +1,55 @@
...
+
+from detect_secrets.core.usage import PluginOptions
+
+if os.name == 'posix' and sys.version_info[0] < 3:
+    import subprocess32 as subprocess
+else:
+    import subprocess
+
+
+def getArguments():
```
domanchi: Please keep the style consistent for function naming (`snake_case` rather than `camelCase`).

scripts/bench.py
```diff
@@ -0,0 +1,55 @@
...
+    startTime = monotonic()
+    subprocess.call('detect-secrets scan'.split() + arguments + flags, stdout=subprocess.DEVNULL)
+    return monotonic() - startTime
+
+
+def printLine(name, time):
+    print('{:<20s}{:>20s}s'.format(name, str(time)))
+
+
+flags = list()
```
domanchi: Even for one-off scripts like this, it's always better to put this under a `main` function. See https://stackoverflow.com/questions/11241523/why-does-python-code-run-faster-in-a-function for a good breakdown.

scripts/bench.py
```diff
@@ -0,0 +1,55 @@
...
+aditionalArguments = getArguments()
+
+# Fill flags list with PluginOptions flag text
+for c, value in enumerate(PluginOptions.all_plugins, 0):
```
domanchi: Don't need to specify the `0`, because that's by default.

scripts/bench.py
```diff
@@ -0,0 +1,55 @@
...
+    temp = list(flags)
+    temp.pop(c)
+    if c == 0:
+        print('Scanning: ' + str(aditionalArguments))
+        print('-' * 42)
+        print('{:<20s}{:>20s}'.format('benchmark', 'time'))
+        print('-' * 42)
+        benchTime = runBench([], aditionalArguments)
+        printLine('all-plugins', runBench([], aditionalArguments))
+    printLine(value[5:], runBench(temp, aditionalArguments))
```
domanchi: Where did this `5` value come from?

scripts/bench.py
```diff
@@ -0,0 +1,55 @@
...
+aditionalArguments = getArguments()
+
+# Fill flags list with PluginOptions flag text
+for c, value in enumerate(PluginOptions.all_plugins, 0):
+    flags.append(value.disable_flag_text)
+
+# Run bechmarks for all the cases
+for c, value in enumerate(flags, 0):
+    temp = list(flags)
+    temp.pop(c)
```
domanchi: I don't think this will do what you think it will do.

```
$ cat sample.py
a = [1, 2, 3]
b = a
b.pop(0)

print(a)
print(b)
$ python sample.py
[2, 3]
[2, 3]
```

For a really good talk of why this is, check out https://www.youtube.com/watch?v=_AEJHKGk9ns

scripts/bench.py
```diff
@@ -0,0 +1,55 @@
...
+    subprocess.call('detect-secrets scan'.split() + arguments + flags, stdout=subprocess.DEVNULL)
+    return monotonic() - startTime
+
+
+def printLine(name, time):
+    print('{:<20s}{:>20s}s'.format(name, str(time)))
+
+
+flags = list()
+aditionalArguments = getArguments()
```
domanchi: Whoops. Typo :)

scripts/bench.py
```diff
@@ -0,0 +1,55 @@
...
+import argparse
+import os
+import sys
+
+from monotonic import monotonic
+
+from detect_secrets.core.usage import PluginOptions
+
+if os.name == 'posix' and sys.version_info[0] < 3:
+    import subprocess32 as subprocess
```
domanchi: I'm actually super surprised by this, because we use `subprocess` quite liberally in the codebase! ([example](https://github.com/Yelp/detect-secrets/blob/caa0e509708c5bfec157f917508d6d0abd25b792/detect_secrets/core/baseline.py#L261))

Can you give an example of when `subprocess32` succeeds when `subprocess` doesn't? Maybe we need to upgrade the other usages as well 🤔

scripts/bench.py
```diff
@@ -0,0 +1,55 @@
...
+
+flags = list()
+aditionalArguments = getArguments()
+
+# Fill flags list with PluginOptions flag text
+for c, value in enumerate(PluginOptions.all_plugins, 0):
+    flags.append(value.disable_flag_text)
+
+# Run bechmarks for all the cases
+for c, value in enumerate(flags, 0):
```
KevinHock: Is there a more descriptive name than `c`?

scripts/bench.py
dgzlopes: Fixed!

scripts/bench.py
dgzlopes: Fixed, thank you for the link! Cool info inside :)

scripts/bench.py
dgzlopes: Fixed!

scripts/bench.py
dgzlopes: It crops the original value.disable_flag_text for removing the `--no-` part. e.g. from `--no-basic-auth-scan` to `basic-auth-scan`. Maybe one cleaner way to do this is instead of slicing the value.disable_flag_text using the value.classname! What do you think?

scripts/bench.py
dgzlopes: Um. I thought that... If I want to exclude all plugins except one in each run this was the expected result:
```
➜  ~ cat something.py 
flag_list = ['1','2','3','4']
for flag_number, flag in enumerate(flag_list):
    temp = list(flag_list)
    temp.pop(flag_number)
    print(temp)
➜  ~ python something.py 
['2', '3', '4']
['1', '3', '4']
['1', '2', '4']
['1', '2', '3']
```
Maybe there are other cleaner ways to do this! What do you think? I will take a look at the talk lately today, thank you :blush:

scripts/bench.py
dgzlopes: Fixed!

scripts/bench.py
dgzlopes: Seems like in the second example of #185 `subprocess.call('detect-secrets scan'.split(), stdout=subprocess.DEVNULL)`, on Python2 breaks saying `'module' object has no attribute 'DEVNULL'`! Other GitHub issues recommended using subprocess32 and even the Python 2 wiki recommends using it [[0](https://docs.python.org/2/library/subprocess.html)] as it has better behavior in many situations... But I switched to the other option( using check_output) and now seems to work great with default subprocess. So I removed the subprocess32 dependency as it isn't strictly necessary right now for this case. 

On the other hand, as on the main project, there are many lines using subprocess... maybe further investigation on subprocess vs subprocess32 could offer some improvements to the codebase.

[[0](https://docs.python.org/2/library/subprocess.html)] https://docs.python.org/2/library/subprocess.html

scripts/bench.py
dgzlopes: Fixed to `flag_number`!

scripts/bench.py
domanchi: Gotcha. Then, a better way would be to do:

```
value[len('--no-'):]
```

scripts/bench.py
domanchi: Oh! Never mind -- you're absolutely right!

I guess the `list` was sufficient to copy the data, rather than type-casting. dgzlopes++

scripts/bench.py
dgzlopes: Fixed!

scripts/bench.py
dgzlopes: Closing this one as resolved then :)

dgzlopes:
Squashed all the commits for a cleaner git log!

PR #186, opened 2019-05-28
Benchmark script improvements

domanchi:
### Summary

Building off https://github.com/Yelp/detect-secrets/pull/181, this adds the following features:

- machine consumable output, by default. (`--pretty` preserves the human readable table)
- you can select which plugins you want to test
- timeout functionality, so scripts won't run forever
- customizing number of iterations, so you can get more accurate results

### Example Output

```
$ python scripts/benchmark.py --plugin PrivateKeyDetector --plugin Base64HighEntropyString -n 5 | jq
Running performance tests on: PrivateKeyDetector, Base64HighEntropyString
for: ['~/pg/detect-secrets']
{
    "Base64HighEntropyString": 0.8232977999992727,
    "PrivateKeyDetector": 0.6896803999989061
}
```

```
$ python scripts/benchmark.py --plugin Base64HighEntropyString --harakiri 0.5 --pretty
Running performance tests on: Base64HighEntropyString
for: ['~/pg/detect-secrets']
------------------------------------------
plugin                              time
------------------------------------------
Base64HighEntropyString   Timeout exceeded!
------------------------------------------
```


scripts/benchmark.py
```diff
@@ -0,0 +1,192 @@
...
+        default=None,       # needs to be None, otherwise append won't work as expected
+        choices=plugins,
+        action='append',
+        help=(
+            'Specifies a plugin to test. May provide multiple values. '
+            'Defaults to all.'
+        ),
+    )
+    parser.add_argument(
+        '--harakiri',
```
KevinHock: lol

scripts/benchmark.py
```diff
@@ -0,0 +1,192 @@
...
+        help=(
+            'Specifies an upper bound for the number of seconds to wait '
+            'per execution.'
+        ),
+    )
+    parser.add_argument(
+        '-n',
+        '--num-iterations',
+        default=1,
+        # TODO: assert non-negative
```
KevinHock: [This seems like the way to do that](https://stackoverflow.com/questions/14117415/in-python-using-argparse-allow-only-positive-integers)

scripts/benchmark.py
```diff
@@ -0,0 +1,192 @@
...
+            'Specifies the number of times to run the test. '
+            'Results will be averaged over this value.'
+        ),
+    )
+
+    args = parser.parse_args()
+    if not args.filenames:
+        args.filenames = [
+            os.path.realpath(
+                os.path.join(
```
KevinHock: Super nit: variable names could maybe help readability here

scripts/benchmark.py
domanchi: Whoops. Forgot to remove this lol.

Yeah, we're not breaking new ground here. We've done this before: https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/usage.py#L396

scripts/benchmark.py
domanchi: FWIW, I copied uwsgi's syntax for this: https://uwsgi-docs.readthedocs.io/en/latest/Glossary.html

Issue #180, opened 2019-05-17
Support scanning multiple specified files on the command line


MVrachev:
Hi, how can I scan multiple files?

I don't want to scan a whole folder I want to scan only a few specific files.
I want to achieve something like that: `detect-secrets scan test.js test.yaml` but right now when I run it I get: 

![image](https://user-images.githubusercontent.com/16246778/57930580-03d6b480-78bf-11e9-9110-950a9ef4fedc.png)

dgzlopes:
I think (from researching a little on the code out of curiosity) that it's only scanning the first argument (file or directory). Maybe this one is the 'offending' function: https://github.com/Yelp/detect-secrets/blob/1f3322cbe41fdeed76abb164a6c8cbb99b1e9b2c/detect_secrets/core/usage.py#L120

As it's using '?' nargs that reads only one argument if possible [[0](https://docs.python.org/3/library/argparse.html#nargs)]. I'm sure that @domanchi or @KevinHock can throw more light on this issue.

[[0](https://docs.python.org/3/library/argparse.html#nargs)] https://docs.python.org/3/library/argparse.html#nargs

domanchi:
@dgzlopes++

That's accurate. Shouldn't be too hard to modify this such that it can scan multiple files with the command line invocation you suggested: https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/baseline.py#L37

MVrachev:
I close it by mistake.

domanchi:
Closing issue, as #188 has been merged.

PR #188, opened 2019-05-29
Add multiple file support

dgzlopes:

Small fix for #180. Multiple file support might be interesting for the new benchmark script.

### DIFF
--- a/detect_secrets/core/baseline.py
+++ b/detect_secrets/core/baseline.py
@@ -5,14 +5,17 @@ import os
 import re
 import subprocess
 
+from detect_secrets.core.log import get_logger
 from detect_secrets.core.secrets_collection import SecretsCollection
+
+log = get_logger(format_string='%(message)s')
 
 
 def initialize(
+    path,
     plugins,
     exclude_files_regex=None,
     exclude_lines_regex=None,
-    path='.',
     scan_all_files=False,
 ):
     """Scans the entire codebase for secrets, and returns a
@@ -23,7 +26,7 @@ def initialize(
 
     :type exclude_files_regex: str|None
     :type exclude_lines_regex: str|None
-    :type path: str
+    :type path: list
     :type scan_all_files: bool
 
     :rtype: SecretsCollection
@@ -34,13 +37,17 @@ def initialize(
         exclude_lines=exclude_lines_regex,
     )
 
-    if os.path.isfile(path):
-        # This option allows for much easier adhoc usage.
-        files_to_scan = [path]
-    elif scan_all_files:
-        files_to_scan = _get_files_recursively(path)
-    else:
-        files_to_scan = _get_git_tracked_files(path)
+    files_to_scan = list()
+    for element in path:
+        if os.path.isdir(element):
+            if scan_all_files:
+                files_to_scan.extend(_get_files_recursively(element))
+            else:
+                files_to_scan.extend(_get_git_tracked_files(element))
+        elif os.path.isfile(element):
+            files_to_scan.append(element)
+        else:
+            log.error("detect-secrets: " + element + ": No such file or directory")
 
     if not files_to_scan:
         return output
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -120,7 +120,7 @@ class ScanOptions(object):
     def _add_initialize_baseline_argument(self):
         self.parser.add_argument(
             'path',
-            nargs='?',
+            nargs='*',
             default='.',
             help=(
                 'Scans the entire codebase and outputs a snapshot of '
--- a/tests/core/baseline_test.py
+++ b/tests/core/baseline_test.py
@@ -31,13 +31,13 @@ class TestInitializeBaseline(object):
 
     def get_results(
         self,
-        path='./test_data/files',
+        path=['./test_data/files'],
         exclude_files_regex=None,
         scan_all_files=False,
     ):
         return baseline.initialize(
+            path,
             self.plugins,
-            path=path,
             exclude_files_regex=exclude_files_regex,
             scan_all_files=scan_all_files,
         ).json()
@@ -45,10 +45,12 @@ class TestInitializeBaseline(object):
     @pytest.mark.parametrize(
         'path',
         [
-            './test_data/files',
+            [
+                './test_data/files',
 
-            # Test relative paths
-            'test_data/../test_data/files/tmp/..',
+                # Test relative paths
+                'test_data/../test_data/files/tmp/..',
+            ],
         ],
     )
     def test_basic_usage(self, path):
@@ -57,6 +59,31 @@ class TestInitializeBaseline(object):
         assert len(results.keys()) == 2
         assert len(results['test_data/files/file_with_secrets.py']) == 1
         assert len(results['test_data/files/tmp/file_with_secrets.py']) == 2
+
+    def test_with_multiple_files(self):
+        results = self.get_results(path=[
+            'test_data/files/file_with_secrets.py',
+            'test_data/files/tmp/file_with_secrets.py',
+        ])
+
+        assert len(results['test_data/files/file_with_secrets.py']) == 1
+        assert len(results['test_data/files/tmp/file_with_secrets.py']) == 2
+        assert 'test_data/files/file_with_secrets.py' in results
+        assert 'test_data/files/tmp/file_with_secrets.py' in results
+
+    def test_with_multiple_non_existent_files(self):
+        results = self.get_results(path=['non-existent-file.A', 'non-existent-file.B'])
+
+        # No expected results, because files don't exist
+        assert not results
+
+    def test_with_folders_and_files(self):
+        results = self.get_results(path=['test_data/', 'non-existent-file.B'])
+
+        assert 'test_data/files/file_with_secrets.py' in results
+        assert 'test_data/files/tmp/file_with_secrets.py' in results
+        assert 'test_data/files/file_with_no_secrets.py' not in results
+        assert 'non-existent-file.B' not in results
 
     def test_exclude_regex(self):
         results = self.get_results(exclude_files_regex='tmp*')
@@ -82,7 +109,7 @@ class TestInitializeBaseline(object):
                 ),
             ),
         ):
-            results = self.get_results(path='will_be_mocked')
+            results = self.get_results(path=['will_be_mocked'])
 
         assert not results
 
@@ -94,12 +121,12 @@ class TestInitializeBaseline(object):
             'Super hidden value "BEEF0123456789a"',
             'detect_secrets.core.secrets_collection.codecs.open',
         ):
-            results = self.get_results('will_be_mocked')
+            results = self.get_results(path=['will_be_mocked'])
 
         assert len(results['will_be_mocked']) == 1
 
     def test_scan_all_files(self):
-        results = self.get_results(path='test_data/files', scan_all_files=True)
+        results = self.get_results(path=['test_data/files'], scan_all_files=True)
         assert len(results.keys()) == 2
 
 
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -41,7 +41,7 @@ class TestMain(object):
             plugins=Any(tuple),
             exclude_files_regex=None,
             exclude_lines_regex=None,
-            path='test_data',
+            path=['test_data'],
             scan_all_files=False,
         )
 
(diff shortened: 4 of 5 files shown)


detect_secrets/core/baseline.py
```diff
@@ -12,7 +12,7 @@ def initialize(
     plugins,
     exclude_files_regex=None,
     exclude_lines_regex=None,
-    path='.',
```
domanchi: Generally speaking, it's a bad idea to have a list as a default parameter, due to the following reason:

```
def foobar(L=['a']):
    L.append('b')
    print(L)

foobar()    # ['a', 'b']
foobar()    # ['a', 'b', 'b']
```

While we may not be appending to `path` here, we're introducing a "gotcha" that may come to bite us later on. A better way of doing this is:

```
def foobar(L=None):
    if not L:
        L = []
```

And since we're doing this, we can also do:

```
def foobar(L='.'):
    if not isinstance(L, list):
        L = [L]
```

which provides a nicer API for us to use (rather than always calling the function with a list of length 1).

detect_secrets/core/baseline.py
```diff
@@ -34,13 +34,15 @@ def initialize(
         exclude_lines=exclude_lines_regex,
     )
 
-    if os.path.isfile(path):
+    if len(path) > 1 or os.path.isfile(path[0]):
         # This option allows for much easier adhoc usage.
-        files_to_scan = [path]
+        files_to_scan = list()
+        for elem in path:
+            files_to_scan.append(elem)
```
domanchi: Since `path` is already a list, isn't this just:

```
files_to_scan = path
```

?

detect_secrets/core/baseline.py
```diff
@@ -34,13 +34,15 @@ def initialize(
         exclude_lines=exclude_lines_regex,
     )
 
-    if os.path.isfile(path):
+    if len(path) > 1 or os.path.isfile(path[0]):
```
domanchi: I'm doubtful this carries over the same logic as the previous version.

After all, this should allow invocations like:

```
$ detect-secrets scan non-existent-file.A non-existent-file.B
```

right?

detect_secrets/core/baseline.py
dgzlopes: Completely missed this! Sorry, was part of some behavior testing but It's removed now.

detect_secrets/core/baseline.py
dgzlopes: Thanks for pointing this :) Tried to solve It on the last update, looks better now?

detect_secrets/core/baseline.py
dgzlopes: Took a closer look at this and to be fair I think that I misinterpreted what the previous version was doing. So updated it to:
- Running `detect-secrets scan non-existent-file.A non-existent-file.B`  skips the invocation.
- Running `detect-secrets scan` or `detect-secrets scan detect-secrets` works as expected.
- Running `detect-secrets scan non-existent-file.A existent-file.B`  skips the invocation too! But I'm not sure if this is great or not. If one argument isn't a file scanning the other ones should work?

Do you think this is the proper way to solve this?

dgzlopes:
@domanchi Added `multiple valid files` and  `multiple invalid files` tests on the last update! Should I add test for some more edge case?

And thank you for taking some time to review the PR!

tests/main_test.py
```diff
@@ -41,7 +41,7 @@ def test_scan_with_rootdir(self, mock_baseline_initialize):
             plugins=Any(tuple),
             exclude_files_regex=None,
             exclude_lines_regex=None,
-            path='test_data',
```
dgzlopes: I don't understand why tests won't work if I remove this. Running `detect-secrets scan test_data` seems to work properly.

detect_secrets/core/baseline.py
domanchi: That's a good question you raise. When it comes to command line tools usability, I defer back to tried-and-true unix tools:

```
detect-secrets $ cat asdf
cat: asdf: No such file or directory
detect-secrets $
detect-secrets $ cat asdf requirements-dev.txt
cat: asdf: No such file or directory
coverage
flake8==3.5.0
mock
monotonic
pre-commit==1.11.2
pytest
pyyaml
tox>=3.8
tox-pip-extensions
unidiff
```

Perhaps we should do something similar? You can print to stderr, or you can just use:

```
log.error(...)
```

detect_secrets/core/baseline.py
```diff
@@ -34,13 +34,19 @@ def initialize(
...
-    if os.path.isfile(path):
+    if not isinstance(path, list):
+        path = [path]
+
+    if check_all_are_files(path):
         # This option allows for much easier adhoc usage.
-        files_to_scan = [path]
-    elif scan_all_files:
-        files_to_scan = _get_files_recursively(path)
+        files_to_scan = path
```
domanchi: If we decide to pursue the option to display errors for non existing files, we should probably change this to a filter.

detect_secrets/core/baseline.py
```diff
@@ -34,13 +34,19 @@ def initialize(
         exclude_lines=exclude_lines_regex,
     )
 
-    if os.path.isfile(path):
```
domanchi: Another reason for the original logic was that it allowed us to do:

```
detect-secrets scan test_data/files/file_with_secrets.py
```

as well as

```
detect-secrets scan test_data/
```

using `_get_git_tracked_files` to get files, if it is a directory.

How does this change handle that?

dgzlopes:
Learned a lot from the review! Both about Python and form the problem. So I tried to fix the missing logic and all the small bits. Now for each element on the path, I check if It's a directory or a file: 

-If It's a directory, in case we have the `scan_all_files` flag active It runs` _get_files_recursively(element)`... If not It runs `_get_git_tracked_files(element)`. In both cases, It extends the returned list to the `files_to_scan` list.
- If it's a file It's appended to the` files_to_scan` list.
- If the element doesn't exist the error is logged!

detect_secrets/core/baseline.py
dgzlopes: Should be working now :)

detect_secrets/core/baseline.py
dgzlopes: On the new update didn't use the filter as I changed the whole logic! With the new changes, It's still needed? As I'm not checking for errors on the whole list...

detect_secrets/core/baseline.py
```diff
@@ -60,6 +66,13 @@ def initialize(
     return output
 
 
+def check_all_are_files(path):
+    for elem in path:
+        if os.path.isfile(elem) is False:
```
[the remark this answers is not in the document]
dgzlopes: Thanks for the point!

tests/core/baseline_test.py
```diff
@@ -58,6 +58,23 @@ def test_basic_usage(self, path):
...
+    def test_with_multiple_files(self):
+        results = self.get_results(path=[
+            'test_data/files/file_with_secrets.py',
+            'test_data/files/tmp/file_with_secrets.py',
+        ])
+
+        assert len(results['test_data/files/file_with_secrets.py']) == 1
+        assert len(results['test_data/files/tmp/file_with_secrets.py']) == 2
+        assert 'test_data/files/file_with_secrets.py' in results
+        assert 'test_data/files/tmp/file_with_secrets.py' in results
```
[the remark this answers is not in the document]
dgzlopes: Added test case with a mix of folders and files and some negative asserts that check if some files are or not in the results!

detect_secrets/core/baseline.py
```diff
@@ -23,7 +23,7 @@ def initialize(
 
     :type exclude_files_regex: str|None
     :type exclude_lines_regex: str|None
-    :type path: str
+    :type path: list
```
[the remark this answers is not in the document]
dgzlopes: Fixed!

detect_secrets/core/baseline.py
```diff
@@ -34,13 +34,19 @@ def initialize(
...
+    if check_all_are_files(path):
         # This option allows for much easier adhoc usage.
-        files_to_scan = [path]
-    elif scan_all_files:
-        files_to_scan = _get_files_recursively(path)
+        files_to_scan = path
     else:
-        files_to_scan = _get_git_tracked_files(path)
+        if len(path) > 1:
+            return output
```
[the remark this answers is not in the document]
dgzlopes: It should be working now!

tests/main_test.py
dgzlopes: Ok! Thanks for explaining :)

detect_secrets/core/baseline.py
dgzlopes: Added error logging!

tests/core/baseline_test.py
```diff
@@ -58,6 +58,30 @@ def test_basic_usage(self, path):
...
+    def test_with_multiple_non_existent_files(self):
+        results = self.get_results(path=['non-existent-file.A', 'non-existent-file.B'])
+
+        # No expected results, because files don't exist
+        assert not results
+
+    def test_with_folders_and_files(self):
+        results = self.get_results(path=['test_data/', 'non-existent-file.B'])
+
+        assert len(results.keys()) == 10
```
domanchi: Is there a better way you can do this?

My only concern for this magic number is:

1. Another developer creates a new test file under `test_data/`
2. They write a new test, that leverages that new file. This works.
3. For some bizarre reason, this test fails now.

The CI will catch failing tests, but it's just less convenient if there's a better option.

detect_secrets/core/usage.py
```diff
@@ -120,7 +120,7 @@ def add_arguments(self):
     def _add_initialize_baseline_argument(self):
         self.parser.add_argument(
             'path',
-            nargs='?',
+            nargs='*',
```
KevinHock: I'm not sure, but would it be possible to use `'+'`? So you wouldn't have to do
```python
:type path: str|list
```
and the `isinstance` check

MVrachev:
I believe it will be useful to add information about this feature in the README.
Maybe just a small example calling `detect-secret` on multiple files.

tests/core/baseline_test.py
dgzlopes: Added a couple of known asserts (to check that directory scan works as expected) and removed the hardcoded test!

detect_secrets/core/usage.py
dgzlopes: Well, as '+' generates an error message if there isn't at least one command-line argument present... I thought that as we are using default's, '*' would be a better fit. Running the test suite with '+' breaks with `detect-secrets scan: error: too few arguments` (Well, as intended!).

Maybe I'm missing something... But If you think it's an improvement I can take a closer look at it :)

dgzlopes:
@MVrachev I thought that multiple file support was an obvious feature for this type of tool! But I added a little Unix-like doc to the readme explaining `scan` mode inputs as this might not be that obvious :)

detect_secrets/core/usage.py
KevinHock: You’re completely right :) 

From reading https://stackoverflow.com/questions/23490152/list-of-arguments-with-argparse#23490179 it seems it should be always be a list with *, so maybe that’s true?

dgzlopes: Yes! Looks like all is passed as a list, even the default argument ` '.' `

So the isInstance() might be redundant as when we pass (on the initialize()) `path='.' `  it's never used... Running scan without path(s) will trigger argparse's default. 

This was closely related to https://github.com/Yelp/detect-secrets/pull/188#discussion_r288614772 (@domanchi) so maybe using the list as a default isn't that great... But I think '+' passes all as a list too! What's the best way to handle this?

KevinHock: If I understand, the 

```python
    if not isinstance(path, list):
        path = [path]
```
is always dead-code, so we can remove that. 

From reading the code, it seems a path is always passed in to `initialize`, so there doesn't seem to be a need for a default argument I think, right?

If so we can change
```python
def initialize(
    plugins,
    exclude_files_regex=None,
    exclude_lines_regex=None,
    path='.',
    scan_all_files=False,
):
```
to
```
def initialize(
    plugins,
    path,
    exclude_files_regex=None,
    exclude_lines_regex=None,
    scan_all_files=False,
):
```

dgzlopes: As I thought :) Thanks for the writeup. Updated the PR and fixed the failing tests.

KevinHock: Looks awesome 👏

PR #194, opened 2019-06-15
Verifiable secrets scaffolding

domanchi:
### Summary

After talking with @killuazhu (and referencing the whitepaper from https://github.com/Yelp/detect-secrets/issues/159), this seems like a completely feasible option.

This is my first take of supporting this.

I've written a Slack verifier (as an example usage for bearer secrets), and an AWS verifier (as an example usage for multi-factor secrets). Test cases were run with no network connectivity, just to make sure that test cases don't rely on network calls.

The `verify` functions are manually tested, because (obviously) we don't want to commit real secrets to have automated tests for it.

### Example Usage

```
$ cat slack_token.txt | detect-secrets scan --string
```

or

```
$ detect-secrets scan test_data/each_secret.py --no-verify
```

### DIFF
--- a/detect_secrets/plugins/aws.py
+++ b/detect_secrets/plugins/aws.py
@@ -3,9 +3,17 @@ This plugin searches for AWS key IDs
 """
 from __future__ import absolute_import
 
+import hashlib
+import hmac
 import re
+import string
+import textwrap
+from datetime import datetime
+
+import requests
 
 from .base import RegexBasedDetector
+from detect_secrets.core.constants import VerifiedResult
 
 
 class AWSKeyDetector(RegexBasedDetector):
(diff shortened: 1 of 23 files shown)


detect_secrets/plugins/aws.py
```diff
@@ -15,3 +23,162 @@ class AWSKeyDetector(RegexBasedDetector):
     denylist = (
         re.compile(r'AKIA[0-9A-Z]{16}'),
```
KevinHock: I wonder what the verify code would look like if we supported all the key types in https://github.com/dxa4481/truffleHogRegexes/pull/13/files

domanchi: We probably can cross that bridge, when we support them.

OR, if you have another valid prefixed key to test, that's easily testable too.

KevinHock: ++ for that 🤔 , I'll make an issue later for the other types 👍

PR #195, opened 2019-06-17
New Artifactory / Slack Patterns

adrianbn:
The modifications in this PR are twofold:

1. Add ability to detect Slack Webhooks
2. Improved the Artifactory password regex to catch passwords of different lengths and rotated passwords (Third char increments after user rotates password).

### DIFF
--- a/detect_secrets/plugins/artifactory.py
+++ b/detect_secrets/plugins/artifactory.py
@@ -11,7 +11,7 @@ class ArtifactoryDetector(RegexBasedDetector):
 
     denylist = [
         # artifactory tokens begin with AKC
-        re.compile(r'(?:\s|=|:|"|^)AKC\w{10,}'),    # api token
-        # artifactory encrypted passwords begin with AP6
-        re.compile(r'(?:\s|=|:|"|^)AP6\w{10,}'),    # password
+        re.compile(r'(?:\s|=|:|"|^)AKC[a-zA-Z0-9]{10,}'),    # api token
+        # artifactory encrypted passwords begin with AP[A-Z]
+        re.compile(r'(?:\s|=|:|"|^)AP[\dABCDEF][a-zA-Z0-9]{8,}'),    # password
     ]
--- a/detect_secrets/plugins/slack.py
+++ b/detect_secrets/plugins/slack.py
@@ -13,5 +13,13 @@ class SlackDetector(RegexBasedDetector):
     secret_type = 'Slack Token'
 
     denylist = (
+        # Slack Token
         re.compile(r'xox(?:a|b|p|o|s|r)-(?:\d+-)+[a-z0-9]+', flags=re.IGNORECASE),
+        # Slack Webhooks
+        re.compile(
+            r"""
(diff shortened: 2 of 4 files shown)


detect_secrets/plugins/slack.py
```diff
@@ -10,8 +10,16 @@
 
 class SlackDetector(RegexBasedDetector):
 
-    secret_type = 'Slack Token'
+    secret_type = 'Slack Token / Webhook'
```
domanchi: This is going to be a [non-backwards compatible change](https://github.com/Yelp/detect-secrets/blob/master/CHANGELOG.md#boom-breaking-changes-3).

Are you sure you want to do this?

KevinHock: (Backwards compatibility aside)

I think we may want to make a different plugin, I say this because we should only have 1 `verify` method once #194 is merged, and verifying token's vs. webhooks is different.

detect_secrets/plugins/artifactory.py
```diff
@@ -11,7 +11,7 @@ class ArtifactoryDetector(RegexBasedDetector):
 
     denylist = [
         # artifactory tokens begin with AKC
-        re.compile(r'(\s|=|:|"|^)AKC\w{10,}'),    # api token
-        # artifactory encrypted passwords begin with AP6
-        re.compile(r'(\s|=|:|"|^)AP6\w{10,}'),    # password
+        re.compile(r'(\s|=|:|"|^)AKC[a-zA-Z0-9]{10,}'),    # api token
+        # artifactory encrypted passwords begin with AP[A-Z]
+        re.compile(r'(\s|=|:|"|^)AP[\dABCDEF][a-zA-Z0-9]{8,}'),    # password
```
KevinHock: cc @justineyster just as a heads up

detect_secrets/plugins/slack.py
domanchi: I'm fine with having the same plugin.

In the `verify` method, we can do:

```
def verify(self, token):
    if token.startswith('https://hooks.slack.com'):
        # do logic
```

KevinHock: As far as backwards compatibility, I don't mind but there may be a user that does. Slack isn't one of the noisier plugins so I'd say this is okay :) What do you think?

(To explain better, the reason this breaks backwards compatibility is because all of a user's old acknowledged Slack tokens will be re-alerted on, because the "[hashed_secret](https://github.com/Yelp/detect-secrets/blob/master/CONTRIBUTING.md#technical-details)" value will change.)

domanchi: I'm in favor of the slightly less descriptive, but more compatible decision to keep `secret_type`. IMO, adding the clarity that it also catches webhooks is less important.

killuazhu: I would agree with @domanchi on more compatible approach.

adrianbn: No objection here. I'll leave the type as it was.

adrianbn:
I've returned the secret type and rebased master. Let me know if there are further changes you want :)

Issue #191, opened 2019-06-10
Add an audit --display option


KevinHock:
This will be extraordinarily helpful when developing new plugins, so I think it might be great to incorporate into detect-secrets.

`--display` would display, per plugin, the 

```
Plugin A:
    Positives:
         - some_real_key
    Negatives:
         - "foo"
    Unknowns:
        - "bar"
...
```

This would take some of the manual effort involved with adding a new plugin and adjusting the sensitivities and regexes, e.g. right now it is roughly: run scan, `cat` the baseline and see what was found by what plugins by comparing line numbers w/ the file you scanned, and then `audit`ing to see which plaintext group was captured as a secret.

We can probably use `textwrap` more or something, to make the output look pretty.

domanchi:
To be clear, are you suggesting we determine the `positives` and `negatives` by the `is_secret` attribute?

If this is so, might I suggest making this machine readable instead? And maybe the flag to be `--results`?

```
{
    "Base64HighEntropyString": {
        "results": {
            "positive": [
                "some_real_key",
            ],
            "negative": [
                 "foo",
            ],
            "unknown": [
                 "bar",
            ],
        },
        "config": {
            "base64_limit": 4.5,
        }
    }
}
```

KevinHock:
re: `is_secret` attribute: Yup

++, that sounds great.

KevinHock:
Maybe `--display-results` since results is sort of the baseline itself.

Just noting for later: we could also have this display the entropy count of each secret found. That'd be nice.

OiCMudkips:
What about this?

New option `--output-format` with three options: `baseline-view`, `file-view` and `plugin-view`:

`baseline-view` and `file-view` are the same. It is the *default* option. It makes `detect-secrets` output results in the existing format. `baseline-view` because this is what `pre-commit` consumes as a baseline, and `file-view` because this is literally what the hierarchy has at the upper-most level under the `results` key:
```
{
    ...existing metadata...,
    "results":{
        {filename1}: [SECRETS],
        {filename2}: [SECRETS]
    }
}
```

`plugin-view` would have a plugin at the upper-most level under the `results` key:
```
{
    ...existing metadata...,
    "results":{
        {plugin_name1}: {
            "config": {bla},
            "plugin_results": {
                "positive": [SECRETS],
                "negative": [SECRETS],
                "unknown": [SECRETS]
            }
        },
        {plugin_name2}: {
            "config": {bla},
            "plugin_results": {
[...]

PR #205, opened 2019-07-02
Add audit result view

OiCMudkips:
Implements #191

As of writing this only shows hashed secrets. I'll update this once I get plaintext secrets and git repo info shown in the results.

### DIFF
--- a/detect_secrets/core/audit.py
+++ b/detect_secrets/core/audit.py
@@ -225,7 +225,7 @@ def determine_audit_results(baseline, baseline_path):
         'results': defaultdict(lambda: deepcopy(EMPTY_PLUGIN_AUDIT_RESULT)),
     }
 
-    secret_type_mapping = get_mapping_from_secret_type_to_class_name()
+    secret_type_to_plugin_name = get_mapping_from_secret_type_to_class_name()
 
     for filename, secret in all_secrets:
         plaintext_line = _get_file_line(filename, secret['line_number'])
@@ -239,7 +239,7 @@ def determine_audit_results(baseline, baseline_path):
         except SecretNotFoundOnSpecifiedLineError:
             secret_plaintext = plaintext_line
 
-        plugin_name = secret_type_mapping[secret['type']]
+        plugin_name = secret_type_to_plugin_name[secret['type']]
         audit_result = AUDIT_RESULT_TO_STRING[secret.get('is_secret')]
         audit_results['results'][plugin_name]['results'][audit_result].append(secret_plaintext)
 
--- a/detect_secrets/core/audit.py
+++ b/detect_secrets/core/audit.py
@@ -14,6 +14,8 @@ from ..plugins.common import initialize
 from ..plugins.common.filetype import determine_file_type
 from ..plugins.common.util import get_mapping_from_secret_type_to_class_name
 from ..plugins.high_entropy_strings import HighEntropyStringsPlugin
+from ..util import get_git_remotes
+from ..util import get_git_sha
 from .baseline import merge_results
 from .bidirectional_iterator import BidirectionalIterator
 from .code_snippet import CodeSnippetHighlighter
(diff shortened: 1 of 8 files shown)


OiCMudkips:
I ran this branch against the `detect-secrets` repo and got these results: https://i.fluffy.cc/pWlPvpGl4Hz0jl4PGkcxrV3GMcB64CTv.html

If you look at `baseline.txt` you'll notice 2 audited secrets (i.e. have a `is_secret` attribute) and then in `display-results output.txt` the 2 secrets are marked as `positive`/`negative` and the rest are unknown.

`help_text.txt` is the help text of `audit` with the new option.

detect_secrets/plugins/common/initialize.py
```diff
@@ -16,6 +11,7 @@
 from ..stripe import StripeDetector                         # noqa: F401
 from detect_secrets.core.log import log
 from detect_secrets.core.usage import PluginOptions
+from detect_secrets.plugins.common.util import get_mapping_from_secret_type_to_class_name
```
KevinHock: Nit: Since we use relative imports above, for within the plugins folder, so I think we can do `from .util` in this case

OiCMudkips:
[baseline.txt](https://github.com/Yelp/detect-secrets/files/3352775/baseline.txt)
[display-results output.txt](https://github.com/Yelp/detect-secrets/files/3352776/display-results.output.txt)
[help_text.txt](https://github.com/Yelp/detect-secrets/files/3352777/help_text.txt)

detect_secrets/plugins/common/initialize.py
OiCMudkips: Changed as suggested

OiCMudkips:
Updated baseline (well this should be the same as before) and output:

[display-results.txt](https://github.com/Yelp/detect-secrets/files/3357034/display-results.txt)
[baseline.txt](https://github.com/Yelp/detect-secrets/files/3357035/baseline.txt)

OiCMudkips:
New output: 
[display-results.txt](https://github.com/Yelp/detect-secrets/files/3364088/display-results.txt)

Note it has the SHA of the last commit of this PR!

tests/core/audit_test.py
```diff
@@ -475,6 +475,228 @@ def new_baseline(self):
...
+    @pytest.fixture
+    def mock_get_git_remotes(self):
+        with mock.patch(
+            'detect_secrets.core.audit.get_git_remotes',
+            return_value=None,
+            autospec=True,
+        ) as _mock:
+            yield _mock
+
+    def get_audited_baseline(self, plugin_config={}, is_secret=None):
```
KevinHock: Since `plugin_config` is always passed, afaict, I don't think you need the default value. (There's also the python gotcha w/ dict's as a default arg.)

Also you can do e.g. `get_audited_baseline(plugin_config=foo, is_secret=True)` rather than `get_audited_baseline(foo,True)` as it's a little more readable what the args are from seeing the call.

tests/core/audit_test.py
```diff
@@ -475,6 +475,228 @@ def new_baseline(self):
...
+        plugin_config,
+    ):
+        plaintext_secret = 'some_plaintext_secret'
+        mock_get_raw_secret_value.return_value = plaintext_secret
+        baseline = self.get_audited_baseline(plugin_config, None)
+
+        results = audit.determine_audit_results(baseline, '.secrets.baseline')
+
+        if plugin_config:
+            assert results['results']['HexHighEntropyString']['config'].items() \
```
KevinHock: Not sure if I follow, what does this comparison do?

detect_secrets/core/audit.py
```diff
@@ -176,6 +197,81 @@ def compare_baselines(old_baseline_filename, new_baseline_filename):
...
+        "repo_info": {
+            "remote": "remote url",
+            "sha": "sha of repo checkout"
+        },
+    }
+    """
+    all_secrets = _secret_generator(baseline)
+
+    audit_results = {
+        'results': defaultdict(lambda: deepcopy(EMPTY_PLUGIN_AUDIT_RESULT)),
```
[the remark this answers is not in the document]
KevinHock: Great use of `lambda` btw 🐑

detect_secrets/core/audit.py
```diff
@@ -424,6 +520,17 @@ def _handle_user_decision(decision, secret):
         del secret['is_secret']
 
 
+def _get_file_line(filename, line_number):
+    """
+    Attempts to read a given line from the input file.
+    """
+    try:
+        with codecs.open(filename, encoding='utf-8') as file:
```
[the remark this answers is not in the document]
OiCMudkips: Good point, fixed

detect_secrets/core/audit.py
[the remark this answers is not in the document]
OiCMudkips: I'm OK with renaming it to `plugins`, but I think there needs to be a key. If we can get git config info we dump it into a key called `repo_info` on the top level. If the plugins were also at the top-level then it would be really annoying for clients to have to filter out the plugin results in iteration (assuming the client is a machine or `jq`).

detect_secrets/core/audit.py
```diff
@@ -176,6 +197,81 @@ def compare_baselines(old_baseline_filename, new_baseline_filename):
...
+        },
+    }
+    """
+    all_secrets = _secret_generator(baseline)
+
+    audit_results = {
+        'results': defaultdict(lambda: deepcopy(EMPTY_PLUGIN_AUDIT_RESULT)),
+    }
+
+    secret_type_mapping = get_mapping_from_secret_type_to_class_name()
```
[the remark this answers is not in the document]
OiCMudkips: Sounds good, done.

tests/core/audit_test.py
OiCMudkips: Makes sense, fixed

tests/core/audit_test.py
OiCMudkips: This checks that the config in the test baseline is actually returned from `determine_audit_results`.

Also, I noticed the `if plugin_config` check doesn't really do anything useful, so I took it out.

detect_secrets/core/audit.py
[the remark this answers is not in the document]
KevinHock: ++

PR #207, opened 2019-07-08
Improve sequential string detection

### LINKED ISSUE #200: Add base 64  charset to filters.py
Right now we alert off of 
`_chars: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/",`

We can add something to https://github.com/Yelp/detect-secrets/blob/13575ede7f9d6f7cb17ac317a19ec5060c97170b/detect_secrets/plugins/common/filters.py#L19-L38, to prevent this.

OiCMudkips:
* Ensure that we catch all base64 strings correctly
* Add an explicit check for alphanumeric sequences
* Add some more tests for sequential string detection

Fixes #200

### DIFF
--- a/tests/plugins/common/filters_test.py
+++ b/tests/plugins/common/filters_test.py
@@ -11,11 +11,25 @@ class TestIsSequentialString:
     @pytest.mark.parametrize(
         'secret',
         (
+            # ascii sequence
             'ABCDEF',
+            'ABCDEFGHIJKLMNOPQRSTUVWXYZ',
 
             # Number sequences
             '0123456789',
             '1234567890',
+
+            # Alphanumeric sequences
+            'abcdefghijklmnopqrstuvwxyz0123456789',
+            '0123456789abcdefghijklmnopqrstuvwxyz',
+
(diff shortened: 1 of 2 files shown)


tests/plugins/common/filters_test.py
```diff
@@ -11,11 +11,25 @@ class TestIsSequentialString:
...
         (
+            # ascii sequence
             'ABCDEF',
+            'ABCDEFGHIJKLMNOPQRSTUVWXYZ',
 
             # Number sequences
             '0123456789',
             '1234567890',
+
+            # alphanumeric sequences
```
KevinHock: Super nit lol: Most comments start capitalized, e.g. `Number sequences` above

OiCMudkips:
Right, just for posterity, I tested this manually with a YAML file with the config `config: ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/` and the resulting baseline had no secrets, which is expected with this change.

Issue #92, opened 2018-11-05
Add options for baseline diff minimizing


KevinHock:
`--no-line-numbers` (in baseline)
`--no-generated-at` (in baseline)
and "Make pre-commit hook only look at the git diff" options.

Are all possible ideas.

stephenpaulger:
In an active project these fields can change quite often and make merging and code review slightly more difficult. I would definitely make use of these options if they were available.

domanchi:
Removing `good first issue` label.

The super tricky thing about this is that there's a reason for each of the fields. For example:

- The `line_number` field powers the `audit` functionality
- The `version` field indicates which version the baseline was generated at, to assist with backwards compatibility, and debugging issues when something breaks due to a version bump.
- The `generated_at` field allows security teams to manually verify post-hoc whether this tool would have caught a secret found, if clients had used this pre-commit tool. This really helps with evangelization and promotion of best practices, seeing that pre-commit by its nature, is mostly an opt-in tool.

It's currently unclear what we would do for a "minimized baseline" to address the above points.

KevinHock:
>The generated_at field allows security teams to manually verify post-hoc whether this tool would have caught a secret found, if clients had used this pre-commit tool. This really helps with evangelization and promotion of best practices, seeing that pre-commit by its nature, is mostly an opt-in tool.

What are the benefits of this over `git log`?

> The version field indicates which version the baseline was generated at, to assist with backwards compatibility, and debugging issues when something breaks due to a version bump.

This doesn't change with many multiple users running their `pre-commit` hooks, causing merge conflicts etc. So I don't think this needs to be trimmed in the hypothetical future `--minimize-baseline-diff` option.

>The line_number field powers the audit functionality.

This is tricky, we'd need to change the audit functionality to look only for a secret matching the [`hased_secret`](https://github.com/Yelp/detect-secrets/issues/175) before doing this.

I don't know if the e.g. `--no-generated-at (in baseline)` or `"Make pre-commit hook only look at the git diff"` options are complicated though.

PR #208, opened 2019-07-15
Use plugin `analyze` function in audit functionality

### LINKED ISSUE #189: Audit fails to find non-quoted base64 strings
### Reproduction Steps

```
$ cat test.file
SECRET_ACCESS_KEY=dGhpcyBuZWVkcyB0byBiZSBsb25nIGVub3VnaCB0byBjcmVhdGUgZW50cm9weQ
$ detect-secrets scan test.file > baseline.json
$ detect-secrets audit baseline.json
Secret:      1 of 1
Filename:    test.file
Secret Type: Base64 High Entropy String
----------
ERROR: Secret not found on line 1!
Try recreating your baseline to fix this issue.
----------
What would you like to do? (s)kip, (q)uit:
```

But it works if we do:

```
$ cat test.file
SECRET_ACCESS_KEY="dGhpcyBuZWVkcyB0byBiZSBsb25nIGVub3VnaCB0byBjcmVhdGUgZW50cm9weQ"
$ detect-secrets scan test.file > baseline.json
$ detect-secrets audit baseline.json
Secret:      1 of 1
Filename:    test.file
Secret Type: Base64 High Entropy String
----------
1:SECRET_ACCESS_KEY="dGhpcyBuZWVkcyB0byBiZSBsb25nIGVub3VnaCB0byBjcmVhdGUgZW50cm9weQ"
----------
Is this a valid secret? i.e. not a false-positive (y)es, (n)o, (s)kip, (q)uit:
```

OiCMudkips:
## Old algo

The audit functionality used to use the following algorithm to get the plaintext of a secret:
(1) Take the secret type and get the right plugin `PluginA`.
(2) Take the secret file `FileA` and line number `x`.
(3) Get a list of secrets in `FileA` on line `x` using `PluginA.secret_generator()`.
(4) Return the plaintext which matches the input secret's hash, or raise if the secret isn't found.

## New algo

This PR changes the algorithm. Specifically, it replaces (3) with:

(a) Scan the whole file `FileA` for secrets using `PluginA.analyze()`.

## Benefits

The immediate motivation for this PR was #189.

This adjustment makes the `audit` functionality less flakey. Plugins can implement more specific behaviour regarding splitting lines in its `analyze` function, and this PR makes it so that we can use that information in `audit` as well, making it more likely we'll be able to rediscover the secret.

As @KevinHock also points out, this will allow `audit` to be more flexible, in that the secret can change lines, but `audit` will still find the secret.

## Why not just run `scan` over the whole codebase again?

It would be a waste of time to scan literally every file with every plugin when we have a list of `(file, plugin)` pairs which actually yielded secrets in the baseline.

This approach was also easier to code IMO.

## Why are there also random new tests?

Replacing the old algorithm resulted in less LOC in `detect_secrets/`, which reduced overall code coverage, which `make test` complained about :)

## Notes for reviewing

`get_raw_secret_value` was completely rewritten so the `Split` code-review view is a lot more useful.

Auto close tags: Fixes #189.

### DIFF
--- a/detect_secrets/core/audit.py
+++ b/detect_secrets/core/audit.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import codecs
+import io
 import json
 import os
 import subprocess
@@ -16,9 +17,7 @@ except ImportError:  # pragma: no cover
     from functools32 import lru_cache
 
 from ..plugins.common import initialize
-from ..plugins.common.filetype import determine_file_type
 from ..plugins.common.util import get_mapping_from_secret_type_to_class_name
-from ..plugins.high_entropy_strings import HighEntropyStringsPlugin
 from ..util import get_git_remotes
 from ..util import get_git_sha
 from .baseline import merge_results
@@ -27,7 +26,6 @@ from .code_snippet import CodeSnippetHighlighter
 from .color import AnsiColor
 from .color import colorize
 from .common import write_baseline_to_file
-from .potential_secret import PotentialSecret
 
 
 class SecretNotFoundOnSpecifiedLineError(Exception):
@@ -233,16 +231,17 @@ def determine_audit_results(baseline, baseline_path):
     secret_type_to_plugin_name = get_mapping_from_secret_type_to_class_name()
 
     for filename, secret in all_secrets:
-        plaintext_line = _get_file_line(filename, secret['line_number'])
+        file_contents = _open_file_with_cache(filename)
+
         try:
             secret_plaintext = get_raw_secret_value(
-                secret_line=plaintext_line,
                 secret=secret,
                 plugin_settings=baseline['plugins_used'],
+                file_handle=io.StringIO(file_contents),
                 filename=filename,
             )
         except SecretNotFoundOnSpecifiedLineError:
-            secret_plaintext = plaintext_line
+            secret_plaintext = _get_file_line(filename, secret['line_number'])
 
         plugin_name = secret_type_to_plugin_name[secret['type']]
         audit_result = AUDIT_RESULT_TO_STRING[secret.get('is_secret')]
@@ -604,9 +603,9 @@ def _get_secret_with_context(
         )
 
         raw_secret_value = get_raw_secret_value(
-            snippet.target_line,
             secret,
             plugin_settings,
+            io.StringIO(file_content),
             filename,
         )
 
@@ -624,20 +623,20 @@ def _get_secret_with_context(
 
 
 def get_raw_secret_value(
-    secret_line,
     secret,
     plugin_settings,
+    file_handle,
     filename,
 ):
     """
-    :type secret_line: str
-    :param secret_line: the line on which the secret is found
-
     :type secret: dict
     :param secret: see caller's docstring
 
     :type plugin_settings: list
     :param plugin_settings: see caller's docstring
+
+    :type file_handle: file object
+    :param file_handle: Open handle to file where the secret is
 
     :type filename: str
     :param filename: this is needed, because PotentialSecret uses this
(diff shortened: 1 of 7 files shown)


detect_secrets/core/audit.py
```diff
@@ -612,36 +602,16 @@ def get_raw_secret_value(
...
-        secret_line,
-        filetype=determine_file_type(filename),
-    ):
-        secret_obj = PotentialSecret(
-            plugin.secret_type,
-            filename,
-            secret=raw_secret,
-        )
+    with codecs.open(filename, encoding='utf-8') as f:
+        plugin_secrets = plugin.analyze(f, filename)
```
domanchi: This looks like we're re-scanning the file **per secret audited**. How does this perform with large files?

domanchi:
FWIW, the audit functionality used line numbers as an optimization: there was no point re-scanning every line in the file, if you knew exactly which line to go to. Furthermore, since the baseline was ordered by increasing line number, the file would only have to be read once.

After all, the audit tool is an automated tool for the manual effort of opening each file, go to the line, find the secret yourself. These changes seem to cause a performance hit, and the benefit is not super clear.

OiCMudkips:
As an aside, I don't know if the file is only read once as the code is written now?

In audit `audit_baseline`, we `_print_context` for every secret. `_print_context` calls `_get_secret_with_context` calls `CodeSnippetHighlighter. get_code_snippet` calls `CodeSnippetHighlighter. _get_lines_in_file` which [open the whole file](https://github.com/Yelp/detect-secrets/blob/a58a07d795046c6eb0357e7d42d1b8587b4f50ab/detect_secrets/core/code_snippet.py). I didn't see any caching in this chain.

Doing some Googling, it seems like there's a [standard module for opening a single line](https://docs.python.org/3/library/linecache.html#module-linecache) but it doesn't seem to be in use here. I don't know that we actually want to include this module considering it seems to be couple with the `traceback` module and that it's only `>= py35`.

---

To clarify, the benefit here is to fix #189. Specifically, the high-entropy plugin has filetype-specific logic that was lost in the audit mode, since `audit` calls a different function from `scan`.

I agree that this change is opening and scanning the whole file, and this is a performance hit versus the status quo of opening the whole file and scanning 1 line. Let me think of a way to avoid this.

OiCMudkips:
I think @KevinHock also has some reasons for why we would potentially want this in the comments of #189.

KevinHock:
While I agree we should check the performance of this, I think the benefits of this make it worth doing:





Firstly, the statements
>there was no point re-scanning every line in the file, if you knew exactly which line to go to.

and

>[The line number does not play a part in the identification of a potential secret because code is expected to move around through continuous iteration.](https://github.com/Yelp/detect-secrets/blame/master/CONTRIBUTING.md#L198)

Kind of contradict each other. Even if we explicitly call it out in our docs, it is beneficial to some degree to not rely on line numbers for auditing, and to be able to not have to tell the user about something somewhat unintuitive.



Secondly, it is blocking the [only complicated part](https://github.com/Yelp/detect-secrets/issues/92#issuecomment-505558043) of baseline diff minimization (#92), which is a good option to have for an internal reason, and we've had at least one external person ask us for it.



Lastly and most importantly, it will improve maintainability due to preventing any bugs like #189 in the future.


As an aside, people may very well scan thousands of repos in a cron like fashion with detect-secrets, however nobody audit's thousands of repos in a cron like fashion, so perf when scanning is significantly more important.

detect_secrets/core/audit.py
KevinHock: It seems like we can avoid this by either changing or not using `_secret_generator` before calling this method

https://github.com/Yelp/detect-secrets/blob/2b40c99e4d5282102ec34d45dac9bd7884ebcd64/detect_secrets/core/audit.py#L288-L292

Right now we call this method on every secret, when we only need to call it for each key in `baseline['results']`.

tests/core/audit_test.py
```diff
@@ -528,32 +534,39 @@ def get_audited_baseline(self, plugin_config, is_secret):
...
         return baseline_fixture
 
     @pytest.mark.parametrize(
-        'plugin_config', [{}, {'hex_limit': 2}],
+        'plugins_used',
+        [
+            # NOTE: The first config here needs to be
+            # the HexHighEntropyString config for this test to work.
+            [{'name': 'HexHighEntropyString'}],  # plugin w/o config
+            [{'name': 'HexHighEntropyString', 'hex_limit': 2}],  # plguin w/config
```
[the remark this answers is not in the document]
OiCMudkips: Whoops, fixed

OiCMudkips:
This branch is now based off #213 and #213 should be reviewed before this.

OiCMudkips:
With this in mind, you actually want to review 0d499fd48dd6daa3ad80742b94694f7d511624e3 (Make a line of code more pythonic) and later commits.

OiCMudkips:
In c8b60c7 I added a new test for the force-show-secret feature. This was because the previous commit was lacking coverage in Python 2 and failing the coverage test.

Why wasn't Python 3 failing coverage tests too? This was [biting us](https://stackoverflow.com/questions/9163940/property-getter-setter-have-no-effect-in-python-2) but it only applies for Python 2.

PR #223, opened 2019-08-14
Handle binary values in YAML files

### LINKED ISSUE #202: Improve handling of !!binary in yaml files
There are various explicit tags in yaml https://yaml.org/spec/1.2/spec.html, after some testing I don't think we handle `!!binary` that well, for instance. Resulting in false-negatives. I will investigate more thoroughly and update this issue later.

OiCMudkips:
(WIP as of writing)

This is step (1) in supporting binary in both YAML and non-YAML files.

This makes it so that instead of immediately converting the base64-encoded binary into a binary value in python, we just interpret the binary as a normal string, but annotate it as such with the `is_binary` flag.

This is needed so that plugins can scan a different value from the value hashed into baselines.

**Alternatives**:
* Convert the binary to the desired form discussed [here](https://github.com/Yelp/detect-secrets/issues/202#issuecomment-521033056) in `YamlFileParser` and put the desired form into `__value__`
    * Why not?: We need the original base64-format in the plugin to hash into baselines. It's possible to retrieve it, but why do extra work?
* Instead of `__value__` and `__is_binary__`, have fields `__scan_value__` and `__plaintext_value__`
    * Why not?: I think we're going to end up with a new method in the high-entropy plugin `analyze_binary_content` (in addition to [`analyze_string_content`](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/high_entropy_strings.py#L89)) which knows how to do the conversion to the "desired format", but that depends on us knowing that the string is a binary as opposed to some other special sauce, so I think the `is_binary` information is important.
    * Also my gut feeling says what I've written will be fine; I don't think there's any other data types we'll need to consider.

Closes #202

------

The next steps would be to make the high-entropy plugin actually use the `is_binary` information to do the conversion and output results correctly.

### DIFF
--- a/detect_secrets/plugins/common/yaml_file_parser.py
+++ b/detect_secrets/plugins/common/yaml_file_parser.py
@@ -16,7 +16,7 @@ class YamlFileParser(object):
     Therefore, we take a different approach: intercept the parsing of the yaml
     file to identify string values. This assumes:
 
-        1. Secrets are strings
+        1. Secrets are strings or binaries
         2. Secrets are not keys
 
     Then, we calculate the entropy of those string values.
@@ -75,7 +75,10 @@ class YamlFileParser(object):
         """
         new_values = []
         for key, value in map_node.value:
-            if not value.tag.endswith(':str'):
+            if not (
+                value.tag.endswith(':str') or
+                value.tag.endswith(':binary')
+            ):
                 new_values.append((key, value))
                 continue
 
@@ -83,14 +86,19 @@ class YamlFileParser(object):
                 tag=map_node.tag,
                 value=[
                     self._create_key_value_pair_for_mapping_node_value(
-                        '__value__',
-                        value.value,
-                        'tag:yaml.org,2002:str',
+                        key='__value__',
+                        value=value.value,
+                        tag='tag:yaml.org,2002:str',
                     ),
                     self._create_key_value_pair_for_mapping_node_value(
-                        '__line__',
-                        str(value.__line__),
-                        'tag:yaml.org,2002:int',
+                        key='__line__',
+                        value=str(value.__line__),
+                        tag='tag:yaml.org,2002:int',
+                    ),
+                    self._create_key_value_pair_for_mapping_node_value(
+                        key='__is_binary__',
+                        value=str(value.tag.endswith(':binary')),
+                        tag='tag:yaml.org,2002:bool',
                     ),
                 ],
             )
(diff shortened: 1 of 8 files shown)


OiCMudkips:
The next steps would be to make the high-entropy plugin actually use the `is_binary` information to do the conversion and output results correctly.

detect_secrets/plugins/common/yaml_file_parser.py
```diff
@@ -75,7 +75,7 @@ def _tag_dict_values(self, map_node):
         """
         new_values = []
         for key, value in map_node.value:
-            if not value.tag.endswith(':str'):
+            if not value.tag.endswith(':str') and not value.tag.endswith(':binary'):
```
KevinHock: super nit: feel free to ignore
i feel like
```python
if not (
    value.tag.endswith(':str')
    or
    value.tag.endswith(':binary')
):
```
is a little prettier ⚡️ 💟

detect_secrets/plugins/common/yaml_file_parser.py
```diff
@@ -92,6 +92,11 @@ def _tag_dict_values(self, map_node):
                         str(value.__line__),
                         'tag:yaml.org,2002:int',
                     ),
+                    self._create_key_value_pair_for_mapping_node_value(
+                        '__is_binary__',
```
KevinHock: I realize the other calls of this don't do this, but maybe we can use keyword args to make things clearer, it helps with the other function calls in this file.

OiCMudkips:
This is done. The general strategy is:
1) Parse YAML binaries, so that they eventually become a bytes object `b'something'`. In Python 2, this will be the same as `'something'` though, so we have an `__is_binary__` field to tell us whether the value is actually a binary in the plugin.
2) For any value with the `is_binary` field set to `True`, `decode_binary` it to turn it into a `str` and run it through the plugin's `analyze_string` function.
3) If the value is a secret, create a `PotentialSecret` with the `str`. This `str` isn't what's actually in the file, so apply the reverse of `decode_binary` (which is `encode_to_binary`), `yaml.dump` it, then strip the `!!binary`. This will leave us with the secret encoded in the binary yaml format. Set this as the secret in the `PotentialSecret` object.

The complicated-ness of (3) is needed so that we can actually find the secret in the file afterwards using the existing `audit` functionality.

KevinHock:
re: ‘HighEntropyStringsPlugin is now an abstract class’ in commit message, for the explanation of the new class in the test, I’m on my phone but I don’t see that it’s abstract.

KevinHock:
Oh wait, since it has abstract methods it’s automatically an abstract class? That makes sense, maybe we can make that more explicit like with, I believe Base plugin might be a good example.

OiCMudkips:
khock++ I think the test was able to instantiate an abstract class because we didn’t actually mark any methods as abstract, but once I added some abstract methods, the test started failing so I had to fix it.

PR #229, opened 2019-08-23
[Keyword Plugin] Various accuracy improvements 

KevinHock:
🔭 ➕ ➕


detect_secrets/plugins/common/filetype.py
```diff
@@ -18,22 +41,8 @@ def determine_file_type(filename):
...
-    elif filename.endswith('.py'):
-        return FileType.PYTHON
-    elif (
-        filename.endswith(
-            ('.eyaml', '.yaml', '.yml'),
-        )
-    ):
-        return FileType.YAML
-    return FileType.OTHER
+    _, file_extension = os.path.splitext(filename)
```
OiCMudkips: Does this handle `my.weird.filename`?

detect_secrets/plugins/keyword.py
```diff
@@ -80,53 +81,81 @@
     "\\k.*'",
     '`cat',
     '`grep',
+    '`sudo',
     'account_password',
+    'api_key',
```
OiCMudkips: Why is this in both `DENYLIST` and `FALSE_POSITIVES`?

detect_secrets/plugins/keyword.py
```diff
@@ -80,53 +81,81 @@
...
     'null',
     'null,',
     'null.*"',
     "null.*'",
     'null;',
+    'pass',
+    'pass)',
     'password',
     'password)',
+    'password))',
```
OiCMudkips: Just to clarify, does this mean a word that contains `password))` will be marked as a false-positive?

detect_secrets/plugins/keyword.py
```diff
@@ -147,6 +176,17 @@
         secret=SECRET,
     ),
 )
+FOLLOWED_BY_EQUAL_SIGNS_OPTIONAL_BRACKETS_OPTIONAL_AT_SIGN_QUOTES_REQUIRED_REGEX = re.compile(
+    # e.g. my_password = "bar"
+    # e.g. my_password = @"bar"
+    # e.g. my_password[] = "bar";
```
OiCMudkips: Based on below this is Objective-C. What does this actually do in Objective-C? (I couldn't Google it)

detect_secrets/plugins/keyword.py
```diff
@@ -275,24 +313,26 @@ def secret_generator(self, string, filetype):
...
-                or lowered_secret == 'new'
+        any(
+            false_positive in lowered_secret
+            for false_positive in (
+                '/etc/',
+                'fake',
+                'forgot',
             )
+        ) or lowered_secret in FALSE_POSITIVES
+        or lowered_secret.count('/') >= 3
```
OiCMudkips: Why?

detect_secrets/plugins/common/filetype.py
KevinHock: Yup
```python
>>> os.path.splitext('my.weirdo.filename')
('my.weirdo', '.filename')
```

detect_secrets/plugins/keyword.py
KevinHock: So this is for the case wherein silly code does `magical_api_key: api_key`, or something to that effect, normally in a test.

`DENYLIST` will catch the fact that it could be an API key from `magical_api_key` on the LHS, but then `FALSE_POSITIVES` will check the RHS and cause us not to alert on it.

detect_secrets/plugins/keyword.py
KevinHock: The short-answer is, only if it `==` `password))`, not contains.

The `FALSE_POSITIVES` set is for RHS values normally resulting in false-positives, you could make your real password `password))`, but it's more likely that e.g. I made some mistake in not requiring quotes for that particular filetype, or haven't made a custom regex for that language when I should, this set patches over certain things like that.

detect_secrets/plugins/keyword.py
KevinHock: @ means it is a constant
[] means it is a C style array e.g. `char myString[] = "This is a C character array";`

Something I do not handle yet, but can ticket, are strings like that following
```
password = [str1 stringByAppendingFormat:@"World"];
NSMutableString *password = [NSMutableString stringWithString:@"This string is mutable"];
```

Prior to this PR, we'd capture e.g. `[str1` and `[NSMutableString`, which was 🤐

detect_secrets/plugins/keyword.py
KevinHock: Good catch, added comment in https://github.com/Yelp/detect-secrets/pull/229/commits/5de43e278169ba8f040f6cf24b7b41e134d265fc

detect_secrets/plugins/common/filetype.py
OiCMudkips: Oh right, and does this work with no file extension?

OiCMudkips: Oh it does, the `get` call in the next line has a default

KevinHock: For posterity, it'll be `''` and then get set to `FileType.OTHER`

(I had an internal 🤔 about `setdefault` vs. the `.get` fallback but 🤷‍♂ I kinda liked the way it looked.)

PR #239, opened 2019-09-10
JWT Detector Plugin

gdemarcsek:
# New plugin checking for JWTs

## Motivation
The current high entropy base64 string checker would not find JWTs because they contain dots too. JWTs are more and more widespread - one very good example is Kubernetes API server tokens. The following very typical way of using detect-secrets for example would not detect a kubeconfig file committed in the codebase:

`detect-secrets scan --use-all-plugins --no-keyword-scan`

Of course, Kubernetes is not the only use case and because JWTs have an easy-to-validate format, I thought it would be worth checking for them as well - I think it's reasonable to expect low false positive ratio from this plugin.

## How does it work?
I have implemented a simple regex based plugin that first filters for potential tokens just by looking at the character set and format (see https://tools.ietf.org/html/rfc7519 for details). Then it refines the search by validating the JWT format and encoding (base64url+json). I choose not to include a full JWT library just for this as I do not need to support signature validation or expiry checking here - I only intend to check that the JWT is formally correct. It shouldn't want to be too smart anyways. It is generally not possible to know where that JWT is accepted just by looking at it, so verification by network call is not supported by this plugin.

Test cases are documented in the source code for easier review.

### DIFF
--- a/detect_secrets/plugins/jwt.py
+++ b/detect_secrets/plugins/jwt.py
@@ -8,7 +8,13 @@ import json
 import re
 
 from .base import RegexBasedDetector
-from detect_secrets.core.constants import VerifiedResult
+
+try:
+    # Python 2
+    from future_builtins import filter
+except ImportError:
+    # Python 3
+    pass
 
 
 class JwtTokenDetector(RegexBasedDetector):
@@ -17,7 +23,14 @@ class JwtTokenDetector(RegexBasedDetector):
         re.compile(r'eyJ[A-Za-z0-9-_=]+\.[A-Za-z0-9-_=]+\.?[A-Za-z0-9-_.+/=]*'),
     ]
 
-    def verify(self, token, content=''):
+    def secret_generator(self, string, *args, **kwargs):
+        return filter(
+            self.is_formally_valid,
+            super(JwtTokenDetector, self).secret_generator(string, *args, **kwargs),
+        )
+
+    @staticmethod
+    def is_formally_valid(token):
         parts = token.split('.')
         for idx, part in enumerate(parts):
             try:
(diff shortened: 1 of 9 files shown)


detect_secrets/plugins/jwt.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def secret_generator(self, string, *args, **kwargs):
+        return filter(
+            self.is_formally_valid,
+            super(JwtTokenDetector, self).secret_generator(string, *args, **kwargs),
+        )
+
+    @staticmethod
+    def is_formally_valid(token):
+        parts = token.split('.')
```
KevinHock: Noting for posterity the spec at https://tools.ietf.org/id/draft-jones-json-web-token-04.html#rfc.appendix.Appendix%20B aligns with this code

(Specifically)
```
If the length mod 4 is 0, no padding is added;
if the length mod 4 is 2, two '=' padding characters are added;
if the length mod 4 is 3, one '=' padding character is added;
if the length mod 4 is 1, the input is malformed.
```

detect_secrets/plugins/jwt.py
```diff
@@ -0,0 +1,53 @@
...
+    from future_builtins import filter
+except ImportError:
+    # Python 3
+    pass
+
+
+class JwtTokenDetector(RegexBasedDetector):
+    secret_type = 'JSON Web Token'
+    denylist = [
+        re.compile(r'eyJ[A-Za-z0-9-_=]+\.[A-Za-z0-9-_=]+\.?[A-Za-z0-9-_.+/=]*'),
```
KevinHock: Can you add a `?` after the `*`? I believe this will make it non-greedy i.e. more efficient.

gdemarcsek: @KevinHock Absolutely, and thanks for pointing it out.

PR #245, opened 2019-09-26
Support url-safe base64 secrets

OiCMudkips:
This updates the base64 plugin to support url-safe plugins by just adding `-` and `_` to the charset. However, this won't be merged until I add some additional functionality to reduce noise.

### DIFF
--- a/detect_secrets/plugins/base.py
+++ b/detect_secrets/plugins/base.py
@@ -97,7 +97,7 @@ class BasePlugin(object):
         potential_secrets = {}
         file_lines = tuple(file.readlines())
         for line_num, line in enumerate(file_lines, start=1):
-            results = self.analyze_string(line, line_num, filename)
+            results = self.analyze_line(line, line_num, filename)
             if not self.should_verify:
                 potential_secrets.update(results)
                 continue
@@ -121,7 +121,7 @@ class BasePlugin(object):
 
         return potential_secrets
 
-    def analyze_string(self, string, line_num, filename):
+    def analyze_line(self, string, line_num, filename):
         """
         :param string:    string; the line to analyze
         :param line_num:  integer; line number that is currently being analyzed
@@ -163,7 +163,7 @@ class BasePlugin(object):
     @abstractmethod
     def secret_generator(self, string, *args, **kwargs):
         """Flags secrets in a given string, and yields the raw secret value.
-        Used in self.analyze_string for PotentialSecret creation.
+        Used in self.analyze_line for PotentialSecret creation.
 
         :type string: str
         :param string: the secret to scan
@@ -178,7 +178,7 @@ class BasePlugin(object):
         check what different plugins say regarding a single line/secret. This
         supports that.
 
-        This is very similar to self.analyze_string, but allows the flexibility
+        This is very similar to self.analyze_line, but allows the flexibility
         for subclasses to add any other notable info (rather than just a
         PotentialSecret type). e.g. HighEntropyStrings adds their Shannon
         entropy in which they made their decision.
@@ -191,7 +191,7 @@ class BasePlugin(object):
             <classname>: <returned-value>
         """
         # TODO: Handle multiple secrets on single line.
-        results = self.analyze_string(
+        results = self.analyze_line(
             string,
             line_num=0,
             filename='does_not_matter',
--- a/detect_secrets/plugins/base.py
+++ b/detect_secrets/plugins/base.py
@@ -53,18 +53,34 @@ class BasePlugin(object):
     def secret_type(self):
         raise NotImplementedError
 
-    def __init__(self, exclude_lines_regex=None, should_verify=False, **kwargs):
+    def __init__(
+        self,
+        exclude_lines_regex=None,
+        should_verify=False,
+        false_positive_heuristics=None,
+        **kwargs
+    ):
         """
         :type exclude_lines_regex: str|None
         :param exclude_lines_regex: optional regex for ignored lines.
 
         :type should_verify: bool
+
+        :type false_positive_heuristics: List[Callable]|None
+        :param false_positive_heuristics: List of fp-heuristic functions
+        applicable to this plugin
         """
         self.exclude_lines_regex = None
         if exclude_lines_regex:
             self.exclude_lines_regex = re.compile(exclude_lines_regex)
 
         self.should_verify = should_verify
+
+        self.false_positive_heuristics = (
+            false_positive_heuristics
+            if false_positive_heuristics
+            else []
+        )
 
     @classproperty
     def disable_flag_text(cls):
@@ -231,6 +247,19 @@ class BasePlugin(object):
         :rtype: VerifiedResult
         """
         return VerifiedResult.UNVERIFIED
+
+    def is_secret_false_positive(self, token):
+        """
+        Checks if the input secret is a false-positive according to
+        this plugin's heuristics.
+
+        :type token: str
+        :param token: secret found by current plugin
+        """
+        return any(
+            func(token)
+            for func in self.false_positive_heuristics
+        ) if self.false_positive_heuristics else False
 
     @property
     def __dict__(self):
--- a/detect_secrets/plugins/common/filters.py
+++ b/detect_secrets/plugins/common/filters.py
@@ -80,13 +80,34 @@ def is_sequential_string(secret, *args):
     return False
 
 
-ALL_FALSE_POSITIVE_HEURISTICS = (
+# This only finds UUIDs which only have lowercase characters.
+_UUID_REGEX = re.compile(r'[a-f0-9]{8}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{12}')
+
+
+def is_potential_uuid(secret, *args):
+    """
+    Determines if a potential secret contains any UUIDs.
+
+    :type secret: str
+
+    :rtype: bool
+    Returns True if the string has a UUID, false otherwise.
+    """
+
+    # Using a regex to find strings that look like false-positives
+    # will find us more false-positives than if we just tried validate
+    # the input string as a UUID (for example, if the string has a prefix
+    # or suffix).
+    return len(_UUID_REGEX.findall(secret.lower())) > 0
+
+
+DEFAULT_FALSE_POSITIVE_HEURISTICS = [
     is_found_with_aho_corasick,
     is_sequential_string,
-)
+]
 
 
-# NOTE: this doesn't handle key-values on a line properly.
+# NOTE: this doesn't handle multiple key-values on a line properly.
 # NOTE: words that end in "id" will be treated as ids
 _ID_DETECTOR_REGEX = re.compile(r'[iI][dD][^A-Za-z0-9]')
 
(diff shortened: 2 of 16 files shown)


detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -330,8 +330,15 @@ class Base64HighEntropyString(HighEntropyStringsPlugin):
     secret_type = 'Base64 High Entropy String'
 
     def __init__(self, base64_limit, exclude_lines_regex=None, automaton=None, **kwargs):
+        charset = (
+            string.ascii_letters
+            + string.digits
+            + '+/'  # regular base64
```
KevinHock: Super nit: maybe start capitalized :D 🇨🇻

test_data/config.yaml
```diff
@@ -1,5 +1,5 @@
 credentials:
-    some_value_here: not_a_secret
+    some_value_here: not_secret
```
KevinHock: Was this necessary b/c the entropy calculation with the new chars alerted on `not_a_secret`?

OiCMudkips: Yeah, I don't think we need to be too concerned though because we now have the wordlist filtering.

detect_secrets/plugins/high_entropy_strings.py
OiCMudkips: :+1:

detect_secrets/plugins/common/filters.py
```diff
@@ -97,3 +78,79 @@ def _is_sequential_string(secret, *args):
...
+    """
+    if secret not in line:
+        return False
+
+    secret_index = line.index(secret)
+    return _ID_DETECTOR_REGEX.findall(line, pos=0, endpos=secret_index)
+
+
+ALL_FALSE_POSITIVE_WITH_LINE_CONTEXT_HEURISTICS = [
+    is_likely_id_string,
```
KevinHock: From previous iteration tests, (running it on open-source python repos, adding heuristics to give better [signal stats from `--display-results` output](https://github.com/Yelp/detect-secrets/pull/241/commits/433b75e82347b1c806af3723ae2d8f3c70fc3827)), we added a few heuristics [1] we can add here.

We can add a UUID check when calling `is_false_positive` from `high_entropy_strings.py` and a "value by itself on a line" check to `is_false_positive_with_line_context`. The latter being when files had
```python
some_list_of_ids = [
     'frequently high entropy but not sensitive',
]
```






[1] Hacky code only used in a one-off branch to improve signal, not meant to be merged as written
```python
            rest_of_the_line = string.replace(result, '').lower().strip()
            if any(
                false_positive in rest_of_the_line
                for false_positive in (
                    'id',          # Often not of value
                    're.match',    # Regular expressions have high-entropy
                    're.compile',  # Regular expressions have high-entropy
                    '<script>',    # Often not of value
                )
            ):
                continue

            if len(result) == 36:
                UUID_RE = re.compile(
                    r'^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89AB][0-9a-f]{3}-[0-9a-f]{12}$',
                    re.IGNORECASE,
                )
                if UUID_RE.match(result):
                    continue

            if is_false_positive(result, self.automaton):
                continue

            # Often times IDs are listed in a list.
            # If the secret is on a line alone by itself,
            # it is likely not to be of value
            # I went from 15% signal to 41% with this on one test corpus
            BY_ITSELF_OPTIONAL_COMMENT = re.compile(
                r'^(b)?(\'\'|""),\s*?#',
                re.IGNORECASE,
            )
            if BY_ITSELF_OPTIONAL_COMMENT.match(rest_of_the_line):
                continue
```

KevinHock: That UUID regex may only be for certain kinds of UUIDs, we can always make it better, I just wanted to say that explicitly in case you wanted to improve it, or if it didn't work for a test UUID.

KevinHock: Aside from the heuristics above, I'd like some data backing up that this does improve signal, via [1], otherwise we don't really know if it does what we intend it to do.


[1]
1. Run it on a corpus
1. Run `--display-results`
1. Change the plugin to trim what are clearly false-positives
1. Output Baseline A (only running this plugin)
1. Audit Baseline A
1. Try to trim down false-positives even more
1. Output a new baseline B (only running this plugin)
1. (Err, we don't have code in detect-secrets right now to propagate `is_secret` decisions from Baseline A to Baseline B, I had to write a hacky script to do it, I can share it.)
1. Compare how much of a signal improvement there is
1. See if signal is above 40%
1. Repeat process with new corpus

detect_secrets/plugins/common/filters.py
```diff
@@ -97,3 +78,79 @@ def _is_sequential_string(secret, *args):
...
+
+ALL_FALSE_POSITIVE_HEURISTICS = (
+    is_found_with_aho_corasick,
+    is_sequential_string,
+)
+
+
+# NOTE: this doesn't handle key-values on a line properly.
+# NOTE: words that end in "id" will be treated as ids
+_ID_DETECTOR_REGEX = re.compile(r'[iI][dD][^A-Za-z0-9]')
```
KevinHock: We might be able to do `_id`, we'll see what the data says though.

test_data/config.yaml
KevinHock: I'm more concerned that, we'll have large diffs in baseline's when people update detect-secrets. 

This isn't as concerning as changing a secret type like we did in #26, (where all old secrets were removed and re-added), but it is a little, especially if it reduces TP's to some extent. (We'll see what the data says though, I can't really say how it'll effect signal.)

detect_secrets/plugins/common/filters.py
OiCMudkips: I'm a bit ambivalent on ignoring UUIDs. Based on some googling there's at least some usage of UUIDs as secrets in the wild.

detect_secrets/plugins/common/filters.py
OiCMudkips: I guess it depends on whether we want to ignore keys like `BusinessId`. I think at Yelp this isn't likely but it's probably more likely in `camelCase` language repos.

detect_secrets/plugins/common/filters.py
OiCMudkips: Actually I take that back, looks like I misunderstood the docs I was reading.

detect_secrets/plugins/common/filters.py
KevinHock: That's a good point, we do have a lot of python biases.

detect_secrets/plugins/common/filters.py
```diff
@@ -80,13 +80,34 @@ def is_sequential_string(secret, *args):
...
+
+    :rtype: bool
+    Returns True if the string has a UUID, false otherwise.
+    """
+
+    # Using a regex to find strings that look like false-positives
+    # will find us more false-positives than if we just tried validate
+    # the input string as a UUID (for example, if the string has a prefix
+    # or suffix).
+    return len(_UUID_REGEX.findall(secret.lower())) > 0
```
KevinHock: I would maybe do `bool(_UUID_REGEX.search(secret))` since `findall` might be more than we need

I realize we either do `.lower()` or `.upper()` in each filter function (`.lower` in this func and `is_found_with_aho_corasick`, then `.upper` in `is_sequential_string`, maybe we can DRY that up in `is_false_positive`.

detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -113,7 +115,11 @@ def analyze_string_content(self, string, line_num, filename):
         output = {}
 
         for result in self.secret_generator(string):
-            if is_false_positive(result, self.automaton):
+            # py2+py3 compatible way of copying a list
+            functions = list(DEFAULT_FALSE_POSITIVE_HEURISTICS)
+            functions.append(is_potential_uuid)
+
+            if is_false_positive(result, self.automaton, functions=functions):
```
KevinHock: What are your thoughts on passing `additional_heuristics` instead? I'm not sure when we would want to call `is_false_positive` without the defaults (main motivation is prettifying this part though)

OiCMudkips: I was actually thinking about moving `is_false_positive` to be a method in `BasePlugin` and then make subclass re-implement it. This would allow us to override the filters used on a plugin-level (suggested in #250), but also set some reasonable defaults. In addition we can include the heuristics used in the configs for the plugins in baselines.

i.e.

```
class BasePlugin():
    def __init__(self, false_positive_heuristics=None):
        self.false_positive_heuristics = false_positive_heuristics if false_positive_heuristics else []

    def is_false_positive(self, potential_secret):
         return any(func(potential_secret) for func in self.false_positive_heuristics)

    def get_config(self):
         # include the fp heuristics used if applicable


class Plugin(BasePlugin):
    def __init__(self, false_positive_heuristics=DEFAULT_HEURSTICS_FOR_PLUGIN):  # I remember the default list in Python function constructor, I'll fix it in real code :)
        super(Plugin, self).__init__(false_positive_heuristics)

    def analyze_string_content(self, string):
        for potential_secret in self.secret_generator(string):
            if self.is_false_positive(string):
                continue
```

test_data/config.yaml
OiCMudkips: Why will we have large diffs? A lot of new secrets?

detect_secrets/plugins/high_entropy_strings.py
KevinHock: That sounds great to me 🎈 

I'm only unsure of the `In addition we can include the heuristics used in the configs for the plugins in baselines.` part, as I'm kind of okay with leaving that part blind to the user. (There are also the lesser possible objections someone could say that diffs in baselines should be minimal, and I'm not sure how we would say which heuristics each plugin used in a DRY way.)

detect_secrets/plugins/common/filters.py
OiCMudkips: I just pushed a change to make these regexes case insensitive :)

Good point about the `search` vs `findall` distinction, I will do that

test_data/config.yaml
KevinHock: If all the e.g. `not_a_secret` potential secrets disappear from existing baselines, then there is a possibility we will have large diffs, in the case of FP's that's great, in the case of TP's that would be a regression visible to users. (We can't really say there are minimal regressions without data though.)

detect_secrets/plugins/base.py
```diff
@@ -53,19 +53,33 @@ class BasePlugin(object):
...
         """
         self.exclude_lines_regex = None
         if exclude_lines_regex:
             self.exclude_lines_regex = re.compile(exclude_lines_regex)
 
         self.should_verify = should_verify
 
+        self.false_positive_heuristics = false_positive_heuristics \
+            if false_positive_heuristics \
+            else []
```
KevinHock: If the only use is `is_secret_false_positive` and that has an `if self.false_positive_heuristics` predicate, we probably don't need the `\`
`\`
since it'll be None by default.

If there is a different reason, maybe we can do
```python
self.false_positive_heuristics = (
    false_positive_heuristics
    if false_positive_heuristics
    else []
)
```

PR #251, opened 2019-10-14
Fix signal metric calculation in audit

OiCMudkips:
:floats-triggered:

Before, the results view would show 0% signal for everything, making the metric useless. This PR fixes the calculation to give us an actual number.

Testing done: Manual testing on internal repos


KevinHock:
Ahh @OiCMudkips, Py2 vs. Py3 lol, worked great on Py3 👍

PR #252, opened 2019-10-17
Add one debug option to output raw secret

killuazhu:
Added one debug option flag to output the raw secret in the output. This feature is disabled by default, and it's intended to help troubleshooting.

### DIFF
--- a/detect_secrets/core/potential_secret.py
+++ b/detect_secrets/core/potential_secret.py
@@ -52,9 +52,9 @@ class PotentialSecret(object):
         self.filename = filename
         self.lineno = lineno
         self.set_secret(secret)
-        self.secret_hash = self.hash_secret(secret)
         self.is_secret = is_secret
         self.is_verified = False
+        self.output_raw = output_raw
 
         # If two PotentialSecrets have the same values for these fields,
         # they are considered equal. Note that line numbers aren't included
@@ -99,7 +99,7 @@ class PotentialSecret(object):
         }
 
         if self.output_raw:
-            attributes['secret'] = self.secret
+            attributes['secret'] = self.secret_value
 
         if self.is_secret is not None:
             attributes['is_secret'] = self.is_secret
--- a/detect_secrets/core/potential_secret.py
+++ b/detect_secrets/core/potential_secret.py
@@ -21,6 +21,7 @@ class PotentialSecret(object):
         secret,
         lineno=0,
         is_secret=None,
+        output_raw=False,
     ):
         """
         :type typ: str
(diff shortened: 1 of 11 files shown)


killuazhu:
@KevinHock @domanchi would we be able to get a review on this PR soon?

KevinHock:
@domanchi may have opinions on this one.

I'm not sure if it would help for your use-case, but we did add an `audit --display-results` feature in which you can get all the plaintext secrets from the baseline.

killuazhu:
> I'm not sure if it would help for your use-case, but we did add an `audit --display-results` feature in which you can get all the plaintext secrets from the baseline.

@KevinHock Thanks for pointing out that. I was not aware of `audit --display-results`. After evaluating its output, it does not look like what we want.

Essentially we want to extract the raw secrets from scanning results. For multi factor tokens, it would include all the factors. That's why we are proposing #252 and one more yet to come PR to also output other factors. Since we are in a control Github instance, combing this with pre-receive hook allows us to find tokens accidentally leaked by our users really well.

detect_secrets/core/potential_secret.py
```diff
@@ -68,6 +73,10 @@ def set_secret(self, secret):
         #       we don't want to create a file that contains all plaintext secrets
         #       in the repository.
         self.secret_value = secret
+        # If two PotentialSecrets have the same values for these fields,
+        # they are considered equal. Note that line numbers aren't included
+        # in this, because line numbers are subject to change.
+        self.fields_to_compare = ['filename', 'secret_hash', 'type']
```
domanchi: Hmm. Why is this duplicated?

detect_secrets/core/usage.py
```diff
@@ -209,6 +210,19 @@ def _add_adhoc_scanning_argument(self):
...
+        self.parser.add_argument(
+            '--debug-output-raw',
+            action='store_true',
+            help=(
+                'Outputs the raw secret in the baseline file.'
+                'For development/debug purposes.'
+                'Do NOT use this option in a repo scanning context.'
+            ),
+        )
+        return self
```
domanchi: If we add debug options, we should put this under a different argparse category. e.g. https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/usage.py#L336

This way, we can clearly distinguish debug options, and put a large clear warning that tells users not to use it during production runs.

detect_secrets/core/potential_secret.py
killuazhu: Unclean rebase. Will remove

PR #253, opened 2019-10-17
Add option to display verified false result

killuazhu:
On top of #252, add one flag `--debug-output-verified-false` to support output the entries which fail to validate. It's also intended to help with troubleshooting.

This PR needs to be rebased once $252 has been approved and merged.

### DIFF
--- a/detect_secrets/core/baseline.py
+++ b/detect_secrets/core/baseline.py
@@ -22,6 +22,7 @@ def initialize(
     word_list_hash=None,
     should_scan_all_files=False,
     debug_output_raw=False,
+    output_verified_false=False,
 ):
     """Scans the entire codebase for secrets, and returns a
     SecretsCollection object.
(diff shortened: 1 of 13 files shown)


domanchi:
Hi @killuazhu,

Man, the further I go back, the more I realize how long it's been since I've touched this repository! I'm curious to learn more about your use case for this (if still relevant): would comparing a generated baseline with `--no-verify` and one without meet the same needs? The less marketed `audit --diff` functionality gives a decent way of interactively checking baselines, if it's intended for debugging purposes.

killuazhu:
@domanchi haha, it's definitely been awhile since we sync. 

> Comparing the baseline with and without `--no-verify` option

I think the result would be similar, but that does require running the scan on the same file twice. I just check our code base and noticed we are not actively using this debug option at the moment. Prod env is just running normal verify scan.

I'm fine with closing this PR since it's not been used now!

PR #254, opened 2019-10-18
Add Detector for Softlayer tokens

killuazhu:
Add Detector for Softlayer tokens, it supports looking for 2 factors: username and password. It also supports verification.


requirements-dev.txt
```diff
@@ -9,3 +9,4 @@ pyyaml
 tox>=3.8
 tox-pip-extensions
 unidiff
+responses
```
KevinHock: Maybe put after `pyyaml` to keep the requirements alphabetical.

tests/plugins/softlayer_test.py
```diff
@@ -0,0 +1,167 @@
...
+                [
+                    SL_USERNAME,
+                    'test2@testy.test',
+                    'test3@testy.testy',
+                    'notanemail',
+                ],
+            ),
+        ),
+    )
+    def test_find_username(self, content, expected_output):
```
KevinHock: Just in case you find that this type of secret is used in function calls, the following commit may help you add detection for that. (Code still in review --  https://github.com/Yelp/detect-secrets/pull/255/commits/c6037d998fc3b03fb2662b1df178a062d6566705)

detect_secrets/plugins/softlayer.py
```diff
@@ -0,0 +1,77 @@
...
+        for username in usernames:
+            return verify_softlayer_key(username, token)
+
+        return VerifiedResult.VERIFIED_FALSE
+
+
+def find_username(content):
+    # opt means optional
+    username_keyword = r'(?:username|id|user|userid|user-id|user-name|' + \
+        r'name|user_id|user_name|uname)'
```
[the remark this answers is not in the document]
KevinHock: Hmm, GitHub suggestions merged my line 46 and 47 suggestions :/ Mainly meant to say I prefer parenthesis over `\`, feel free to ignore though.

detect_secrets/plugins/base.py
```diff
@@ -260,6 +260,39 @@ class FooDetector(RegexBasedDetector):
     def denylist(self):
         raise NotImplementedError
 
+    @staticmethod
+    def assign_regex_generator(prefix_regex, password_keyword_regex, password_regex):
```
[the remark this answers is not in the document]
killuazhu: I renamed `password_keyword_regex ` -> `secret_keyword_regex`, `password_regex ` -> `secret_regex`. I have left out the `prefix_regex`. Let me know if that's what you mean.

detect_secrets/plugins/softlayer.py
```diff
@@ -0,0 +1,77 @@
...
+        response = requests.get(
+            'https://api.softlayer.com/rest/v3/SoftLayer_Account.json',
+            auth=(username, token), headers=headers,
+        )
+
+        if response.status_code == 200:
+            return VerifiedResult.VERIFIED_TRUE
+        else:
+            return VerifiedResult.VERIFIED_FALSE
+    except Exception:
```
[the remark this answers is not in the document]
killuazhu: > call can throw, or just requests.exceptions.RequestException. 

Good point.

> (Up to you if you want to try retrying on e.g. requests.exceptions.Timeout)

I will leave retry out.

> For posterity, this made me think about if we want to do this for other plugins like Slack Stripe etc

Wrapping some exception would be necessary, otherwise with current code, if Slack, Stripe or AWS verification endpoint is down or unreachable for whatever reason, we will throw exception instead of gracefully report unverified. As you said, it can force user to use `--no-verify` to keep using d-s.

detect_secrets/plugins/softlayer.py
```diff
@@ -0,0 +1,77 @@
...
+
+        for username in usernames:
+            return verify_softlayer_key(username, token)
+
+        return VerifiedResult.VERIFIED_FALSE
+
+
+def find_username(content):
+    # opt means optional
+    username_keyword = r'(?:username|id|user|userid|user-id|user-name|' + \
```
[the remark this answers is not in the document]
killuazhu: Using `()` looks good. I will update.

requirements-dev.txt
killuazhu: Adjusted.

tests/plugins/softlayer_test.py
killuazhu: Unlike AWS tokens which has a specific format, the Softlayer username here is almost free format. It might be hard to match them in function calls. I will prefer to leave them out for now.

killuazhu: Thanks for the suggestions. @KevinHock. Unlike AWS tokens which has a specific format, the Softlayer username here is almost free format. It might be hard to match them in function calls. I will prefer to leave them out for now.

killuazhu:
@KevinHock all comments addressed.

PR #255, opened 2019-10-18
Bring-your own-plugins (BYOP), via --custom-plugins option

KevinHock:
In case you wanted to make e.g. a new entropy plugin with different metacharacters, or a new keyword type detector, or a company-specific plugin, or use it/tune it privately before merging upstream.

To be used like so: `detect-secrets scan  --custom-plugins custom_plugins/ test_data/each_secret.py`

We had a "[truffleHog](https://github.com/dxa4481/truffleHog) vs. [metasploit](https://github.com/rapid7/metasploit-framework)"  discussion internally. With the former, you can e.g. edit a config file with regexes in it like [`shhgit`](https://github.com/eth0izzle/shhgit/blob/master/config.yaml), with the latter you write your own modules.

Main reasons why this seems like a better fit for `detect-secrets`
- More general purpose
- We would still need verification functions written in Python if we put all our regexes in a config
- This encourages writing tests


detect_secrets/core/usage.py
```diff
@@ -44,6 +65,36 @@ def add_no_verify_flag(parser):
...
+    The pre-commit hook gets e.g. `--no-jwt-scan` type options
+    as well as the subparser for `detect-secrets scan`.
+
+    :rtype: argparse.ArgumentParser
+    :returns: argparse.ArgumentParser to pass into PluginOptions
+    """
+    if parser.prog == 'detect-secrets-hook':
+        return parser
+
+    for action in parser._actions:  # pragma: no cover (Always returns)
```
KevinHock: Bottom answer of https://stackoverflow.com/questions/43688450/how-to-obtain-argparse-subparsers-from-a-parent-parser-to-inspect-defaults had some great pointers, works in both 2 and 3.

KevinHock:
I need to edit the README in this PR still, but might do that after the review.

detect_secrets/core/audit.py
```diff
@@ -339,7 +346,14 @@ def print_audit_results(baseline_filename):
 def _get_baseline_from_file(filename):  # pragma: no cover
     try:
         with open(filename) as f:
-            return json.loads(f.read())
+            baseline = json.loads(f.read())
+
+            # Convert custom_plugin_paths to a tuple so it is later hashable
+            # We want this so we can @lru_cache functions called with this
+            if 'custom_plugin_paths' in baseline:
+                baseline['custom_plugin_paths'] = tuple(baseline['custom_plugin_paths'])
```
KevinHock: Maybe I should just decorate the 2 funcs that take this as an arg 🤔

detect_secrets/core/usage.py
```diff
@@ -78,10 +127,36 @@ def add_console_use_arguments(self):
         return self
 
     def parse_args(self, argv):
-        output = self.parser.parse_args(argv)
-        PluginOptions.consolidate_args(output)
+        # We temporarily remove '--help' so that we can give the full
```
OiCMudkips: I'm not sure if I understand this, can you explain?

detect_secrets/core/usage.py
```diff
@@ -78,10 +127,36 @@ def add_console_use_arguments(self):
...
+                ),
+                argv,
+            ),
+        )
+        known_args, _ = self.parser.parse_known_args(
+            args=argv_without_help,
+        )
 
-        return output
+        # Audit does not use --custom-plugins
```
OiCMudkips: Why not? I'm not sure if that's how I read this PR.

KevinHock: I can amend to `# Audit does not use --custom-plugins, it pulls custom_plugins from the audited baseline`?

detect_secrets/core/usage.py
KevinHock: If you do `--help` with `--custom-plugins`, you want to see the options for e.g. `--no-custom-detector`, if we just parse `--help` before `--custom-plugins`, it will SystemExit and give the help message without any options from the custom plugins.

detect_secrets/plugins/common/util.py
```diff
@@ -8,50 +8,81 @@
 from importlib import import_module
 
 from detect_secrets.plugins.base import BasePlugin
-from detect_secrets.util import get_root_directory
 
 
+def _change_custom_plugin_paths_to_tuple(custom_plugin_paths_function):
```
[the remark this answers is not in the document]
KevinHock: I generally prefer genericized naming too, but as written the argument has to be named `custom_plugin_paths_function`, otherwise it won't work, so I'm ambivalent. Maybe I can try to make it so the decorator tuplifies all arguments 🤔 , ++.

If we put it in the function body, I don't think we could `lru_cache` it, right?

detect_secrets/plugins/common/util.py
```diff
@@ -8,50 +8,81 @@
...
+            # Do not traverse more than 1 level
+            break
 
     plugins = {}
-    for module_name in modules:
-        module = import_module('detect_secrets.plugins.{}'.format(module_name))
-        for name in filter(lambda x: not x.startswith('_'), dir(module)):
-            plugin = getattr(module, name)
+    for module_name in plugin_modules:
+        module = import_module((module_name))
```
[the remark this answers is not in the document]
KevinHock: Yes

tests/core/audit_test.py
```diff
@@ -243,6 +243,7 @@ def mock_env(self, user_inputs=None, baseline=None):
     @property
     def baseline(self):
         return {
+            'custom_plugin_paths': (),
```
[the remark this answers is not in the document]
KevinHock: Great catch! It was originally a tuple, so I must have missed a spot 👍

detect_secrets/core/usage.py
```diff
@@ -27,6 +28,26 @@ def add_word_list_argument(parser):
     )
 
 
+def add_custom_plugins_argument(parser):
+    parser.add_argument(
+        '--custom-plugins',
+        action='append',
+        default=[],
+        dest='custom_plugin_paths',
+        help='Paths containing custom plugin Python files, not searched recursively.',
```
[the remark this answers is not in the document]
KevinHock: Sure, that makes sense, ++

detect_secrets/plugins/common/util.py
```diff
@@ -8,50 +8,81 @@
...
-        for filename in files:
-            if not filename.startswith('_'):
-                modules.append(os.path.splitext(filename)[0])
-
-        # Only want to import top level files
-        break
+    plugin_paths = ['detect_secrets/plugins'] + list(custom_plugin_paths)
+    plugin_modules = []
+    for plugin_path in plugin_paths:
+        for root, _, files in os.walk(  # pragma: no cover (Always breaks)
```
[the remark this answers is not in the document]
KevinHock: Most definitely :)

detect_secrets/core/secrets_collection.py
```diff
@@ -107,16 +113,22 @@ def load_baseline_from_dict(cls, data):
             result.word_list_hash = data['word_list']['hash']
 
             if result.word_list_file:
-                # Always ignore the given `data['word_list']['hash']`
+                # Always ignore the existing `data['word_list']['hash']`
                 # The difference will show whenever the word list changes
                 automaton, result.word_list_hash = build_automaton(result.word_list_file)
 
+        # In v0.12.8 the `--custom-plugins` option got added
+        result.custom_plugin_paths = []
```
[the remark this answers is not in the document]
KevinHock: Great catch, I caught this too after reviewing it again, reminded me of bloodhound code.

detect_secrets/plugins/common/util.py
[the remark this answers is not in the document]
OiCMudkips: sg2m

detect_secrets/plugins/common/util.py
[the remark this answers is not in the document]
OiCMudkips: I'm not sure if I understand why the parameter has to be named `custom_plugin_paths_function`.

I don't know where the `lru_cache` is applied right now, so we can leave it as a decorator if it works as intended already.

detect_secrets/core/usage.py
OiCMudkips: ++, this comment makes more sense

detect_secrets/plugins/common/util.py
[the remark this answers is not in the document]
KevinHock: *Ouch my bad @OiCMudkips, I mis-typed last Friday, the argument to the function being decorated has to be named `custom_plugin_paths`, as it's currently written. So it's not applicable to any function or any function with 1 arg, only 1 arg func's with `custom_plugin_paths` as an arg.

I will try to make it so that it can accept any 1 arg func, so we can name it a more genericized name though.

detect_secrets/pre_commit_hook.py
```diff
@@ -19,13 +19,13 @@
...
 
-def parse_args(argv):
+def parse_args(argv):  # pragma: no cover (Mocked)
     return ParserBuilder()\
         .add_pre_commit_arguments()\
         .parse_args(argv)
 
 
-def main(argv=None):
+def main(argv=sys.argv[1:]):
```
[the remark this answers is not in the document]
KevinHock: Must have missed this last Friday, so by default `argparse` looks at `sys.argv`, if what you pass to it when calling `parse_args()` is None, since I wanted to check both provided arguments, and sys.argv, I explicitly made `argv` come from `sys.argv` when nothing is provided.

detect_secrets/plugins/aws.py
```diff
@@ -43,14 +43,16 @@ def verify(self, token, content):
 
 def get_secret_access_keys(content):
     # AWS secret access keys are 40 characters long.
+    # e.g. some_function('AKIA...', '[secret key]')
```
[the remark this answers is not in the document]
KevinHock: Sure, before we were only looking for secret keys in assignments, whereas this adds looking for them as function arguments too. The test cases I added might shed light on this.

detect_secrets/plugins/common/util.py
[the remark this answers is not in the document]
KevinHock: re: `where the lru_cache is applied`
I wrap functions that accept a list, to turn them into functions that accept a tuple, before the `lru_cache` decorator gets to them, so that they can be cached. Lists are unacceptable for immutability reasons IIRC

detect_secrets/plugins/common/util.py
```diff
@@ -34,7 +39,63 @@ def get_mapping_from_secret_type_to_class_name(custom_plugin_paths):
...
+    # addresses for the same functions as statically
+    # imported classes do, so issubclass does not work.
+    # We use __mro__ to loop through all parent classes.
+    # issubclass does work in Python 3, since parent classes
+    # do have the same addresses.
+    if not any(
+        'plugins.base.BasePlugin' in str(klass)
+        for klass in attr.__mro__
+    ):
+        return False
```
KevinHock: Due to the issue explained in this comment, `super` calls in Python for plugins no longer work!

e.g. the `super` call in 
```python
    @property
    def __dict__(self):
        output = super(HighEntropyStringsPlugin, self).__dict__
        output.update({
            'hex_limit': self.entropy_limit,
        })

        return output
```
will throw a `super(type, obj): obj must be an instance or subtype of type` error because it thinks `self` is not a subtype, when really it just has different addresses for its functions. I patched around the same issue with `issubclass` in this highlighted code, but with `super` there is not a desirable path forward that I can see, I shall ask on StackOverflow for advice.

KevinHock: For posterity, I did look into this, although I time-boxed myself.

You can change the metaclass of BasePlugin to be e.g.
```diff
+class AlwaysTrue(ABCMeta):
+    def __instancecheck__(self, instance):
+    def __subclasscheck__(self, instance):
```
however, super still throws the above error, only `isinstance` and `issubclass` work.

In other words, I believe this is definitely blocked on dropping Python 2 support https://github.com/Yelp/detect-secrets/issues/260. Either that or have the user supply the import path as well as the file path, but I don't want to do that.

detect_secrets/core/usage.py
```diff
@@ -44,6 +74,36 @@ def add_no_verify_flag(parser):
...
+    :rtype: argparse.ArgumentParser
+    :returns: argparse.ArgumentParser to pass into PluginOptions
+    """
+    if parser.prog == 'detect-secrets-hook':
+        return parser
+
+    for action in parser._actions:  # pragma: no cover (Always returns)
+        if isinstance(action, argparse._SubParsersAction):
+            for subparser in action.choices.values():
+                if subparser.prog.endswith('detect-secrets scan'):
```
killuazhu: Is this for testing purpose? The main reason I ask is because during plugin dev phase, I would use something like below to test out the code without installing it.

```shell
env PYTHONPATH=/Users/username/workspace/detect-secrets python /Users/username/workspace/detect-secrets/detect_secrets/main.py scan tmp/testfile.txt
```

The hard code part here ruined the execution of my case.

killuazhu:
> To be used like so: detect-secrets scan --custom-plugins custom_plugins/ test_data/each_secret.py

In general it sounds like a good idea. Have we thought about how to deal with things like additional pip packages required by new plugins? 

Another thing is it would be nice to include a helloworld or sample plugin in the sample_custom_plugins folder, so others can easily follow and start developing new plugins.

detect_secrets/core/usage.py
KevinHock: I wonder if I change it to just `endwith('scan')`, if it'll work, it probably does.

killuazhu: You are right, changing it `endwith('scan')` is working! @KevinHock

KevinHock:
>Have we thought about how to deal with things like additional pip packages required by new plugins?

For the server-side case it's easy to include any packages you need next to e.g. `detect-secrets( or -server)`. But for the pre-commit hook case, I don't think it's desirable to try to pip install packages dynamically when someone runs `detect-secrets`, since it's so unexpected.

I did include a couple custom plugins in the tests, but those are kind of too basic. Since we [link to the plugins/ dir in the README right before we mention `--custom-plugins`](https://github.com/Yelp/detect-secrets#currently-supported-plugins) I think most users will be able to look at that directory for inspiration. 

We should also add a link to the right `contributing.md` section though](https://github.com/Yelp/detect-secrets/blob/master/CONTRIBUTING.md#writing-a-plugin), I can add that.

killuazhu:
> I did include a couple custom plugins in the tests, but those are kind of too basic. Since we link to the plugins/ dir in the README right before we mention --custom-plugins I think most users will be able to look at that directory for inspiration.

I think this should work. Given some of the relative imports might be broken (e.g. [example](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/aws.py#L16)) depends on who user structure their custom import folders.

> We should also add a link to the right contributing.md section though](https://github.com/Yelp/detect-secrets/blob/master/CONTRIBUTING.md#writing-a-plugin), I can add that.

👍

detect_secrets/core/secrets_collection.py
```diff
@@ -42,12 +46,17 @@ def __init__(
         :param word_list_hash: optional iterated sha1 hash of the words in the word list.
         """
         self.data = {}
+        self.version = VERSION
+
         self.plugins = plugins
+        self.custom_plugin_paths = (
```
KevinHock: Note: Improve this to be `= custom_plugin_paths or []`

detect_secrets/core/usage.py
```diff
@@ -44,6 +74,36 @@ def add_no_verify_flag(parser):
     )
 
 
+def add_shared_arguments(parser):
+    """
+    These are arguments that are for both
+    pre-commit and regular detect-secrets scripts.
```
KevinHock: Nit: `and command-line detect-secrets`. Is more clear than regular.

detect_secrets/plugins/common/util.py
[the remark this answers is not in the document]
KevinHock: Following up with this: the reason the arg must be named `custom_plugin_paths`, is because any function decorated by this decorator may be called with `custom_plugin_paths` as a keyword argument. 

i.e. If I changed this to
```python
    def wrapper_of_custom_plugin_paths_function(one_arg):
        return custom_plugin_paths_function(tuple(one_arg))
```

it would result in
```
============================================== ERRORS ===============================================
________________________________ ERROR collecting tests/main_test.py ________________________________
tests/main_test.py:76: in <module>
    class TestMain:
tests/main_test.py:294: in TestMain
    'name': 'Base64HighEntropyString',
tests/main_test.py:33: in get_list_of_plugins
    for name, plugin in import_plugins(custom_plugin_paths=[]).items():
E   TypeError: wrapper_of_custom_plugin_paths_function() got an unexpected keyword argument 'custom_plugin_paths'
```

tests/plugins/common/initialize_test.py
```diff
@@ -10,37 +10,45 @@ class TestFromPluginClassname:
...
+            custom_plugin_paths=[],
             hex_limit=4,
         )
 
-        assert isinstance(plugin, HexHighEntropyString)
+        # Dynamically imported classes have different
+        # addresses for the same functions as statically
+        # imported classes do, so isinstance does not work.
+        assert str(plugin.__class__) == str(HexHighEntropyString)
+        assert dir(plugin.__class__) == dir(HexHighEntropyString)
```
KevinHock, line 22: Had to do
```
-        assert all(
-            func(plugin.__class__) == func(Base64HighEntropyString)
-            for func in (str, dir)
-        )
+        assert str(plugin.__class__) == str(Base64HighEntropyString)
+        assert dir(plugin.__class__) == dir(Base64HighEntropyString)
```
because otherwise it was `94%   21->exit, 71->exit`

detect_secrets/core/usage.py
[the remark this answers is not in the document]
KevinHock: Noting that it now works for directories and files.

KevinHock:
I rebased with master so this should be ready to review now @OiCMudkips / @killuazhu 😁

KevinHock:
Polite re-ping on this one 💟 I have some fun funemployed time at the moment.

detect_secrets/core/audit.py
```diff
@@ -283,13 +290,13 @@ def determine_audit_results(baseline, baseline_path):
         audit_results['stats'][audit_result]['count'] += 1
         audit_results['stats'][audit_result]['files'][filename] += 1
         total += 1
-    audit_results['stats']['signal'] = str(
+    audit_results['stats']['signal'] = '{:.2f}'.format(
```
OiCMudkips, line 293: The `%` go into this format string

detect_secrets/pre_commit_hook.py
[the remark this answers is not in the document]
OiCMudkips: Makes sense.

detect_secrets/plugins/aws.py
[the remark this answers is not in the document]
OiCMudkips: Makes sense.

detect_secrets/plugins/common/util.py
```diff
@@ -1,53 +1,149 @@
...
 import os
 from abc import abstractproperty
 from functools import lru_cache
-from importlib import import_module
 
 from detect_secrets.plugins.base import BasePlugin
 from detect_secrets.util import get_root_directory
 
 
+def change_custom_plugin_paths_to_tuple(custom_plugin_paths_function):
```
OiCMudkips, line 11: This seems kinda excessive, can we just call `tuple(argument)` wherever we want to try to be safer?

One alternative is to make it so that the `ArgumentParser` returns a `tuple` by implementing a [custom action](https://docs.python.org/3.7/library/argparse.html#action) but honestly I'd be OK with us just not worry about the accidentally mutating the list of paths.

detect_secrets/core/audit.py
KevinHock: Great catch!! I no longer need the `[:4]` so that makes a lot of sense 🎊

detect_secrets/plugins/common/util.py
KevinHock: The reason for tuplifying is [so that we can `lru_cache` the functions this decorator is used on](https://github.com/Yelp/detect-secrets/pull/255#discussion_r339195196), not b/c I'm afraid the list will get mutated.

I'll re-read the code to see if tuplifying as soon as we parse args (/the baseline) makes things easier. 👍 Convert to X as early as possible seems like always good advice, I'll have to see if there was a reason I wrote it this way. 🤔

KevinHock: It does not appear there was a good reason 😁 I'll go ahead and do the custom action, something similar to https://stackoverflow.com/questions/50365835/how-to-make-argparse-argumentparser-return-a-tuple-or-an-np-array-rather-than-a

KevinHock:
Made new PR due to change in GitHub permissions. So closing this one.

PR #261, opened 2019-10-30
Contribute cloudant plugin

killuazhu:
Add one more plugin to detect [Cloudant tokens](https://cloud.ibm.com/docs/services/Cloudant?topic=cloudant-authentication).


detect_secrets/plugins/cloudant.py
```diff
@@ -0,0 +1,131 @@
...
+        headers = {'Content-type': 'application/json'}
+        request_url = 'https://{hostname}:' \
+            '{token}' \
+            '@{hostname}.' \
+            'cloudant.com'.format(
+                hostname=hostname,
+                token=token,
+            )
+
+        response = requests.get(
```
KevinHock: Can you do the specific exception thing here?

Rest of the PR ✅ ⛵️ 🚢 👍

killuazhu:
> I didn't realize this in the other PR, but it is a little confusing that the numbers referenced in your commit messages map to unrelated issue/PR #'s, presumably b/c they're internal numbers.

Yes, we use Github Enterprise internally, which might mess up with the PR. I can remove the PR id from commit message to avoid confusion.

detect_secrets/plugins/cloudant.py
killuazhu: Comment addressed.

killuazhu:
Pushed again with updated commit message.

killuazhu:
Retrigger Travis build.

PR #263, opened 2019-11-01
Add plugin for IBM Cloud Object Storage HMAC

killuazhu:
Add plugins to detect the [HMAC credential](https://cloud.ibm.com/docs/services/cloud-object-storage/hmac?topic=cloud-object-storage-hmac) for IBM Cloud Object Storage.


detect_secrets/plugins/ibm_cos_hmac.py
```diff
@@ -0,0 +1,163 @@
...
+
+        if not key_id_matches:
+            return VerifiedResult.UNVERIFIED
+
+        try:
+            for key_id in key_id_matches:
+                verify_result = verify_ibm_cos_hmac_credentials(
+                    key_id, token,
+                )
+                if verify_result is True:
```
domanchi: Probably don't need the `is True` portion of this.

killuazhu: Comment addressed in latest commit.

killuazhu: Comment addressed in latest commit.

PR #267, opened 2019-11-09
Create plugin to detect Twilio API keys.

EdOverflow:
This commit contains a `RegexBasedDetector` plugin for Twilio API keys. Twilio's API requires an "account SID" and "auth token" which can be detected by this plugin using two regular expressions.

```
Example account SID: ACxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Regex: AC[a-z0-9]{32}

Example auth token: SKxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Regex: SK[a-z0-9]{32}
```

It might be worth further expanding this plugin to include a verify function that sends a GET request to `{}:{}@api.twilio.com/2010-04-01/Accounts.json`.

### DIFF
--- a/tests/plugins/twilio_test.py
+++ b/tests/plugins/twilio_test.py
@@ -23,4 +23,4 @@ class TestTwilioKeyDetector(object):
     def test_analyze(self, payload, should_flag):
         logic = TwilioKeyDetector()
         output = logic.analyze_line(payload, 1, 'mock_filename')
-        assert len(output) == int(should_flag)
+        assert output
--- a/tests/plugins/twilio_test.py
+++ b/tests/plugins/twilio_test.py
@@ -12,17 +12,15 @@ class TestTwilioKeyDetector(object):
         [
             (
                 'SKxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
-                True
+                True,
             ),
             (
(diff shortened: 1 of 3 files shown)


tests/plugins/twilio_test.py
```diff
@@ -0,0 +1,26 @@
...
+            (
+                'ACxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
+                True,
+            ),
+        ],
+    )
+    def test_analyze(self, payload, should_flag):
+        logic = TwilioKeyDetector()
+        output = logic.analyze_line(payload, 1, 'mock_filename')
+        assert len(output) == int(should_flag)
```
domanchi: Probably can just do:

```
assert output
```

as that will make sure that it is a non-empty dictionary.

EdOverflow:
@domanchi, do you think we should add some form of key verification or just leave it as is?

KevinHock:
Just realized it's @EdOverflow :) Thank you for contributing!

EdOverflow:
:)

domanchi:
> @domanchi, do you think we should add some form of key verification or just leave it as is?

Naw. Key verification can always be done at a later iteration. :)

PR #277, opened 2020-02-25
Add Python 3.8 to Travis

gliptak:

Issue #260, opened 2019-10-29
Deprecate Python 2 support?


KevinHock:
_This needs more investigation, labeling as question and making the issue anyway_

Anyone have strong objections to this?

How do you run a Python 3 pre-commit hook from a Python 2 repo? (i.e. what would the migration look like when upgrading post-deprecation.)
https://pre-commit.com/#overriding-language-version seems straightforward, but what about e.g. `pre-commit autoupdate`?

domanchi:
https://github.com/Yelp/detect-secrets/pull/292 enacts this change. Needless to say, this is a breaking change for a number of people in our community. However, we've had several internal discussions about this, and in the spirit of transparency, I wanted to share our summarized reasons:

#### 1. Python2.7 has been marked EOL.

As regular python users may know, Python2.7 has been marked for EOL on January 1st, 2020. However, enterprise users (the audience that this tool was designed for) may have a harder time upgrading their internal systems to use Python3.

Despite this, we have confidence that our enterprise users will eventually move to a newer Python version if they haven't already. Therefore, we will move to Python3 to do continue to keep our systems and dependents off no-longer-supported Python versions.

#### 2. We've attempted to solicit community feedback for other reasons to postpone this change.

This thread has been opened for six months, and we have not heard anything from the community. This seems like a sufficiently long public forum to gather feedback, given other related factors, which gives us reason to believe that this change may affect less people than we may expect.

#### 3. This modification blocks further feature development.

Lastly, there are several features that we want to implement that are made significantly harder when needing to support Python2. This change will allow the creation of more features for this tool, so that the community can continue to benefit from its developments.

---

For other members in the community that still need Python2.7 support, don't fret! To our knowledge, this change should only affect you if use `detect-secrets` as an importable module. And in those cases, you will still be able to install it via `pip install detect-secrets==0.13.1`.

KevinHock:
Closing this as #292 is now merged. 

To add to the above comment, I suspect large companies have their own fork, and medium/small companies have more agility to move to away from Python 2. That is just educated conjecture though.

PR #278, opened 2020-02-25
Remove Python 2.7 from Travis

gliptak:
https://github.com/Yelp/detect-secrets/pull/277


KevinHock:
Linking to #260, as this is a first-step

PR #287, opened 2020-03-25
Supporting multiword adhoc string scans

domanchi:
## Issue

I wanted to perform an adhoc scan for a line of code, using the `scan --string` option. However, I came across this strange behavior:

```
$ detect-secrets scan --string 'aHR0cHM6Ly93d3cueW91dHViZS5jb20vd2F0Y2g/dj1vSGc1U0pZUkhBMA=='
Base64HighEntropyString: True  (5.023)
$ detect-secrets scan --string 'aHR0cHM6Ly93d3cueW91dHViZS5jb20vd2F0Y2g/dj1vSGc1U0pZUkhBMA== blah'
Base64HighEntropyString: False
```

## Solution

Support multi-word scans through this functionality. We maintain strict backwards compatibility, so this should be a sole addition. There may be some confusion as to what the entropy value returned from a multi-word string refers to, but it seems like this is a better option than adding another `--line` functionality to handle this case.

### DIFF
--- a/detect_secrets/plugins/high_entropy_strings.py
+++ b/detect_secrets/plugins/high_entropy_strings.py
@@ -144,7 +144,7 @@ class HighEntropyStringsPlugin(BasePlugin):
     def adhoc_scan(self, string):
         # Since it's an individual string, it's just bad UX to require quotes
         # around the expected secret.
-        with self.non_quoted_string_regex():
+        with self.non_quoted_string_regex(is_exact_match=False):
             results = self.analyze_line(
                 string,
                 line_num=0,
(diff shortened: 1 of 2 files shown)


detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -153,24 +153,50 @@ def adhoc_scan(self, string):
 
             # NOTE: Trailing space allows for nicer formatting
             output = 'False' if not results else 'True '
-            if self.regex.search(string):
+            if results:
                 output += ' ({})'.format(
-                    round(self.calculate_shannon_entropy(string), 3),
+                    round(
+                        self.calculate_shannon_entropy(
+                            list(results.keys())[0].secret_value,
```
KevinHock: Nit: Maybe add a `# There can only be 1 secret in the results` comment

Issue #288, opened 2020-03-26
IBM COS HMAC Detector False Positives?


domanchi:
When testing these new plugins, I noticed this plugin was capturing more than what I'd expect it to capture.

### Reproduction Steps

```
$ cat aws_credentials.json
{
    "accessKeyId": "AKIAIOSFODNN7EXAMPLE",
    "secretAccessKey": "81442db33ceee1092586feb635d7b3c5d72257c775ee323414bac8e673d41af3"
}
$ detect-secrets scan aws_credentials.json > baseline
$ detect-secrets audit baseline
```

It looks like the regex is capturing `81442db33ceee1092586feb635d7b3c5d72257c775ee3234`, rather than the full string.

### Proposed Fix

@killuazhu : Does this [password regex](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/ibm_cos_hmac.py#L25) need to be `$` terminated?

killuazhu:
@domanchi That would be a good idea. Would put up a fix for this.

PR #289, opened 2020-03-27
Terminate token matching

killuazhu:
Fixes https://github.com/Yelp/detect-secrets/issues/288

Terminate the string matching after required number of characters.


killuazhu:
@domanchi added a fix for the issue #288 you raised.

Issue #294, opened 2020-04-17
Option to scan through git history


FYJen:
Love this tool! Really friendly to use.

Does it support git history scanning? From the description, it doesn't look like the tool supports it as a base line file is generated from the current state of a repository. I am wondering if there will be any feature like git history scanning for the users who adopt this tool into their own projects the first time. It will be particularly helpful if I can use this tool to also scan through the git history before open-sourcing a project.

domanchi:
Hi @FYJen, glad to hear you like it!

This tool specifically doesn't do git history scanning, since that's not what it's really designed for. It's designed to be fast, breadth-based scanning, rather than uncovering all the secrets that could have been in your git repository (depth-based).

If you're looking for a tool like that, I'd recommend https://github.com/dxa4481/truffleHog.

KevinHock:
Just want to say, I might do this somewhere down the line, if consensus can't be reached about adding a [3rd console script](https://github.com/Yelp/detect-secrets/blob/0e8925d685f1fe34a1e315045450cd678104648c/setup.py#L31-L34) to do this, I can make a separate package, that has detect-secrets as a dependency, and [call `scan_diff` like detect-secrets-server does](https://github.com/Yelp/detect-secrets-server/blob/f831d9bcb0c0705bc0d55ce6e332f2985cc63287/detect_secrets_server/repos/base_tracked_repo.py#L152-L157).

It's not on my short-term roadmap, but now that we have verifiable secrets plugins, doing a one-time-only verifiable-secrets-only scan across all repos after adding a "from now on" CI step, seems to make a lot of sense.

perryzjc:
> Hi @FYJen, glad to hear you like it!
> 
> This tool specifically doesn't do git history scanning, since that's not what it's really designed for. It's designed to be fast, breadth-based scanning, rather than uncovering all the secrets that could have been in your git repository (depth-based).
> 
> If you're looking for a tool like that, I'd recommend https://github.com/dxa4481/truffleHog.

@domanchi, I've been pondering this topic and have some input. When using an alternative tool to scan commit histories, such tools as [truffleHog](https://github.com/dxa4481/truffleHog) don't have the exact same plugins developed for this project. This is especially relevant for those who have crafted custom plugins for detect-secrets. I feel it would be beneficial to introduce a specific command that allows users to scan the entire commit history, in addition to the default behavior. I'd love to hear your thoughts on this. Thanks!

Issue #212, opened 2019-07-27
detect-secrets-hook mutates baseline and fails


lirantal:
## Problem statement

When `detect-secrets-hook` runs and mutates the secrets baseline for timestamp and update of the line numbers it also terminates with exit code 1 which means an error, and looks like this:

```
The baseline file was updated.
Probably to keep line numbers of secrets up-to-date.
Please `git add .secrets-baseline`, thank you.
```

This means that if I use it in a pre-commit hooks setup it fails my commit process and requires me to manually re-add the secrets baseline and run it again. It is very unintuitive and requires manual work for developers.
 
## Proposal
Show the same error message about baseline being up to date and adding it to the repository but terminate with exit code 0 to make sure a pre-commit script doesn't fail. This is also a strong point IMO because the error isn't that secrets were detected, just that the baseline gets mutated which is not an exception but rather expected behavior of the tool.

### Alternative
If terminating with an error code is required at least use a different error code so a scripting on top of it can differentiate between the different errors and act accordingly.

domanchi:
Our development of this hook was under the assumption that the hook would be used together with a specific [pre-commit framework](pre-commit.com). IIRC, this framework does not display output unless the error code is non-zero.

I agree that this isn't the best option for it, but this is the design reason for that choice. If reducing manual effort is the problem, maybe we can auto-add this? The issue is that I'm not sure whether auto-adding is expected behavior with the framework.

Personally, I use the following script in my workflow:

```bash
function main() {
    # Run the pre-commit in the tox environment, to mitigate OS issues
    tox -e pre-commit -- run

    # Add changed files (assuming pre-commit modified files)
    if [[ $? == 1 ]]; then
        git diff --staged --name-only --diff-filter=ARM | xargs git add
    fi
}

main
```

As always, we're open to suggestions, but please keep this compatibility in mind with your changes.

lirantal:
Other frameworks like `lint-stated` has the same kind of behavior and will not show a message unless there's an error. My current workaround for this is to wrap the detect-secrets-hook with my own to parse the stderr output and then return a 0 exit code. However as I mentioned, I don't think this specific circumstance is of an error'ish nature. The hook updates the timestamp/line numbers of the file and then requests to re-stage it - that's IMHO not something you want to break the commit for.

How about my suggestion of at least changing the exit code number so that we can differentiate between errors (1) and say file mutations like in this topic and exit with say (2) ?

KevinHock:
[...]

PR #295, opened 2020-04-22
Add --from-commit argument to allow commit-incremental update scans

psedge:
**Problem**
In situations where use of detect-secrets-server isn't possible, it can take a long time to re-scan all files in a repository if it's sufficiently large, and it's rare that every file is worth re-checking.

**Proposal**
Rather than re-scanning all git-tracked files, store the commit of the last successful scan and if a flag is passed only compare the files changed since then. This allows incremental scans which are much more lightweight than a standard scan action. This is functionally similar to the way that detect-secrets-hook works.

- Add a --from-commit argument which accepts either:
    - a commit ref to diff against
    - no argument, taking the commit ref from the baseline
- Scan only the files that have changed
- Store the HEAD ref in a new baseline field `commit`


detect_secrets/core/baseline.py
```diff
@@ -80,6 +99,9 @@ def initialize(
     for file in sorted(files_to_scan):
         output.scan_file(file)
 
+        if from_commit:
+            output.files_scanned.append(file)
```
domanchi, line 103: This line had me thinking for a while. Why do we need to keep track of all the files that we scan in the baseline?

By definition, this would produce a different set of files as compared to `git ls-files`. However, you could also argue that scanning the differential wouldn't seek to create a baseline as we know it, rather, attempts to see whether there's any secrets between the git commit provided, and current HEAD.

If this is true (and there's no other need for having a list of files scanned), you could just use the `results` field to see if there are any secrets between the `from_commit` and HEAD. If there isn't, our job is done. If there is, it's not like we'll be creating a baseline from it anyway?

Granted, this is overloading the baseline format as another way to present detected secrets, but :shrug:

KevinHock:
I guess the main question here is, do we want to store the `last commit scanned` inside or outside the given repo? 


I can think of the following Pros and Cons. I'd like to hear both your thoughts. I tried to put the tl;dr of each point in the first sentence.

Pros
- This works "out of the box." Compared to needing to e.g. setup an S3 bucket or DB table to store the `last commit scanned` information.
- This is more speculative, but to add to the previous point: this would make it so that way more users use detect-secrets for CI. Whatever we are doing now for -server, doesn't seem easily adoptable for other companies. I think because of the previous reason. "Works out of the box" garners way more users, than "Does not work out of the box." More users means more community, more feedback, more PRs, etc. So it is not to be underestimated IMO.

Cons
- Merge conflicts. Sample scenario: PR A adds secret A, PR B adds secret B, merge conflict happens on `commit` field. I think [generated at](https://github.com/Yelp/detect-secrets/issues/92#issuecomment-505558043) is the only currently git conflicting field in the sample  scenario, and we have had multiple people ([e.g.](https://github.com/Yelp/detect-secrets/issues/212#issuecomment-516405023) [e.g.](https://github.com/Yelp/detect-secrets/issues/92#issuecomment-500781668) are some external examples) say that that isn't worth being in the baseline. Maybe digressing, but it is still unclear [what the benefits are of `generated_at` over `git log`](https://github.com/Yelp/detect-secrets/issues/92#issuecomment-505558043).
- It is not strictly necessary to be in the output of the tool. (If whatever-does-the-CI-step can do the tracking of `last commit scanned`.)
- Maybe: It is not useful for the pre-commit hook side of things, right? Baseline is dual-purpose so putting things in there only useful to one-purpose (the CI step) makes this less necessary than other fields. (We could have the pre-commit hook [only check the change of files since HEAD](https://github.com/Yelp/detect-secrets/issues/92#issue-377561647). This would have the usability benefit of, user-who-used-the-hook doesn't have the hook yell at them about changes made by user-who-did-not-use-the-hook. I remember a certain user repeatedly saying "you can call git diff in the pre-commit hook", but our response being the pre-commit hook framework only gives the files that changed, but it seems like we can call `git` ourselves.)

Neutral
- It might seem like a Pro that this PR avoids "secrets that the CI component complains about become out-of-sync with the baseline", but that's not true if `-server` can be used as a server-side pre-commit hook.
- Similar to the previous point, I think where we store `last commit scanned` does not affect the found-secret-feedback process. i.e. The "Hey -server, scan branch X and if a potential secret is found, give the committer of branch X perform some kind of feedback process, and possibly block the commit." process.
- This does not make the tool more amenable to scanning through git history (#294). Maybe at first glance, to those unfamiliar with the tool, it might sound like it. This is because this PR only cares about "secrets currently in HEAD for a given branch", we do not have to worry about representing secrets from a certain commit that are no longer in HEAD. Which is, like the issue feedback said, not what detect-secrets was designed to do.

There is is the `Pro` that this would catch secrets as soon as they hit any branch, and not just the `master` branch, so that is a big win. However, I think [-server](https://github.com/Yelp/detect-secrets-server) is amenable to have an option for scanning all branches -- or a given/current branch, if called in a CI step (i.e. server-side hook) context.

### In Summary

I'm not sure if the Pros outweigh the Cons, or vice versa, I do think that this is a _very_ permanent decision though. I also think we are not able to make an educated decision about how much merge conflicts affect frequently-committed-to-by-many-engineers repos, right now detect-secrets-hook may be borderline unusable for them because of the "yell at them about changes made by user-who-did-not-use-the-hook" example above.

detect_secrets/core/baseline.py
psedge: In this implementation it was necessary due to problems if we're passing the `--update` flag, as we need to know which files were 'scanned'; if a scan has found that secrets have been removed then the `new_results` won't have a key for eg. 'fileA', whereas `old_results` may do. In this case, we need to not include `old_results['fileA']` during our `merge_results` call.

I actually struggled with the changed files concept; if we proceed with this idea I'll revisit this to use `scan_diff` as it sounds like it solves the problem! My concern was that it's possible to change a file and not commit, scan, then revert the file. We'd now have secrets entries which have been invalidated.

killuazhu:
> - Add a --from-commit argument which accepts either:
>    - a commit ref to diff against
>    - no argument, taking the commit ref from the baseline

@psedge I'm also curious what the behavior would be if the commit ref from the baseline no longer exists? Would it scan all the commits? Or just give up on `--from-commit` argument? This can happen consistently if your team use a _squash and merge_ PR approval flow. The commit still existed and referenced by the hidden PR ref on Github server, but local clones might not know that commit at all.

detect_secrets/__init__.py
```diff
@@ -1 +1 @@
-VERSION = '0.13.1'
+VERSION = '0.13.2'
```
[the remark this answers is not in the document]
KevinHock: Remember to bump the version [mentioned in the README](https://github.com/Yelp/detect-secrets#pre-commit-hook) as well.

You can use `python scripts/bumpity.py --bump minor` to do this if you want to.

detect_secrets/core/baseline.py
```diff
@@ -299,6 +335,67 @@ def _get_git_tracked_files(rootdir='.'):
...
+                stderr=fnull,
+            )
+        for filename in git_files.decode('utf-8').split():
+            # As opposed to ls-files which gives path relative to -C, diff gives
+            # paths from git root directory
+            relative_path = util.get_relative_path_if_in_cwd('.', filename)
+            if relative_path:
+                output.append(relative_path)
+    except subprocess.CalledProcessError:
+        pass
```
KevinHock, line 372: Nit: Might be better practice to only wrap the relevant `subprocess` call in a `try:` `except:`

tests/core/secrets_collection_test.py
```diff
@@ -383,6 +413,12 @@ def test_load_baseline_without_exclude(self, mock_log):
             )
         assert mock_log.error_messages == 'Incorrectly formatted baseline!\n'
 
+    def get_point_thirteen_point_two_and_later_baseline_dict(self, gmtime):
+        # In v0.13.2 --from-commit got added
```
KevinHock, line 417: ++, great tests!

psedge:
> To summarize my thoughts on this, I think this is worth having, since it will do what -server does but out of the box not needing something like S3.
> 
> It may contribute to merge conflicts like `generated_at` currently does, but we could always not include the `commit` field if the flag for #92 was passed.

I've been thinking about this over the last few days and I think you are right that merge conflicts are fairly likely on this field in a many-engineers-one-codebase setup, and it's going to sometimes require manual intervention. They become far more likely if we update the `commit` field automatically in [baseline.py:85](https://github.com/Yelp/detect-secrets/pull/295/files#diff-20dc0d9456b2cd2689a9341ba637fa38R85), so would it be best to have this as a hand-updated field only? I'm thinking that this should be used more as a 'last safe point' which can be modified when the delta becomes prohibitively large/slow to scan.

Throughout our discussion I've actually moved towards using the -server component in my CI flow as it has numerous stability advantages over this implementation, but that hasn't solved the problems of the client component repeating work when scanning, where I think this might still be a valid feature. It is a permanent change to the .baseline format though as you say @KevinHock; and its naming may mislead users - if you can suggest better wording of either the flag or the usage instructions please do!

psedge:
> > * Add a --from-commit argument which accepts either:
> >   
> >   * a commit ref to diff against
> >   * no argument, taking the commit ref from the baseline
> 
> @psedge I'm also curious what the behavior would be if the commit ref from the baseline no longer exists? Would it scan all the commits? Or just give up on `--from-commit` argument? This can happen consistently if your team use a _squash and merge_ PR approval flow. The commit still existed and referenced by the hidden PR ref on Github server, but local clones might not know that commit at all.

I ran into this problem when testing; I initially wrote the command to exit in this case but didn't know whether that was a good UX. I'm open to suggestions though, in the current implementation I believe it fails silently and acts as if no files changed between the non-existent ref and HEAD.

tests/core/secrets_collection_test.py
psedge: thanks! I wanted to ask though, I added my additional assertion at [secrets_collection_test.py:390](https://github.com/Yelp/detect-secrets/pull/295/files/43120752a40f8602045adced247e8c4c0aaae21e#diff-1ec51425ce6e3b8ecdcb252ac8a6555eR390) trying to stay in style by duplicating the contents of the previous `test_load_baseline_from_string_with_X` methods; would it have been better to simply call those previous version tests in mine, and then test my assertions?

perryzjc:
@psedge  I am love to see this feature get implemented... but the PR has been closed for 4 years and no further discussions.

lorenzodb1:
@perryzjc consider opening a PR with the changes contained in this one so we can move from there

Issue #316, opened 2020-07-11
Updating baseline fails with version 0.14.0


shaikmanu797:
```console
$ detect-secrets scan --update .secrets.baseline                
Traceback (most recent call last):
  File "/opt/miniconda3/envs/tests/bin/detect-secrets", line 8, in <module>
    sys.exit(main())
  File "/opt/miniconda3/envs/tests/lib/python3.7/site-packages/detect_secrets/main.py", line 56, in main
    word_list_hash,
  File "/opt/miniconda3/envs/tests/lib/python3.7/site-packages/detect_secrets/main.py", line 141, in _perform_scan
    _get_plugins_from_baseline(old_baseline),
  File "/opt/miniconda3/envs/tests/lib/python3.7/site-packages/detect_secrets/main.py", line 101, in _get_plugins_from_baseline
    secrets_collection = SecretsCollection.load_baseline_from_dict(old_baseline)
  File "/opt/miniconda3/envs/tests/lib/python3.7/site-packages/detect_secrets/core/secrets_collection.py", line 131, in load_baseline_from_dict
    **plugin
  File "/opt/miniconda3/envs/tests/lib/python3.7/site-packages/detect_secrets/plugins/common/initialize.py", line 178, in from_plugin_classname
    klass = import_plugins(custom_plugin_paths)[plugin_classname]
TypeError: unhashable type: 'list'
```

KevinHock:
Thanks for making this! I’ll have a look 👀

schosterbarak:
also experiencing the same issue.

KevinHock:
I missed a `tuple()` wrap at 

https://github.com/Yelp/detect-secrets/blob/f9888771b040b581bcf03fcebcdae65ed7341f3e/detect_secrets/core/secrets_collection.py#L118-L119

I will make a PR to fix, but feel free to `pip install -e .`, and at a `tuple()` call in the meantime.

KevinHock:
Oh wow you maintain `checkov` @schosterbarak! A great tool.

schosterbarak:
Thanks @KevinHock ! Are you using it today? Would appreciate any feedback :) trying to learn how ppl are using it (CI/ pre-commit) and how to improve

KevinHock:
[...]

PR #317, opened 2020-07-11
:bug: Fix missing tuple() conversion

KevinHock:
Also correct mentions of v0.13.2, which was the wrongly anticipated version, to v0.14.0.

Note: We have had bugs in the past with --update, a few mentioned in the CHANGELOG and a couple not. We should have a test that would of caught this. I'll think about adding one but for now given the priority that can come in a follow-up PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,15 @@ If you love `detect-secrets`, please star our project on GitHub to show your sup
 ### Unreleased
 -->
 
+### Unreleased
+
+#### :bug: Bugfixes
+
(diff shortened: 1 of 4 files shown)


KevinHock:
Forgot to link to the issue: This fixes #316.

KevinHock:
Thanks @OiCMudkips! p.s. If you run `bumpity.py`, it will [update the README](https://github.com/Yelp/detect-secrets#pre-commit-hook).

OiCMudkips:
Whoops! I'll be sure to do that in the future.

PR #329, opened 2020-08-13
Improve wording of user prompt

heathd:
On the first use, we are prompted:

```
Secret:      1 of 1
Filename:    dev.yml
Secret Type: Secret Keyword
----------
1:memory: 500M
2:db_password: mysecretpassword
3:db_user: ledger
4:db_name: ledger
5:db_ssl_option: ssl=none
6:disk_quota: 500M
7:disable_internal_https: 'true'
----------
Is this a valid secret? i.e. not a false-positive (y)es, (n)o, (s)kip, (q)uit: 
```

The wording of this prompt contains a double-negative which is hard to comprehend for a new user. To avoid the double negative, I propose an alternative clearer wording:

```
Secret:      1 of 1
Filename:    dev.yml
Secret Type: Secret Keyword
----------
1:memory: 500M
2:db_password: mysecretpassword
3:db_user: ledger
4:db_name: ledger
5:db_ssl_option: ssl=none
6:disk_quota: 500M
7:disable_internal_https: 'true'
----------
Prevent this secret from being committed? (y)es, (n)o, (s)kip, (q)uit: 
```

### DIFF
--- a/detect_secrets/core/audit.py
+++ b/detect_secrets/core/audit.py
@@ -579,7 +579,7 @@ def _get_user_decision(prompt_secret_decision=True, can_step_back=False):
             print('Invalid input.')
 
         if 'y' in allowable_user_input:
-            user_input_string = 'Is this a valid secret? i.e. not a false-positive (y)es, (n)o, '
+            user_input_string = 'Prevent this secret from being committed? (y)es, (n)o, '
         else:
             user_input_string = 'What would you like to do? '
         if 'b' in allowable_user_input:


KevinHock:
Auditing can be done asynchronously from committing, you may audit a ton of baselines, and then punt off removing them from git, if all the secrets have been in the repo for years.

I generally agree double-negatives are bad, but I feel like most engineers understand 'false-positive' in this context more immediately than true-positive, that's just my opinion though. 🤔  I'd like some amount of consensus opinions from people that aren't you and me to chime in on this. (To discuss: `i.e. not a false-positive` -> `i.e. a true-positive`)

KevinHock:
I definitely appreciate the PR though, would just like to hear more opinions.

OiCMudkips:
My personal preference would be just to remove the false-positive and not change the wording to much; i.e. `Is this a valid secret? i.e. a true-positive secret`.

As @KevinHock, saying yes here doesn't actually prevent the secret from being committed, I didn't think.

idavidmcdonald:
Hi, I'm using detect-secrets as a new user and I just had to ask for help from people who have used it before because I wasn't clear what 'Is this a valid secret? i.e. not a false-positive' meant. I do understand what a false-positive is but the double negative did make it harder for me. I also faced confusion of what you mean by a 'valid' secret. Is a secret 'valid' if it is something that is really secret and you don't want others to see or does it mean it's valid if it is an 'allowed' secret because you've checked it and you are allowing it to to be included.

I like 'Prevent this secret from being committed?' as that is very clear to me what I should answer although unfortunately I  see this isn't relevant in all situations. 

Regardless, the current text I found hard to confuse and would encourage a change.

Thanks!

heathd:
I have to confess that I used the tool before reading the documentation, and so I hadn't understood that `detect-secrets` allows you to recognise existing secrets in your codebase and allow them to be committed whilst preventing new ones from being added. 

To be honest I am still not entirely clear on what the consequences are of answering this prompt "yes" vs "no" is, and having now re-read the README.md I am still not clear! I _think_ that what's going on is if I say "yes" then I'm saying that the thing that the scanner plugin found really is a secret and needs to be eventually expunged from this repository. If i say "no" then I'm saying that the scanner found something that looks like a secret but actually isn't a secret at all. Possibly what's going on is that the information I enter during this audit doesn't have any effect on the program's behaviour in the future, it's just something for the record so that human attention can be given to sorting out the secrets where I answered "yes" during the audit?

In the example I gave above, where there's a dev configuration used in ephemeral containers during testing, the `mysecretpassword` _is_ a password. So I'm left scratching my head asking whether I should answer "yes" because technically it is a password, or should I answer "no" because although it's used for authentication in certain circumstances, it doesn't need to be kept secret and therefore in some sense is not a "valid" secret. 

I think I'm just not very clear on what is going on during the audit. Help!?

domanchi:
Thanks for your feedback @heathd / @idavidmcdonald! I agree that the README is pretty outdated, and this feature should definitely be explained in much more detail. The `audit` feature was actually birthed from my scripted attempt at analyzing the large set of baselines we had to process when initially rolling out this tool -- and I'm afraid it was just left at that state for a long time.

Let me see what I can do to make this clearer. Mind if I add you as a reviewer for the future PR (porting this feature to our `pre-v1-launch` branch)?

idavidmcdonald:
@domanchi Very happy to help. Thanks in advance 👍

domanchi:
@idavidmcdonald , check out https://github.com/Yelp/detect-secrets/pull/363 and see whether that clears things up for you.

domanchi:
Closing as #363 fixed this in the upcoming v1 branch.

PR #330, opened 2020-08-13
Added support for JUnit XML format

AgrawalAmey:
JUnit XML format has become standard format for storage of test case results. When running detect-secrets on CI platforms like Jenkins and Bitbucket pipelines it becomes easier to read results if they are available in JUnit XML format. This PR adds a --junit-xml arg which can be used to print the result in the said format.

```
<?xml version="1.0" ?>
<testsuites disabled="0" errors="0" failures="150" tests="150" time="0.0">
	<testsuite disabled="0" errors="0" failures="53" name="Secret Keyword" skipped="0" tests="53" time="0">
		<testcase name=".secrets.baseline:26">
			<failure message="Found secret of type Secret Keyword on line 26 in file .secrets.baseline" type="Secret Keyword"/>
		</testcase>
		<testcase name=".secrets.baseline:33">
			<failure message="Found secret of type Secret Keyword on line 33 in file .secrets.baseline" type="Secret Keyword"/>
		</testcase>
```

### DIFF
--- a/detect_secrets/core/usage.py
+++ b/detect_secrets/core/usage.py
@@ -269,6 +269,13 @@ class ScanOptions:
             help='Scan all files recursively (as compared to only scanning git tracked files).',
         )
 
+        self.parser.add_argument(
+            '--junit-xml',
+            action='store_true',
+            dest='junit_xml',
+            help='Returns output in JUnit XML format.',
+        )
+
         return self
 
     def _add_adhoc_scanning_argument(self):
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ setup(
     install_requires=[
         'pyyaml',
         'requests',
+        'junit-xml'
     ],
     extras_require={
         'word_list': [
--- a/detect_secrets/core/baseline.py
+++ b/detect_secrets/core/baseline.py
@@ -3,6 +3,8 @@ import os
 import re
 import subprocess
 
+import junit_xml
+
 from detect_secrets import util
 from detect_secrets.core.log import get_logger
 from detect_secrets.core.secrets_collection import SecretsCollection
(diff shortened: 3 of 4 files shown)


detect_secrets/core/usage.py
```diff
@@ -256,11 +256,11 @@ def _add_initialize_baseline_argument(self):
         # Pairing `--update` with `--scan` because it's only used for
         # initialization.
         self.parser.add_argument(
-            '--update',
+            '--old-baseline',
             nargs=1,
             metavar='OLD_BASELINE_FILE',
-            help='Update existing baseline by importing settings from it.',
-            dest='import_filename',
+            help='Exitinting secrets from existing baseline.',
```
KevinHock, line 262: Nit: Spelling

setup.py
```diff
@@ -21,6 +21,7 @@
     install_requires=[
         'pyyaml',
         'requests',
+        'junit-xml'
```
KevinHock, line 24: I think most users will not need this feature, similar to the `--word-list` option I added, can you put this under `extras_require` due to this?

AgrawalAmey:
@KevinHock So my end to end use case is something like this:

1. We initialise the baseline inside git repo.
2. When the user commits something, it triggers a bitbucket pipeline. In this pipeline, we want to do an incremental scan based on the existing baseline file to check if new secrets have been added.
3. If there are new secrets, we will fail the pipeline and prevent any further steps which might reveal the secret in public domain. And we would show the list of new secrets in by passing the JUnit XML to bitbucket pipeline.

From what I understand, previously the incremental scan support was only present in the pre commit hook. So in the last two commits I have added a --old-baseline option to such incremental scans. At the moment, I have just overriden the existing --update option. If the concept looks fine, will update the PR to support both the options.

domanchi:
@AgrawalAmey, this is certainly an interesting idea. Could you elaborate what you mean by:

> the incremental scan support was only present in the pre commit hook

?

Issue #310, opened 2020-06-05
Future Enhancment: filter potential secrets from baseline


dkmbw:
Hey,
as i said, we have large and many repositories and we start cleaning them up. but we have also some blacklists of words which are NOT a secret. and we got an consent on example secrets which can be used for documentation issues, which shall not be regarded as secrets.

But of course, sometime we have to check that nobody really uses them as actual secrets in the git repository.

Therefore i would like to propose a `detect-secrets filter baseline.json` command which filters following stuff into a new baselinefile.json (via stdout as with scan), without

- certain files
- certain secret
- choosable combination of the above (must be both together, must be at least one of them..)

which we give the filter command via file, perhaps `--omit-files=...` or `--omit-potential-secrets=...`

That makes it easier to go over big repositories and test in them in a semi automated way.

sharadpattanshetti:
This is really good feature to have it.

OiCMudkips:
To exclude files, you can use either the `pre-commit` framework's `exclude` option, or if you're using this tool directly, the [`--exclude-files` flag](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/usage.py#L247-L254). This has to be done at scan time.

We also actually support the ignore-words option if you install this plugin using the [`word_list` extra](https://github.com/Yelp/detect-secrets/blob/master/setup.py#L26) and the [`--word-list` flag](https://github.com/Yelp/detect-secrets/blob/7776f4be3f1de476296d6fa754c730a41613ee6a/detect_secrets/core/usage.py#L19-L27).

Looking at our README we don't actually document these. We should definitely surface these features somewhere.

dryoni:
I actually created an advanced audit tool decoupled from detect-secrets which reads baselines files and allows real time exclusion of line patterns and file patterns.
In addition, it aggregates all found secrets with same values so we can mark all of them as secrets or not secrets.
We've managed to audit ~9000 secrets in about 15 minutes.
This would be impossible to accomplish with the regular detect-secrets audit feature.
I actually pushed a new PR to support the advanced auditing tool (which requires having the secret length and position of the secret in line for each secret) - https://github.com/Yelp/detect-secrets/pull/332
Hope it will be approved, and I'll publish my auditing tool, and help integrate it into detect-secrets

![image](https://user-images.githubusercontent.com/20098914/90309875-00ccd880-def5-11ea-98c2-2c355bcdab16.png)

domanchi:
[...]

PR #332, opened 2020-08-14
Faster raw secret fetch during audit

yonidrayerman:
During audit, to get the raw secret value (for highlighting the secret) each file is re-scanned using every plugin, the hashes of the resulted secrets are then compared to the 'hashed_secret' value of the secret, and if there's a match the raw secret value is returned.
Although this process should allow detecting secrets which are not in the correct line - this is not currently not supported while auditing and for good reason. If the code was changed you'll want to create a new baseline.

To speed up this process, I've added the secret length and position in line into each PotentialSecret object.
During audit, it takes the substring from the file according to the line number, secret length, and line position, hashes it and compares to the hashed value of the secret.
This makes the get_raw_secret_value function go super fast and removes dependancy on plugins.


OiCMudkips:
Hey yoni, I keep on meaning to review this but I remember there might be some context why we didn't want to do this, so I'm looking for a block of time where I have a chance to figure this out. If domanchi or KevinHock remembers, they can take a stab at this too :)

dryoni:
I'm guessing that the reason for not doing lies with the comment on the Contributing page "The line number does not play a part in the identification of a potential secret because code is expected to move around through continuous iteration. However, through the audit tool, these line numbers are leveraged to quickly identify the secret that was identified by a given plugin."

The issue is that the audit tool doesn't only leverage the line numbers to quickly identify the unencrypted secret, but it completely relies on it. If the line number is mismatched, the audit tool will not present the line with the secret and adjacent lines. In this case, the line number (along with the secret length and position in line) can be completely relied on to get the raw secret, while not playing any role in the identification of a potential secret.

On the long term, this feature will allow mass retrieval of raw secrets from big baselines a matter of seconds instead of minutes or even hours.
I can share a benchmark test I ran with the existing function and the modified function that shows the huge run time difference between the functions.

KevinHock:
I thought we removed the `audit` functionality relying on line numbers in https://github.com/Yelp/detect-secrets/pull/208, mainly to fix an issue we were running into. https://github.com/Yelp/detect-secrets/pull/208#issuecomment-512606596

If that is true, then
> If the line number is mismatched, the audit tool will not present the line with the secret and adjacent lines.
doesn't make sense. 


Side question: how are you hitting Y/N when auditing so fast? I'd think that that would be the bottleneck while auditing, more so than pulling from the line number vs. file as a whole.
I was thinking that using string similarity, in combination with `audit --display` to show that raw secrets, one could probably automatically determine false-positives with a big enough large data set, in that true-positives should be fairly unique, more than false-positives.

KevinHock:
Can confirm we aren't relying on line numbers in `audit`, https://github.com/Yelp/detect-secrets/blob/c09b466f7600e35e093c5da0437eca74a7c7a459/detect_secrets/core/audit.py#L744-L751, we forgot to update `CONTRIBUTING.md` with that info, and I think the `SecretNotFoundOnSpecifiedLineError` exception should be renamed to be `SecretNotFoundInFileError`.

detect_secrets/core/audit.py
```diff
@@ -707,45 +701,29 @@ def _get_secret_with_context(
...
+    line_number = secret['line_number']
+
+    file_lines = file_handle.readlines()
+    if len(file_lines) >= line_number:
+        line = file_lines[line_number - 1]
+        possible_secret = line[line_position:line_position + secret_length]
 
-    plugin_secrets = plugin.analyze(file_handle, filename)
+        possible_secret_hash = PotentialSecret.hash_secret(possible_secret)
+        lowered_possible_secret_hash = PotentialSecret.hash_secret(possible_secret.lower())
```
KevinHock, line 724: Can you share a test case where `.lower()` vs. not makes a difference?

dryoni:
@KevinHock  - you are correct, the current get_raw_secret_value function does return the raw secret regardless of the line number.
However, when trying to highlight the line during the audit (https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/audit.py#L693) if the line number is mismatched you get the same exception as if the secret was never found by get_raw_secret_value - SecretNotFoundOnSpecifiedLineError(secret['line_number']) - Essentially relying on the line number for raw secret retrieval during audit.

detect_secrets/core/audit.py
dryoni: I found two examples in which the hashed secret was the hash of the lowercased secret:
Secret Line: "Password": "SomeKindOfPassword123!@#",
The hashed secret is d1b7e1c27fc85c2bb6e5f2923451bbc76b156e72 (SHA1 of 'your_secret') instead of e383adc92711d348f6df827db2452eddcbd32384

Line: secret = YOUR_SECRET
The hashed secret is 94bc66069e1b40e1294da9bc98187564db71eefd (SHA1 of 'somekindofpassword123!@#') instead of 9c423450db31f2428d1f9eea2c152a4f76b21dbd

dryoni: As I'm seeing this mostly with the Secret Keyword plugin, I'm guessing it has to do with this line: https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/keyword.py#L340

KevinHock: Great catch!

dryoni:
@KevinHock Regarding your side question, as I mentioned in https://github.com/Yelp/detect-secrets/issues/310 - I've built an advanced audit tool decopuled from detect-secrets that reads baselines files and allows real time exclusion of line patterns and file patterns. To accomplish this, the raw secrets must be fetched pre-audit to be able to combine similar secrets - so this process has to be very fast.

As a comparison, when using the --display-results option, all the raw secrets are fetched (both false-positives and true-positives). I ran it with the current version on a baseline with 600 secrets - it took 83 seconds to print out.
With the new version it took 3.6 seconds.

domanchi:
> I ran it with the current version on a baseline with 600 secrets - it took 83 seconds to print out. With the new version it took 3.6 seconds.

That is **super** cool. I really wonder what is the cause of such a discrepancy -- let me play around with this in the new `pre-v1-launch` branch, and report back.

gpflaum:
@dryoni Are you still using detect-secrets? If your [audit tool](https://github.com/Yelp/detect-secrets/issues/310#issuecomment-674376195) works with the latest detect-secrets, are you able to publish it?

Issue #326, opened 2020-08-12
Repeated secrets in a file are not being captured


shaikmanu797:
@KevinHock According, to the comments [here](https://github.com/Yelp/detect-secrets/blob/masterdetect_secrets/core/potential_secret.py#L54-L56) if any secret is found more than once that occurrence is ignored. A re-scan had to be performed after fixing the first occurrence to detect the leftover secrets and so on.

Instead, I would suggest adding line number as prefix salt before hashing to make it unique.

https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/potential_secret.py#L59-L60

```python
def set_secret(self, secret, salt=None):
   if salt is None:
      salt = self.lineno
   self.secret_hash = self.hash_secret(f"{salt}{secret}")
  ...
```

shaikmanu797:
@KevinHock any suggestion for above.

KevinHock:
>A re-scan had to be performed after fixing the first occurrence to detect the leftover secrets and so on.

Sorry to hear that 😞 

It is intentional that line numbers are not counted, as in most use-cases it would be annoying for a user if the lines changed around. At least that is the case for pre-commit hooks, for e.g. arc, it only yells at the changed lines.

The way one would do this, is to add the line number to `fields_to_compare`, it makes it a lot more straightforward than the salt.


I'd be open to you adding this under a flag, so that it wouldn't be the default behaviour. Does that sound amenable to you @shaikmanu797?

Thanks so much for your interest in this project btw :D

shaikmanu797:
@KevinHock, will come up with a PR to allow the `fields_to_compare` input from cli.

lorenzodb1:
We're going to close this issue as it hasn't received any update in a very long time. Feel free to re-open it if you think it's still relevant.

PR #335, opened 2020-08-27
Added extra field to compare argument to compare generated potential secrets

shaikmanu797:
#326 

@KevinHock, I have added a new argument `--extra-field-to-compare` tuple with only allowed choices and passed along the workflow. Included unit tests as well.

### DIFF
--- a/detect_secrets/core/baseline.py
+++ b/detect_secrets/core/baseline.py
@@ -20,6 +20,7 @@ def initialize(
     word_list_file=None,
     word_list_hash=None,
     should_scan_all_files=False,
+    extra_fields_to_compare=None,
 ):
     """Scans the entire codebase for secrets, and returns a
     SecretsCollection object.
@@ -43,6 +44,9 @@ def initialize(
 
     :type should_scan_all_files: bool
 
+    :type extra_fields_to_compare: Tuple[str]|None
+    :param extra_fields_to_compare: Extra fields to be used during secrets comparison.
+
     :rtype: SecretsCollection
     """
     output = SecretsCollection(
@@ -52,6 +56,7 @@ def initialize(
         exclude_lines=exclude_lines_regex,
         word_list_file=word_list_file,
         word_list_hash=word_list_hash,
+        extra_fields_to_compare=extra_fields_to_compare,
     )
 
     files_to_scan = []
--- a/detect_secrets/core/potential_secret.py
+++ b/detect_secrets/core/potential_secret.py
@@ -21,6 +21,7 @@ class PotentialSecret:
         secret,
         lineno=0,
         is_secret=None,
+        extra_fields_to_compare=None,
     ):
         """
         :type typ: str
@@ -41,6 +42,9 @@ class PotentialSecret:
         :type is_secret: bool|None
         :param is_secret: whether or not the secret is a true- or false- positive
 
+        :type extra_fields_to_compare: Tuple[str]|None
+        :param extra_fields_to_compare: Extra fields to be used during secrets comparison.
+
         :type is_verified: bool
         :param is_verified: whether the secret has been externally verified
         """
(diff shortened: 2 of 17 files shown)


KevinHock:
Hi @dryoni / @yonidrayerman, would this be something you'd want as well, re: your auditing PR?

I'm wondering if we want a boolean `--include-lineno-in-secret-hash` or this PR, and you may be interested in it. If we want more than just `--line-number` to matter, it would be good to say now.

It seems like there is a tradeoff, between  
'state of the world' , auditing all existing secrets quickly and efficiently (multiple potential secrets with the same hash in the same file [this PR], thousands of secrets in huge huge file [#332], etc.)
vs.
'from now on prevention', not being a pain to developers using the pre-commit hook (line number affecting the secret hash, would do this, even if we analyzed only the lines changed like arc does, or like we could do if we really wanted to e.g. call `git diff`).

Both your PRs address pain points in the former.

dryoni:
Hi @KevinHock, I think that only the line number make any difference. 
Duplicate secrets one the same are probably very rare, but can be detected in the future by the CodeSnippet.highlight_line function and have all occurrences highlighted.

I agree there's a tradeoff here, but it's mostly due to the the current one-secret-at-a-time audit process.
One of the steps to improve the auditing experience and allow new auditing options is having faster raw secret retrieval which is the essence of my PR: https://github.com/Yelp/detect-secrets/pull/332
It doesn't add any complexity to the developers or the auditors.

https://github.com/Yelp/detect-secrets/pull/335 adds more pain to developers but provides better security, and I highly recommend adding it.
Auditing is ultimately the responsibility of the security team, so improving the security on the cost of adding a few more duplicate secrets to audit, sounds like a good tradeoff.

It might be worth adding a special Inline Allowlisting comment to ignore all duplicate secrets on the same file.
Adding a comment on every duplicate line, especially if it's all false-positives can be very annoying, so developers might just skip pre-commit tests.

shaikmanu797:
@KevinHock are you awaiting any additional changes on this PR before you review?

shaikmanu797:
@OiCMudkips, could you take a look at this PR?

KevinHock:
Have we determined feasibility of counting the secrets in the files? :thonk: 🤔 

That way, moving lines around doesn't affect developers -- yet we will have the alerting for multiple secrets in the same file. The best of both worlds.

domanchi:
Hi @shaikmanu797,

Thanks for your contribution to this project, and I apologize for it taking so long to merge. I will be taking the next couple of days to personally review the backlog that has been stacking up over the past months, and hope to see many of these great community additions being added to this tool.

Reading this PR makes my heart pang -- all you wanted to do was change something in `PotentialSecret`, and you had to touch >200 lines to do so. We are currently undergoing a large re-architecture of this tool to make development easier (`pre-v1-launch` branch), and hopefully that will make issues like this much easier to modify in the future.

Within that branch, I will have a PR out to optionally remove line numbers from baselines, and ultimately seal the *long* discussion on whether or not line numbers should be part of it. I realized that since line numbers are used for auditing alone, as long as we disable the auditing feature if the baseline was generated without line numbers, we should be able to leave teams with the decision of whether or not to keep the line numbers around when generating baselines.

Do you think that would address this issue?

shaikmanu797:
> Hi @shaikmanu797,
> 
> Thanks for your contribution to this project, and I apologize for it taking so long to merge. I will be taking the next couple of days to personally review the backlog that has been stacking up over the past months, and hope to see many of these great community additions being added to this tool.
> 
> Reading this PR makes my heart pang -- all you wanted to do was change something in `PotentialSecret`, and you had to touch >200 lines to do so. We are currently undergoing a large re-architecture of this tool to make development easier (`pre-v1-launch` branch), and hopefully that will make issues like this much easier to modify in the future.
> 
> Within that branch, I will have a PR out to optionally remove line numbers from baselines, and ultimately seal the _long_ discussion on whether or not line numbers should be part of it. I realized that since line numbers are used for auditing alone, as long as we disable the auditing feature if the baseline was generated without line numbers, we should be able to leave teams with the decision of whether or not to keep the line numbers around when generating baselines.
> 
> Do you think that would address this issue?

@domanchi, I will take a look at the branch you mentioned, this is a very cool project and I'm interested to learn more about the re-architecture soon.

I agree with your point where a small change leads to changes in multiple files right now thereby causing confusion at times to ensure everything works as expected.

My requirement behind PR is to avoid multiple re-scans if the same secret is encountered multiple times by adding line_number to compare fields at secret filtering.

domanchi:
> My requirement behind PR is to avoid multiple re-scans if the same secret is encountered multiple times by adding line_number to compare fields at secret filtering.

To be clear, are you suggesting an optimization that doesn't require `O(num_plugins)` to scan the line, when we have already identified that it is indeed a secret (on the same line)?

PR #336, opened 2020-08-27
Fix adhoc scan verify

OiCMudkips:
The `self.verify` call was missing a parameter in the adhoc string scanning function. This PR adds the missing parameter.

Also, I noticed that a parameter name for `verify` was misnamed. I fixed that in this PR too.

### DIFF
--- a/detect_secrets/plugins/base.py
+++ b/detect_secrets/plugins/base.py
@@ -144,7 +144,7 @@ class BasePlugin:
                     lines_of_context=LINES_OF_CONTEXT,
                 )
 
-                is_verified = self.verify(result.secret_value, content=str(snippet))
+                is_verified = self.verify(result.secret_value, context=str(snippet))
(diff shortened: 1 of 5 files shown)


detect_secrets/plugins/base.py
```diff
@@ -239,7 +239,7 @@ def adhoc_scan(self, string):
 
         return output[verified_result]
 
-    def verify(self, token, content=''):
```
domanchi, line 242: It's actually meant to be `context`.

A couple of lines down, it reads:

```
:param context: lines of context around identified secret
```

OiCMudkips: Right, that's why I renamed it to `context`.

domanchi: 🤦 I must be getting cross-eyed.

PR #340, opened 2020-09-07
Change exclude *test* example

0atman:
For our usage, we've found the biggest potential leaker of secrets is using un-redacted customer data in tests.

The example config should **not** have `tests/` excluded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ $ cat .pre-commit-config.yaml
     hooks:
     -   id: detect-secrets
         args: ['--baseline', '.secrets.baseline']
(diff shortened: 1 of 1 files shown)


domanchi (approved):
Super interesting! Also, very cool that you use this tool to identify *customer data*, rather than just secrets?

LGTM.

0atman:
@domanchi I have already said too much :laughing:

0atman:
To be clear, it's a hypothetical usage we thought of with data scientists - who might be querying a lot of data that may or may not be effectively redacted. Better to be safe than sorry!

PR #342, opened 2020-09-13
Fix binary secrets parsing issue in yaml files

dryoni:
I've found that binary YAML secrets (Base64 High Entropy String) can't be audited, including the test_data/config.yaml and test_data/config2.yaml files in the repo.

Example:
test_data/config.yaml
Line number: 14
Type: Base64 High Entropy String
Secret: MmIwMDA0MmY3NDgxYzdiMDU2YzRiNDEwZDI4ZjMzY2Y==
Hashed secret is: 906d27fd62478c802abc0dbf4211c3efae3054f0 (which is the sha1 of this string: '|\n  MjNjcnh1IDJieXJpdXYyeXJpaTJidnl1MnI4OXkyb3UwMg==\n' )

The hashed secret should be 10ab7ab9856ae2aa93fcfeba71e99511dd55f70c

The reason for this is that the yaml.dump function adds a few more characters which are not omitted, which results in a wrong secret hash.

```
>>> yaml.dump(
...     base64.b64decode('MmIwMDA0MmY3NDgxYzdiMDU2YzRiNDEwZDI4ZjMzY2Y==')
... ).replace(
...     '!!binary ',
...     '',
... )
'|\n  MmIwMDA0MmY3NDgxYzdiMDU2YzRiNDEwZDI4ZjMzY2Y==\n'
```

### DIFF
--- a/detect_secrets/plugins/high_entropy_strings.py
+++ b/detect_secrets/plugins/high_entropy_strings.py
@@ -307,9 +307,9 @@ class HighEntropyStringsPlugin(BasePlugin):
             secret_in_yaml_format = yaml.dump(
                 self.encode_to_binary(potential_secret.secret_value),
             ).replace(
-                '!!binary ',
+                '!!binary |\n  ',
                 '',
-            )
+            ).rstrip()
 
             potential_secret.set_secret(secret_in_yaml_format)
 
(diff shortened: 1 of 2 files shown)


KevinHock (approved):
lgtm

thanks @dryoni

Issue #324, opened 2020-08-07
Make quotes required in keyword detector golang regexes


KevinHock:
Two liner to fix, changing  
https://github.com/Yelp/detect-secrets/blob/cd8b3999e141ed4faec75e5db087df0db2339f58/detect_secrets/plugins/keyword.py#L229-L230
to be
    FOLLOWED_BY_COLON_QUOTES_REQUIRED_REGEX: 5,
    FOLLOWED_BY_EQUAL_SIGNS_QUOTES_REQUIRED_REGEX: 4,

I think

From https://blog.golang.org/strings I think this is fine

KevinHock:
cc @baboateng I _think_ will be fine and not result in false-negatives, but wanted to give you a heads-up.

KevinHock:
Note-to-self: Add `api` to the fp's list, too.

PR #344, opened 2020-09-16
Modify regex patterns for keyword detector to identify patterns with keyword and appended string

digjanaik:
1. This change is mainly to identify keywords with variable names like -
PASSWORD**_FOO**
API_KEY**_XYZ**

Keyword detector doesn't consider variable names with above pattern as a secret, but detects variable names with pattern as a secret- 
XYZ_PASSWORD
FOO_API_KEY 

Variable names can have strings appended before/after the keyword.

2. Removed password, secret from the false positives list in keyword.py as these can be the passwords and it isn't a good idea to filter them out.

### DIFF
--- a/detect_secrets/plugins/keyword.py
+++ b/detect_secrets/plugins/keyword.py
@@ -107,14 +107,16 @@ FALSE_POSITIVES = {
     'null;',
     'pass',
     'pass)',
-    'password',
+    # Secrets with this value not detected (Could be and actual password)
+    #'password',
     'password)',
     'password))',
     'password,',
     'password},',
     'prompt',
     'redacted',
-    'secret',
+    # Secrets with this value not detected (Could be and actual password)
+    #'secret',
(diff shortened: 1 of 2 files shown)


OiCMudkips:
I'm gunna ping @domanchi or @KevinHock to verify but I'm pretty sure we've found internally that `secret` and `password` were more noise than signal for this plugin, since a lot of the time, if these terms are in the value, they're placeholders for real secrets.

KevinHock:
Can confirm @OiCMudkips, there is some reference, hopefully in an open GitHub issue, about making the keyword plugin more 'bring-your-own signals', so people can override the false-positives. 

But a plugin like this can always be more customized to suite a particular companies/targets coding patterns. I tried to make what it does as generalizable as possible. https://github.com/Yelp/detect-secrets/issues/324 is the main open issue I have about improving this plugin.

Issue #343, opened 2020-09-16
Fix compatibility issue with detect-secrets-server after introduction of "---custom-plugins"


francisluz:
#### 🗒️  Description

As `detect-secrets-server` doesn't support the flag `--custom-plugins` it's casing an error when you try to run the server with the latest client version.

#### 🐛 Bug Report

This bug has been reported at `detect-secrets-server` [62](https://github.com/Yelp/detect-secrets-server/issues/62)

#### 💡 Solution

Some code check need to be done when you're using the argument `custom_plugin_paths` and when you try to delete an attribute that is not present in collection.

lorenzodb1:
We're going to close this issue as it hasn't received any update in a very long time. Feel free to re-open it if you think it's still relevant.

PR #345, opened 2020-09-16
Fix issue with detect-secrets-server and -custom-plugins #343

francisluz:
#### 🐛 Bug Report
This bug has been reported at [#343]

#### 💡 Solution

Implemented double check code when using `custom_plugin_paths` and verify if the attribute is on the collection before delete it.


KevinHock:
I think this may be better with a detect-secrets-server PR, and not a detect-secrets PR, thoughts?

PR #356, opened 2020-11-09
Adding secret verification as builtin filter

domanchi:
## Summary

This adds secret verification as a new filter in the new proposed architecture. With the exception of the type hints and variable renames in `detect_secrets.plugins.*`, this is (hopefully) a pretty clear cut PR.

Some improvements over the old style:

1. Introduction of `--only-verified` flag, so the baseline can be created with only secrets that were able to be verified.
2. Verification of secrets can now be done through the `scan --string` functionality too.
3. The baseline verification of secrets' preference is now recorded and stored with the baseline.


OiCMudkips (approved):
lg2m

PR #357, opened 2020-11-10
Re-adding custom plugins

domanchi:
## Summary

This adds custom plugin support to the new proposed architecture. It's pretty elegant how it works (if I say so myself): the `detect_secrets.core.plugins.util.get_mapping_from_secret_type_to_class` is responsible for initializing the catalog for all available plugins, so we process the configured settings and import files as necessary in this function as well.

For plugins that are passed in via command line, we will attempt to initialize the plugins before adding them to this catalog, so that they we can raise errors early if it is an invalid file. Furthermore, since this is processed *after* the baseline post processing step (for the CLI), it will support adding plugins to an already configured baseline. (i.e. `detect-secrets scan  --baseline .secrets.baseline -p testing/plugins.py`).

Lastly, this should play well with disabled plugins, given that the decision of whether to use a plugin is found in `detect_secrets.core.scan.get_plugins`, which is executed well after the initialization of the catalog. This means that you should be able to disable custom plugins, even if they've been initialized in the baseline.

Some modifications over the old style:

- `--custom-plugins` has been changed to `--plugin`, which accepts multiple arguments. This way, you can call it as such:

  ```
  detect-secrets scan \
    -p testing/plugins.py \
    -p blah \
    test_data
  ```

  instead of having it all in one line.

  NOTE: A future PR will modify the `--disabled-plugins` flag to make it follow this style instead.

- Instead of having a (snowflake) `custom_plugins_path` entry in the baseline, it is represented in a more cohesive manner through the `plugins_used` entry, as such:

  ```
  {
      "plugins_used": [
          {
              "name": "AWSKeyDetector"
          },
          {
              "name": "HippoDetector",
              "path": "file:///Users/aaronloo/Documents/github/detect-secrets/testing/plugins.py"
          }
      ]
  }
  ```

  in doing so, we open the option for well-specified plugins (within the same file), as well as possible parameter injection as necessary (even though that would require a manual modification of the baseline, as its written right now).


OiCMudkips (approved):
Looks good to me!

PR #363, opened 2020-11-13
clearer help text for audit and scan functionality

domanchi:
## Summary

Adds better help text to `detect-secrets` arguments.

**Help Text**:

```
$ detect-secrets -h
usage: detect-secrets [-h] [-v] [--version] {scan,audit} ...

positional arguments:
  {scan,audit}
    scan         Creates a baseline by scanning a repository for secrets.
    audit        Manually assesses a baseline to determine validity of secrets
                 found.

optional arguments:
  -h, --help     show this help message and exit
  -v, --verbose  Verbose mode.
  --version      Display version information.
$
$ detect-secrets scan --help
usage: detect-secrets scan [-h] [--string [STRING]] [--all-files]
                           [--baseline FILENAME] [--force-use-all-plugins]
                           [--base64-limit [BASE64_LIMIT]] [--hex-limit [HEX_LIMIT]]
                           [--disabled-plugins DISABLED_PLUGINS] [-n]
                           [--exclude-lines EXCLUDE_LINES]
                           [--exclude-files EXCLUDE_FILES]
                           [--word-list WORD_LIST_FILE]
                           [path [path ...]]

Scans a repository for secrets in code. The generated output is compatible
with `detect-secrets-hook --baseline`.

positional arguments:
  path                  Scans the entire codebase and outputs a snapshot of
                        currently identified secrets.
...
$
$ detect-secrets audit --help
usage: detect-secrets audit [-h] [--diff] [--stats] [--json] filename [filename ...]

Auditing a baseline allows analysts to optimize plugins for the highest
signal-to-noise ratio, for their environment.

positional arguments:
  filename    Audit a given baseline file to distinguish the difference
              between false and true positives.
...
```

**Auditing:**

```
$ detect-secrets scan test_data > test.baseline
$ detect-secrets audit test.baseline
Secret:      1 of 80
Filename:    test_data/baseline.file
Secret Type: Secret Keyword
----------
59:    }
60:  ],
61:  "results": {
62:    "config.env": [
63:      {
64:        "hashed_secret": "513e0a36963ae1e8431c041b744679ee578b7c44",
65:        "is_verified": false,
66:        "line_number": 1,
67:        "type": "Base64 High Entropy String"
68:      }
69:    ],
----------
Is this a secret that should be committed to this repository? (y)es, (n)o, (s)kip, (q)uit: 
```

I hope this text will be self-explanatory, and give users the option to decide what to do for ambiguous cases (e.g. dev secret). The delicate balance as to whether or not a user should label dev secrets as **real** secrets will be expanded upon in the detailed docs (not in this PR).


detect_secrets/core/usage/audit.py
```diff
@@ -2,7 +2,14 @@
 
 
 def add_audit_action(parent: argparse._SubParsersAction) -> argparse.ArgumentParser:
-    parser = parent.add_parser('audit')
+    parser = parent.add_parser(
+        'audit',
+        help='Manually assesses a baseline to determine validity of secrets found.',
+        description=(
+            'Auditing a baseline allows analysts to optimize plugins for the highest '
```
KevinHock, line 9: Nit: Maybe `Additionally, auditing..`, I more mean that, optimizing plugins is a little secondary compared to migrating out true secrets.

detect_secrets/core/usage/scan.py
```diff
@@ -7,7 +7,14 @@
 
 
 def add_scan_action(parent: argparse._SubParsersAction) -> argparse.ArgumentParser:
-    parser = parent.add_parser('scan')
+    parser = parent.add_parser(
+        'scan',
+        help='Creates a baseline by scanning a repository for secrets.',
+        description=(
+            'Scans a repository for secrets in code. The generated output is compatible with '
+            '`detect-secrets-hook --baseline`.'
```
KevinHock, line 15: An aside: We have had a lot of users, I can pull up references, that want `detect-secrets` itself to take a baseline as well. Lots of people don't use pre-commit hooks, they might use arc linters and other things along those lines, and want to use `--baseline` argument without all the same assumptions that went into the pre-commit hook.

Curious what your thoughts are on this.

KevinHock (approved):
lgtm ;)

detect_secrets/core/usage/scan.py
domanchi: I think you may need to elaborate on this, but v1 does intend to support this. Essentially, the breakdown of the different usages of this tool include:

- `detect-secrets scan`: scans secrets, and adds new secrets to a baseline
- `detect-secrets-hook`: blocks new secrets
- `detect-secrets audit`: analyzes results

In v1, scan does take in a baseline (e.g. `detect-secrets scan --baseline .secrets.baseline`), but since scan's job is to create a baseline, it uses this baseline as a configuration file, and updates it with the latest secrets. The only assumptions in the pre-commit hook (as written) is:

- new secrets should be blocked
- people should be able to see which new secrets were blocked
- baselines should be regularly updated (which can be reduced with a WIP PR for removal of line numbers)

The hook should not be dependent on the `pre-commit` framework, which I intend to address via better documentation.

tests/audit/io_test.py
```diff
@@ -12,7 +12,7 @@
                 'allow_backstep': True,
             },
             (
-                'Is this a valid secret (not a false-positive)? '
+                'Is this a secret that should be committed to this repository? '
```
idavidmcdonald, line 15: This seems much clearer to me although let me just give an example to make sure I understand it correctly.

If I had accidentally copied a production AWS key into my code that we wouldn't want to fall into the wrong hands and the scan found it and asked the question 'Is this a secret that should be committed to this repository?', then I would answer 'no'. Have I understood that correctly?

One thing I wanted to check, if I had been asked 'Is this a valid secret (not a false-positive)?' instead I would have answered 'yes'. Does this mean that the 'yes' and 'no' options have now swapped?

domanchi: 🤦 You're absolutely right! This language would swap the `yes` and `no` options. Let me fix this.

domanchi:
Tests pass, and issues resolved. Merging.

tests/audit/io_test.py
idavidmcdonald: Thanks, so just to make sure I understand, we are going to go with this new wording (which I'm happy with) but it will require you to go and make an additional change to swap the logic for `yes` and `no` options?

idavidmcdonald: Ah infact, I see you added that fix as a commit in this PR. Great, this is looking good!

PR #364, opened 2020-11-14
Adding --list-all-plugins

domanchi:
## Summary

Adds `--list-all-plugins` argument to list all plugins used.

```
$ detect-secrets scan --list-all-plugins
ArtifactoryDetector
AWSKeyDetector
AzureStorageKeyDetector
BasicAuthDetector
CloudantDetector
Base64HighEntropyString
HexHighEntropyString
IbmCloudIamDetector
IbmCosHmacDetector
JwtTokenDetector
KeywordDetector
MailchimpDetector
NpmDetector
PrivateKeyDetector
SlackDetector
SoftlayerDetector
StripeDetector
TwilioKeyDetector
```

or with baseline support:

```
$ detect-secrets scan --disabled-plugins KeywordDetector test_data > test.baseline
$ detect-secrets scan --list-all-plugins --baseline test.baseline | grep 'KeywordDetector'
```

This opens the path for CLI usage as such:

```
$ detect-secrets scan --list-all-plugins | \
    grep -v 'BasicAuthDetector' | \
    sed "s#^#--disable-plugin #g | \
    xargs detect-secrets scan test_data
```

so you can run a specific plugin, without needing to disable everything manually (or writing some funky logic to resolve enablement / disablement of plugins).

### DIFF
--- a/detect_secrets/core/usage/plugins.py
+++ b/detect_secrets/core/usage/plugins.py
@@ -15,6 +15,12 @@ def add_plugin_options(parent: argparse.ArgumentParser) -> None:
         ),
     )
 
+    parser.add_argument(
+        '--list-all-plugins',
+        action='store_true',
+        help='Lists all plugins that will be used for the scan.',
+    )
+
     _add_custom_limits(parser)
     _add_disable_flag(parser)
     # TODO: custom plugins?
--- a/detect_secrets/main.py
+++ b/detect_secrets/main.py
@@ -12,6 +12,7 @@ from .core.scan import get_plugins
 from .core.scan import scan_line
 from .core.usage import ParserBuilder
 from .exceptions import InvalidBaselineError
+from .settings import get_settings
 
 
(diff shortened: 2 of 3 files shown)


OiCMudkips (approved):
Seems fine as-is, but does it make sense to make this another command instead? (`detect-secrets list-all-plugins`). We aren't scanning anything here so `detect-secrets scan --list-all-plugins` is a little confusing. Plus the former is shorter :)

domanchi:
Yeah, I was pondering where to put this.

On one hand, I think subparsers should be used if there were other flags that it can be used with. However, this functionality seems pretty clear cut. I was also thinking of putting it at the root parser (e.g. `detect-secrets --list-all-plugins`) but there is technically a difference between listing all plugins for `audit`, and listing all plugins for `scan`.

As such, I'll probably merge this as is. We can change it later if more functionality comes up. :shrug:

Issue #319, opened 2020-07-15
[question] How to skip password detection in Java Properties File ?


sylvainmouquet:
Hello,

We have tried to allow a secret in a property file but java read the full string `"xyz # pragram: allowlist secrets"`
```
password=xyz # pragma: allowlist secret
```

How to manage it ?

Thanks
Sylvain

OiCMudkips:
If you are using pre-commit you can use the `exclude` option to exclude Java properties files. If you are using this tool directly, you can use the [--exclude-files](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/usage.py#L247-L254) flag.

Let us know if this works for you.

prashpb:
The same issue we are facing.
we cannot add exclude files as user may commit actual passwords and the scan will not detect it.
Any solution on how to add whitelist to .properties file will be helpful.

domanchi:
From https://en.wikipedia.org/wiki/.properties, it looks like the issue with this is that the Java `.properties` doesn't allow inline comments. This mirrors other config-file like files. For example:

```python
>>> import configparser
>>> parser = configparser.ConfigParser()
>>> parser.read_string("""\
...    [section]
...    password = xyz    # pragma: allowlist secret
... """)

>>> print(parser.items('section')[0])
('password', 'xyz  # pragma: allowlist secret')
```

The only way we can address this is by adding some form of "skip next line" pragma, like `eslint-disable-next-line`. Let me see whether I can address this in our new filters system in v1.

nickiaconis:
Fwiw I just encountered this same issue in a Dockerfile today. It will be very helpful to have a "skip next line" pragma. Anything I can do to help move this along?

domanchi:
@nickiaconis : if you wanted to help, you can certainly contribute a PR that bases atop the current `pre-v1-adding-verification` branch -- this will give you access to the dependency injected `context` variable (more information here: https://github.com/Yelp/detect-secrets/blob/pre-v1-launch/docs/filters.md#a-word-about-filter-internals

I'd imagine that with this context object, you can create a new filter that checks the line before to see if the flagged secret has been ignored.

PR #367, opened 2020-11-19
add a "skip next line" pragma

nickiaconis:
Some configuration files, like Java's `.properties` files and Docker's `Dockerfile`, don't allow inline comments. This PR adds support for a "skip next line" pragma to solve this use case. For example:

```
# pragma: allowlist nextline secret
password=xyz
```

This should solve #319.

### DIFF
--- a/detect_secrets/core/scan.py
+++ b/detect_secrets/core/scan.py
@@ -162,7 +162,15 @@ def _process_line_based_plugins(
 
         # We apply line-specific filters, and see whether that allows us to quit early.
         if any([
-            inject_variables_into_function(filter_fn, filename=filename, line=line)
+            inject_variables_into_function(
+                filter_fn,
+                filename=filename,
+                line=line,
+                context=get_code_snippet(
+                    lines=line_content,
+                    line_number=line_number,
+                ),
+            )
             for filter_fn in get_filters_with_parameter('line')
         ]):
             continue
--- a/detect_secrets/core/scan.py
+++ b/detect_secrets/core/scan.py
@@ -159,6 +159,10 @@ def _process_line_based_plugins(
     # filters return True.
     for line_number, line in lines:
         line = line.rstrip()
+        code_snippet = get_code_snippet(
+            lines=line_content,
+            line_number=line_number,
+        )
 
         # We apply line-specific filters, and see whether that allows us to quit early.
         if any([
@@ -166,10 +170,7 @@ def _process_line_based_plugins(
                 filter_fn,
                 filename=filename,
                 line=line,
-                context=get_code_snippet(
-                    lines=line_content,
-                    line_number=line_number,
-                ),
+                context=code_snippet,
             )
             for filter_fn in get_filters_with_parameter('line')
         ]):
@@ -184,10 +185,7 @@ def _process_line_based_plugins(
                         secret=secret.secret_value,
                         plugin=plugin,
                         line=line,
-                        context=get_code_snippet(
-                            lines=line_content,
-                            line_number=line_number,
-                        ),
+                        context=code_snippet,
                     ):
                         log.debug(f'Skipping "{secret.secret_value}" due to `{filter_fn.path}`.')
                         break
--- a/detect_secrets/filters/allowlist.py
+++ b/detect_secrets/filters/allowlist.py
@@ -5,8 +5,10 @@ from typing import Dict
 from typing import List
 from typing import Pattern
 
+from ..util.code_snippet import CodeSnippet
 
-def is_line_allowlisted(filename: str, line: str) -> bool:
+
+def is_line_allowlisted(filename: str, line: str, context: CodeSnippet) -> bool:
     regexes = _get_allowlist_regexes()
 
     _, ext = os.path.splitext(filename)
@@ -15,6 +17,16 @@ def is_line_allowlisted(filename: str, line: str) -> bool:
 
     for regex in regexes:
         if regex.search(line):
+            return True
+
+    previous_line = context.previous_line
+    regexes = _get_allowlist_nextline_regexes()
+
+    if ext[1:] in _get_file_based_allowlist_nextline_regexes():
+        regexes = _get_file_based_allowlist_nextline_regexes()[ext[1:]]
+
+    for regex in regexes:
+        if regex.search(previous_line):
             return True
 
     return False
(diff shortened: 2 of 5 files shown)


tests/filters/allowlist_filter_test.py
```diff
@@ -22,16 +23,57 @@
...
+        ('// ', ' more text'),
+
+        ('/*', '*/'),
+        ('/* ', ' */'),
+
+        ('--', ''),
+        ('-- ', ' more text'),
+
+        ('<!--', '-->'),
+    ),
```
domanchi, line 50: Same thing goes for DRY code. Can we move this into a (global) variable?

detect_secrets/filters/allowlist.py
```diff
@@ -17,6 +19,16 @@ def is_line_allowlisted(filename: str, line: str) -> bool:
...
 
+    previous_line = context.previous_line
+    regexes = _get_allowlist_nextline_regexes()
+
+    if ext[1:] in _get_file_based_allowlist_nextline_regexes():
+        regexes = _get_file_based_allowlist_nextline_regexes()[ext[1:]]
+
+    for regex in regexes:
+        if regex.search(previous_line):
+            return True
```
domanchi, line 30: Can we extract this into a function? I'm thinking something like:

```python
def _get_regexes_for_file(filename: str) -> Tuple[Iterable[Pattern], Iterable[Pattern]]:
    yield ...
```

then we can do:

```python
for payload, regex in zip([line, previous_line], _get_regexes_for_file(filename)):
    if regex.search(payload):
        return True
```

or something that allows us to take advantage of mostly similar code.

domanchi (commented):
Great turnaround on this!

Besides stylistic comments, you will also need to modify https://github.com/Yelp/detect-secrets/pull/356/files#diff-bad716776100dfb6fdad237381526e524042316894d806953446d39984aabe27R165 to inject `context` into the filter -- otherwise, this will run per plugin, which is not something we want.

detect_secrets/filters/allowlist.py
nickiaconis: It's been a while since I last wrote python, so hopefully I implemented this along the lines of what you were thinking. (Aside: thank goodness there are tests to verify whether it's working 😂) Please let me know if there's anything I ought to change 🙂

detect_secrets/filters/allowlist.py
```diff
@@ -2,100 +2,83 @@
...
-    ]
+# Note: Cache size should be 2x the number of comment types
+@lru_cache(maxsize=12)
+def _get_allowlist_regexes(comment_tuple: Tuple[str, str], nextline: bool) -> Pattern:
+    start = comment_tuple[0]
+    end = comment_tuple[1]
+    return re.compile(
+        # Note: Always use allowlist, whitelist will be deprecated in the future
+        r'{}{} *pragma: ?{}{}[ -]secret.*?{}[ \t]*$'.format(
+            r'^[ \t]*' if nextline else r'[ \t]+',
```
domanchi, line 78: I think this can be the same. I mean, this is just **requiring** whitespace for the pragma, which technically is not necessary.

detect_secrets/filters/allowlist.py
```diff
@@ -2,100 +2,83 @@
...
+            start,
+            r'allowlist' if nextline else r'(allow|white)list',
```
domanchi, line 80: I'm also fine if these are the same. Supporting `whitelist` is just for backwards compatibility, and I'm fine with keeping that backwards compatibility around for the nextline regex too.

domanchi (commented):
@nickiaconis, please also modify https://github.com/Yelp/detect-secrets/pull/356/files#diff-bad716776100dfb6fdad237381526e524042316894d806953446d39984aabe27R165 to include the `context` variable in there -- otherwise, this PR won't work.

detect_secrets/filters/allowlist.py
nickiaconis: There is also the difference in that the one on the left requires that nothing else precede the pragma. I think this is valuable to avoid situations that might obscure what is being allowed. For example, this is not allowed by what I've written, but would be allowed if we use `[ \t]*` for both:

```
foo = 'bar' # pragma: allowlist nextline secret
pass = 'kajs8urailjw4euhgy8aueoijrhua8y'
```

detect_secrets/filters/allowlist.py
nickiaconis: I don't mind one way or the other, but my thinking was that, b/c `nextline` is new, it doesn't have any backward compatibility to contend with. Thus, requiring `allowlist` for `nextline` prepares uses of it for the future removal of `whitelist`, whereas allowing both opens the possibility of further entrenching `whitelist`.

detect_secrets/filters/allowlist.py
domanchi: ooo. good point! We should definitely capture that rationale in the code because that wasn't apparent in reading it.

nickiaconis:
Totally missed the top comment and link to the diff in your previous review. Sorry about that. 😅

Just to double check, is this the line in question? https://github.com/Yelp/detect-secrets/blob/pre-v1-adding-verification/detect_secrets/core/scan.py#L165 It looks like GH can't scroll to a line in a diff that's not part of the diff context.

domanchi:
Yup, that's the one.

nickiaconis:
Hey @domanchi, just want to follow up on this. Is there anything left unresolved at this point?

detect_secrets/core/scan.py
```diff
@@ -162,7 +162,15 @@ def _process_line_based_plugins(
...
         if any([
-            inject_variables_into_function(filter_fn, filename=filename, line=line)
+            inject_variables_into_function(
+                filter_fn,
+                filename=filename,
+                line=line,
+                context=get_code_snippet(
+                    lines=line_content,
+                    line_number=line_number,
+                ),
```
domanchi, line 172: Since we use it later on, it's probably worth just creating a variable for this, so we can use it later on as well.

detect_secrets/filters/allowlist.py
```diff
@@ -2,54 +2,84 @@
...
+
+# Note: Cache size should be 2x the number of comment types
+@lru_cache(maxsize=12)
+def _get_allowlist_regexes(comment_tuple: Tuple[str, str], nextline: bool) -> Pattern:
+    start = comment_tuple[0]
+    end = comment_tuple[1]
+    return re.compile(
+        r'{}[ \t]*{} *pragma: ?{}{}[ -]secret.*?{}[ \t]*$'.format(
+            # Note: No text can precede a nextline pragma, this prevents obscuring what is allowed
+            r'^' if nextline else '',
```
domanchi, line 78:
```suggestion
            # Note: No text can precede a nextline pragma, this prevents obscuring what is allowed
            # For instance, we want to prevent the following case from working:
            #     foo = 'bar' # pragma: allowlist nextline secret
            #     pass = 'hunter2'
            r'^' if nextline else '',
```

domanchi (approved):
Thanks for the follow up -- I must have missed the notification that you were able to make the changes.

detect_secrets/filters/allowlist.py
nickiaconis: TIL about this Github feature. Nice! 👍

nickiaconis:
I think that's everything, but just lmk if there's anything more to be done here. 🙂

domanchi:
Ah whoops. Did *not* mean to close this PR by cleaning up my branches. I will merge your changes manually.

domanchi:
This has been merged: https://github.com/Yelp/detect-secrets/pull/355/commits/b73c473c1df6c417c69391ce447a9fc042f9e1ef

sfc-gh-nsfard:
Hello @domanchi ! When can we expect to see this change published?

PR #376, opened 2020-12-02
Deprecate whitelist pragma and hyphen in pragma

### LINKED ISSUE #375: Deprecate white/blacklist in v1
We are releasing v1 of this tool. Let's deprecate support for white/blacklist terminology while we're at it.

OiCMudkips:
This removes support for whitelisting lines using
the "whitelist" pragma. Users should switch to "allowlist".

This also removes support for the hyphen in the
allowlist pragma (e.g. "# allowlist-secret") which apparently
we supported, but wasn't documented anywhere.

Closes #375

Tested with `tox -e py38`.

### DIFF
--- a/detect_secrets/filters/allowlist.py
+++ b/detect_secrets/filters/allowlist.py
@@ -35,8 +35,7 @@ def _get_allowlist_regexes() -> List[Pattern]:
     return [
         re.compile(r)
         for r in [
-            # Note: Always use allowlist, whitelist will be deprecated in the future
-            r'[ \t]+{} *pragma: ?(allow|white)list[ -]secret.*?{}[ \t]*$'.format(start, end)
+            r'[ \t]+{} *pragma: allowlist secret.*?{}[ \t]*$'.format(start, end)
             for start, end in (
                 ('#', ''),                    # e.g. python or yaml
                 ('//', ''),                   # e.g. golang
--- a/tests/filters/allowlist_filter_test.py
+++ b/tests/filters/allowlist_filter_test.py
@@ -28,13 +28,6 @@ def test_basic(prefix, suffix):
     )
 
 
-def test_backwards_compatibility():
-    assert is_line_allowlisted(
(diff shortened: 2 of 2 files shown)


domanchi:
While the v1 release is a great time to kill backwards-compatibility support, I worry that this change will make it significantly harder for clients to update their tool. Thoughts on this?

KevinHock:
Thoughts @killuazhu? You might bring a different perspective on this given how you’ve used the tool

The hyphen possibility removal might be somewhat unnecessary, people can use surprisingly undocumented features of tools, no strong opinions though.

killuazhu:
@KevinHock Thanks for tagging. I don’t have exact data of how many users are still using the legacy whitelist syntax. But we’ve been promoting the new allowlist since early this year. Some of our users are also quite passionate about moving away from the whitelist syntax. I think this change, although breaking, makes a lot of sense. I would support to move forward with it.

@justineyster FYI

OiCMudkips:
> While the v1 release is a great time to kill backwards-compatibility support, I worry that this change will make it significantly harder for clients to update their tool. Thoughts on this?

I agree it's possible that it would make the upgrade harder. How would you feel if I added an adhoc upgrade script that can be included in the 1.0.0 release (and probably removed in subsequent releases)? The adhoc upgrade script would basically find all filtered out lines using the filter function, and then do the appropriate substitution on the line.

domanchi:
That's actually a really good idea. You can probably write your logic in `detect_secrets/core/upgrades/v1_0.py`, and leverage the functionality that powers `--only-allowlisted` to narrow down those lines.

killuazhu:
> I agree it's possible that it would make the upgrade harder. How would you feel if I added an adhoc upgrade script that can be included in the 1.0.0 release (and probably removed in subsequent releases)? The adhoc upgrade script would basically find all filtered out lines using the filter function, and then do the appropriate substitution on the line.

Like this idea!

PR #377, opened 2020-12-02
Migrating Keyword False Positives

domanchi:
`KeywordDetector` had a variety of different false positive heuristics that only applied to `KeywordDetector`. This PR moves it to the new filters architecture.

It also drops support for a hard-coded list of false positives (e.g. `fake`) given that we have `detect_secrets.filters.wordlist`.


tests/plugins/keyword_test.py
```diff
@@ -227,105 +218,24 @@ def test_analyze_objective_c_positives(self, file_content):
...
-                '.yaml',
-                '.yml',
-            )
-        ),
-    )
-    def test_analyze_yaml_negatives(self, file_content, file_extension):
-        assert not KeywordDetector().analyze_line(
-            filename=f'mock_filename{file_extension}',
-            line=file_content,
-        )
```
domanchi, line 316: @KevinHock : do you remember why this test was supposed to behave as such? I spent some time reading the code, but I couldn't understand why YAML files would be treated differently, and that lines that would have been caught in a non-YAML file will be ignored with YAML files.

KevinHock: I added {{ to prefix the standard positives test, and think I made this test to test we skipped them.

I think this had to do specifically with 
```
        ) or (
            filetype == FileType.YAML
            and lowered_secret.startswith('{{')
            and lowered_secret.endswith('}}')
        )
```
at the time

https://github.com/Yelp/detect-secrets/commit/592d077084f4f66d4175d2e6b7055a8529f0ade7#diff-e6de82b29d51ff0d50bb9a4951479046f43f6dad3d40ba27e3dd7e2a9f6fda8bR226

domanchi: Yeah, that's so bizarre.

Looks like it was [removed](https://github.com/Yelp/detect-secrets/commit/a53d12e65c450f5daea44290e7ba0cee744fe686#diff-e6de82b29d51ff0d50bb9a4951479046f43f6dad3d40ba27e3dd7e2a9f6fda8bL308-L310) though, so I don't know how tests kept passing.

I'm going to leave it out for now, and migrate the filters based on the master branch, and the active false positive filters on this plugin.

detect_secrets/filters/heuristic.py
```diff
@@ -113,3 +113,53 @@ def is_non_text_file(filename: str) -> bool:
         '.zip',
     ),
 )
+
+
+def is_templated_secret(secret: str) -> bool:
+    """
+    Filters secrets that are shaped like: {secret}, <secret>, or ${secret}.
+    """
+    if secret[0] == '$':
```
tanx16, line 122: Did you forget to remove this after copying it to the new `is_prefixed_with_dollar_sign ` function?

tests/filters/heuristic_filter_test.py
```diff
@@ -79,3 +81,49 @@ def test_success(self, secret, line):
...
         assert not filters.heuristic.is_likely_id_string(secret, line)
+
+
+@pytest.mark.parametrize(
+    'line, result',
+    (
+        ('secret = $hunter2', False),
+        ('secret = {hunter2}', False),
+        ('secret = <hunter2>', False),
+        ('secret = hunter2', True),
```
tanx16, line 92: Maybe add a test for `secret= ${hunter2}`?

detect_secrets/filters/heuristic.py
```diff
@@ -113,3 +113,53 @@ def is_non_text_file(filename: str) -> bool:
...
+        return True
+
+    if (
+        (secret[0] == '{' and secret[-1] == '}')
+        or (secret[0] == '<' and secret[-1] == '>')
+        or (secret[0] == '$' and secret[1] == '{' and secret[-1] == '}')
```
tanx16, line 128: Is it possible for a secret to be 1 character and `IndexError`ing on this line?

detect_secrets/filters/heuristic.py
domanchi: You're absolutely right. ++

detect_secrets/filters/heuristic.py
domanchi: It's a fair point, but like super unlikely. Entropy scanners definitely require more than one character to trip it up, and regexes should require more than one character to meet the bar. After all, a one-character secret is a terrible secret.

I'll add a try/except clause for any keyword filter trip ups.

tanx16 (approved):

PR #380, opened 2020-12-12
Better scanning UX

domanchi:
This improves the scanning usability of this tool by allowing scans for non-tracked git files, without passing a flag.

e.g.

```bash
$ echo "secret = 'blah'" > blah.py
$ detect-secrets scan blah.py
...
  "results": {
    "blah.py": [
      {
        "type": "Secret Keyword",
        "filename": "blah.py",
        "hashed_secret": "aaf4c61ddcc5e8a2dabede0f3b482cd9aea9434d",
        "is_verified": false,
        "line_number": 1
      }
    ]
  }
}
```

This still does not address the ability to scan files outside the current git repository, because I haven't figured what to do with the baseline files. That is, the implicit assumption for the baseline is that it is run at git root, so that all the files can be found and navigated to with the audit / pre-commit tool. It's really unclear what it would mean if this is run from a non-git repository.

### DIFF
--- a/detect_secrets/core/scan.py
+++ b/detect_secrets/core/scan.py
@@ -53,10 +53,13 @@ def get_files_to_scan(
 
     See test cases for more details.
     """
+    # First, we determine the appropriate filtering mode to be used.
     # If this is True, then it will consider everything to be valid.
     # Otherwise, it will only list the files that are valid.
     valid_paths: Union[bool, Set[str]] = True
     for path in paths:
+        # Since this is not a directory, we assume that it is a file proper, and automatically
+        # consider it valid.
         if not os.path.isdir(path):
             continue
 
@@ -67,6 +70,8 @@ def get_files_to_scan(
                 log.warning('Did not detect git repository. Try scanning all files instead.')
                 valid_paths = False
 
+        # Since valid_paths attempts to get *all* tracked files in the repository, we just need
+        # to initialize it once.
         break
 
     if not valid_paths:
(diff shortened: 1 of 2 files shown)


detect_secrets/core/scan.py
```diff
@@ -26,21 +28,58 @@
...
+    override this behavior, we can specify `should_scan_all_files=True`, which will force
+    the scan for all files.
+
+    See test cases for more details.
+    """
+    # If this is True, then it will consider everything to be valid.
+    # Otherwise, it will only list the files that are valid.
+    valid_paths: Union[bool, Set[str]] = True
+    for path in paths:
+        if not os.path.isdir(path):
```
tanx16, line 60: Maybe this would be clearer if you used `if os.path.isdir(path):`

tests/core/scan_test.py
```diff
@@ -1,33 +1,77 @@
...
 
-def test_handles_broken_yaml_gracefully():
-    with tempfile.NamedTemporaryFile(suffix='.yaml') as f:
-        f.write(
-            textwrap.dedent("""
-            metadata:
-                name: {{ .values.name }}
-            """)[1:].encode(),
+class TestGetFilesToScan:
+    @staticmethod
```
tanx16, line 14: Could we add a test that tries to call the function on a directory and a file at the same time?

detect_secrets/core/scan.py
domanchi: added more comments

tanx16 (approved):

PR #381, opened 2020-12-12
Ensuring consistent serialized baselines

domanchi:
### Summary

Consistent, serialized baselines are important. Otherwise, when the pre-commit hook updates the baselines, irrelevant changes (caused by unstable sorts) may unnecessarily cause the commit to fail. We now test this programmatically by performing a round trip serialization to ensure baselines are identical.

In other news, this PR means that we finally have a stable baseline format for v1!

### Details

Within this PR, I also included bug fixes for:

- Missing default limit for `HighEntropyString` plugin serialization.

  Previously, when `HighEntropyString` plugins were serialized, it would look something like this:

  ```json
  {
    "name": "Base64HighEntropyString"
  }
  ```

  While this works, it is somewhat fragile since this would cause different scan results if the engine changed its default value. As such, we now explicitly include the limit for initialization, so results are more sturdy to version changes.

- `KeywordDetector` `keyword_exclude` now serializes to a string, not a list

  Ah, the magic hanging comma, that accidentally transformed a multi-line string into a tuple. This fixes it so that `KeywordDetector` can be initialized properly from the baseline.

### DIFF
--- a/tests/core/baseline_test.py
+++ b/tests/core/baseline_test.py
@@ -80,18 +80,24 @@ class TestCreate:
             assert get_relative_path_if_in_cwd(f.name) in secrets.data
 
 
-@pytest.mark.xfail(
-    reason=(
-        'TODO: When we\'re done with changes, we\'ll update .secrets.baseline and this should '
-        'work then.'
-    ),
-)
 def test_load_and_output():
     with open('.secrets.baseline') as f:
         data = json.loads(f.read())
 
-    secrets = baseline.load(data)
-    assert baseline.format_for_output(secrets) == data
+    secrets = baseline.load(data, filename='.secrets.baseline')
+    output = baseline.format_for_output(secrets)
+
+    for item in [data, output]:
+        item.pop('generated_at')
+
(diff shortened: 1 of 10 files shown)


tests/core/baseline_test.py
```diff
@@ -80,18 +80,24 @@ def test_scan_all_files():
...
     with open('.secrets.baseline') as f:
         data = json.loads(f.read())
 
-    secrets = baseline.load(data)
-    assert baseline.format_for_output(secrets) == data
+    secrets = baseline.load(data, filename='.secrets.baseline')
+    output = baseline.format_for_output(secrets)
+
+    for item in [data, output]:
+        item.pop('generated_at')
```
calvinli, line 91: FWIW I think this does point out a potential problem, which is you're always going to have a merge-conflict due to `generated_at` if two branches both touch the baseline file.

calvinli (approved):
It seems safer to me to not worry about ordering at the dict level and just output it with `json.dumps(..., sort_keys=True)` (then all your dicts are automatically sorted by key and you only have to worry about lists)

But this looks fine too

PR #385, opened 2020-12-15
cache verification of IBM Cloud IAM keys

gpflaum:
Avoid repeated calls to https://iam.cloud.ibm.com/identity/token with the same token.

The regex used in [IbmCloudIamDetector](https://github.com/Yelp/detect-secrets/blob/68f3d5b7db41617da619190bddc52165682142a0/detect_secrets/plugins/ibm_cloud_iam.py#L21) has many optional terms, and will match any `key|pwd|password|pass|token` keyword followed by a 44 character string of letters, digits, underscores or dashes. I have repos with a lot of files with duplicate false-positives like this one:

```
# represented by key ConstructionCode_NON_CLF_Service_Create_Desc
```
Caching the token verification reduces the number of times we hit https://iam.cloud.ibm.com/identity/token

@killuazhu

### DIFF
--- a/detect_secrets/plugins/ibm_cloud_iam.py
+++ b/detect_secrets/plugins/ibm_cloud_iam.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 from typing import cast
 from typing import Union
 
@@ -34,6 +35,7 @@ class IbmCloudIamDetector(RegexBasedDetector):
             else VerifiedResult.VERIFIED_FALSE
 
 
+@lru_cache()
 def verify_cloud_iam_api_key(apikey: Union[str, bytes]) -> requests.Response:  # pragma: no cover
     if type(apikey) == bytes:
         apikey = cast(bytes, apikey).decode('UTF-8')


domanchi (approved):

domanchi:
Looks like travis has temporarily disabled builds -- will need to manually run tests before merging.

detect_secrets/plugins/ibm_cloud_iam.py
```diff
@@ -34,6 +35,7 @@ def verify(self, secret: str) -> VerifiedResult:
             else VerifiedResult.VERIFIED_FALSE
 
 
+@lru_cache()
```
killuazhu, line 38: I don't think we should enforce caching without given caller an option. Here are my reasons 

1. IBM Cloud token supports [lock/unlock an API key](https://cloud.ibm.com/docs/account?topic=account-userapikey#lock_user_key). You can get different results for the same key
2. detect-secrets plugins can be used as a library. We internally using it as a supporting library for a long run application to continuously scanning tokens. Mandating the cache will prevent us validate a token when it has switched from locked to unlocked state.

I would suggest we either filter off some keywords or use filter plugins during scan to avoid false positives, or provide a way to allow caller to use verify without using cache (solve case 2 ☝️ ).

killuazhu (changes requested):
See inline comment.

detect_secrets/plugins/ibm_cloud_iam.py
domanchi: killuazhu++

gpflaum:
Closing this PR, as killuazhu's objections are valid but I don't need this enough to continue work on it. (We aren't using IBM Cloud, so I'll disable the plugin for my scans.)

PR #386, opened 2020-12-16
improved adhoc string scanning

domanchi:
## Summary

Adhoc string scanning is hella useful. e.g.

```bash
$ detect-secrets scan --string 'butactuallynotasecret'
...
Base64HighEntropyString: False
```

However, sometimes (especially for the HighEntropyString plugins), we want to see *why* it failed. This is existing behavior in 0.14.3; this PR carries this functionality forward. With this change, it will be:

```bash
$ detect-secrets scan --string 'butactuallynotasecret'
...
Base64HighEntropyString: False (3.404)
```

## Reviewer Notes

It was somewhat difficult to surface the non-secret to the scan layer in a generic manner (i.e. without using `if plugin in {Base64HighEntropyString}`), so I had to be creative with the `analyze_line` and `analyze_string` functionality. As a result, for people using the `analyze_string` function directly (like the test case), they would need to be responsible for filtering out the secret themselves.

Not ideal, but it's the best way I could find.

This PR also introduces another gotcha -- for our DI system, we now leverage the pythonic fashion of duck-typing to determine whether there are enough required parameters to inject into the function. This needed to be done to support cases where the underlying function had default values for their parameters (e.g. `HighEntropyStringsPlugin.analyze_line`), and as such, we don't need to supply **all** the parameters that a function needs. However, this also means that if the underlying function raises an uncaught `TypeError`, it will be silenced by the DI system.

I guess future improvements could be done to distinguish default parameters from non-default parameters in a function declaration, but I didn't want to dive too deep into Python magic internals for this PR.

### DIFF
--- a/detect_secrets/core/scan.py
+++ b/detect_secrets/core/scan.py
@@ -67,30 +67,28 @@ def scan_line(line: str) -> Generator[PotentialSecret, None, None]:
     )
     get_filters.cache_clear()
 
-    for plugin in get_plugins():
+    yield from (
+        secret
+        for plugin in get_plugins()
         for secret in _scan_line(
             plugin=plugin,
             filename='adhoc-string-scan',
             line=line,
             line_number=0,
             enable_eager_search=True,
-        ):
-            for filter_fn in get_filters_with_parameter('context'):
-                if call_function_with_arguments(
-                    filter_fn,
-                    filename=secret.filename,
-                    secret=secret.secret_value,
-                    plugin=plugin,
-                    line=line,
-                    context=get_code_snippet(
-                        lines=[line],
-                        line_number=1,
-                    ),
-                ):
-                    log.debug(f'Skipping "{secret.secret_value}" due to `{filter_fn.path}`.')
-                    break
-            else:
-                yield secret
+        )
+        if not _is_filtered_out(
+            required_filter_parameters=['context'],
+            filename=secret.filename,
+            secret=secret.secret_value,
+            plugin=plugin,
+            line=line,
+            context=get_code_snippet(
+                lines=[line],
+                line_number=1,
+            ),
+        )
+    )
 
 
 def scan_file(filename: str) -> Generator[PotentialSecret, None, None]:
@@ -98,7 +96,7 @@ def scan_file(filename: str) -> Generator[PotentialSecret, None, None]:
         log.warning('No plugins to scan with!')
         return
 
-    if _filter_files(filename):
+    if _is_filtered_out(required_filter_parameters=['filename'], filename=filename):
         return
 
     try:
@@ -143,7 +141,10 @@ def scan_for_allowlisted_secrets_in_file(filename: str) -> Generator[PotentialSe
         log.warning('No plugins to scan with!')
         return
 
-    if _filter_files(filename):
+    if _is_filtered_out(
+        required_filter_parameters=['filename'],
+        filename=filename,
+    ):
         return
 
     # NOTE: Unlike `scan_file`, we don't ever have to use eager file transfomers, since we already
@@ -185,10 +186,7 @@ def _scan_for_allowlisted_secrets_in_lines(
         ):
             continue
 
-        if any([
-            call_function_with_arguments(filter_fn, filename=filename, line=line)
-            for filter_fn in get_filters_with_parameter('line')
-        ]):
+        if _is_filtered_out(required_filter_parameters=['line'], filename=filename, line=line):
             continue
 
         for plugin in get_plugins():
@@ -236,7 +234,7 @@ def _get_lines_from_diff(diff: str) -> Generator[Tuple[str, List[Tuple[int, str]
     patch_set = PatchSet.from_string(diff)
     for patch_file in patch_set:
         filename = patch_file.path
-        if _filter_files(filename):
+        if _is_filtered_out(required_filter_parameters=['filename'], filename=filename):
             continue
 
         yield (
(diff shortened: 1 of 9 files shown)


detect_secrets/core/scan.py
```diff
@@ -326,14 +328,28 @@ def _scan_line(
     filename: str,
     line: str,
     line_number: int,
+    **kwargs: Any,
 ) -> Generator[PotentialSecret, None, None]:
     # NOTE: We don't apply filter functions here yet, because we don't have any filters
     # that operate on (filename, line, plugin) without `secret`
     try:
-        secrets = plugin.analyze_line(filename=filename, line=line, line_number=line_number)
+        analyze_line = plugin.__class__.analyze_line
```
calvinli, line 336: This could all use some comments explaining why it's necessary, considering we've gone from 1 line to 14 lines involving "Python magic internals" as you say

breaking it out into a helper function might help with the readability too

detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -49,11 +47,25 @@ def analyze_line(
     ) -> Set[PotentialSecret]:
         output = super().analyze_line(filename=filename, line=line, line_number=line_number)
         if output or not enable_eager_search:
-            return output
+            # NOTE: We perform the limit filter at this layer (rather than analyze_string) so
+            # that we can surface secrets that do not meet the limit criteria when
+            # enable_eager_search=True.
+            return {
+                secret
+                for secret in (output or set([]))
```
calvinli, line 55: just `set()` or just `[]` should work

detect_secrets/plugins/high_entropy_strings.py
```diff
@@ -35,9 +35,7 @@ def analyze_string(self, string: str) -> Generator[str, None, None]:
                 # This occurs on the default regex, but not on the eager regex.
                 result = result[1]
 
-            entropy_value = self.calculate_shannon_entropy(result)
-            if entropy_value > self.entropy_limit:
-                yield result
+            yield result
```
calvinli, line 38: Can you leave a comment here too explaining that the entropy check moved to `analyze_line`? Because from the base class people are going to assume this is supposed to only yield secrets

detect_secrets/core/scan.py
```diff
@@ -326,14 +328,28 @@ def _scan_line(
...
+        if not hasattr(analyze_line, 'injectable_variables'):
+            analyze_line.injectable_variables = set(        # type: ignore
+                get_injectable_variables(plugin.analyze_line),
+            )
+            analyze_line.path = f'{plugin.__class__.__name__}.analyze_line'     # type: ignore
     except AttributeError:
         return
 
+    secrets = inject_variables_into_function(
```
calvinli, line 345: nit: i guess this function isn't new in this PR but it would be clearer as like, `call_function_with_arguments(...)` or something, because it wasn't obvious to me reading this that what this was doing was calling `analyze_line`

detect_secrets/core/scan.py
calvinli: in particular the part where we go from `plugin.analyze_line` -> `plugin.__class__.analyze_line` could use some explanation

detect_secrets/core/scan.py
```diff
@@ -326,14 +328,28 @@ def _scan_line(
...
     # that operate on (filename, line, plugin) without `secret`
     try:
-        secrets = plugin.analyze_line(filename=filename, line=line, line_number=line_number)
+        analyze_line = plugin.__class__.analyze_line
+        if not hasattr(analyze_line, 'injectable_variables'):
+            analyze_line.injectable_variables = set(        # type: ignore
+                get_injectable_variables(plugin.analyze_line),
+            )
+            analyze_line.path = f'{plugin.__class__.__name__}.analyze_line'     # type: ignore
     except AttributeError:
```
calvinli, line 342: does this `except` need to be after `secrets = inject_variables_into_function(...)`?

detect_secrets/plugins/base.py
```diff
@@ -71,10 +70,7 @@ def json(self) -> Dict[str, Any]:
             'name': self.__class__.__name__,
         }
 
-    def format_scan_result(self, secret: Optional[PotentialSecret]) -> str:
-        if not secret:
-            return 'False'
-
+    def format_scan_result(self, secret: PotentialSecret) -> str:
```
calvinli, line 73: can you explain what changed to make `None` no longer an allowed arg here? having trouble seeing it

detect_secrets/util/inject.py
```diff
@@ -13,11 +13,12 @@ def inject_variables_into_function(func: SelfAwareCallable, **kwargs: Any) -> Op
         for key in (variables_to_inject & func.injectable_variables)
     }
 
-    if set(values.keys()) != func.injectable_variables:
+    try:
+        return func(**values)
+    except TypeError:
+        # When parameters are required (and not supplied), it raises a TypeError.
         return None
```
calvinli, line 20: guess it's a little late for this but: why swallow the TypeError? if you don't call a function correctly it seems reasonable for that to raise TypeError, returning `None` seems a bit more surprising especially if the underlying function type signature doesn't even have `None` as an allowed return value

calvinli: and this is way off topic but it might be possible to use mypy generics to get rid of that `-> Optional[Any]`?

calvinli (commented):
looks fine overall, just have a few comments

detect_secrets/core/scan.py
domanchi: That's a much better name. Will change.

detect_secrets/util/inject.py
domanchi: Never too late -- the original PR is much too big, and there's no way that it could have received detailed feedback. Thus, I still appreciate the comments, even though it is "after the fact".

I originally wrote the dependency injection framework to power the filter infrastructure, so there are assumptions built in reflective of this original use case. However, it seems that we are now using it for more things than just filters, so I've made the appropriate changes and handled these exceptions at the caller level.

detect_secrets/plugins/base.py
domanchi: This was my realization that this is never called with `secret=None`. This function is called [here](https://github.com/Yelp/detect-secrets/blob/pre-v1-launch/detect_secrets/main.py#L94), which only passes values in that are actual secrets.

Additionally, we never really have an explicit `None` value to pass into this function. `analyze_line` gives us a set of `PotentialSecret`s, so it doesn't cohere with the other functionality to obtain a `None` value to format.

calvinli (approved):
lgtm

PR #387, opened 2020-12-22
Reporting feature

pablosantiagolopez:
This feature allows to generate reports that include all the detected secrets details. This feature can be executed with the option `--report` of the `audit` command. The generated reports include the following fields for each detected secret:

- `class`: class associated to the secret. It can be `TRUE_POSITIVE`, `FALSE_POSITIVE` or `UNKNOWN`.
- `filename`: file where the secret was detected.
- `lines`: list of lines where the secret was detected. It will contain an entry for each occurrence of the secret in these file.
- `secret`: secret value in plaintext.
- `types`: list of secret types associated to the secret. Note that each secret can be detected by more than one plugin.

Moreover, this feature offer the following options:

- `--only-real`: only includes secrets with class `UNKNOWN` or `TRUE_POSITIVE` in the report.
- `--only-false`: only includes secrets with class `FALSE_POSITIVE` in the report.

Both options are optional, so if no option was selected, the report will include secrets of all classes.

Next, you can see an example of execution:
report

Thank you very much!
I hope that you like this feature


detect_secrets/audit/report.py
```diff
@@ -0,0 +1,137 @@
...
+                finding['lines'][detection.line_number] = get_line_content(filename, detection.line_number)
+                finding['types'] = [secret_type]
+                finding['class'] = secret_class.to_string()
+                secrets[identifier] = finding
+
+    output = []
+    for identifier in secrets:
+        output.append(secrets[identifier])
+
+    print_message(json.dumps(output, indent=4, sort_keys=True))
```
domanchi, line 88: Let's refactor this for easier testing.

This function should return a string, then the caller can call:

```python
print(json.dumps(audit.report.generate_report(args.filename[0], class_to_print)))
```

In doing so, we can test this function directly, without needing to mock stdout. See the pattern for `audit.analytics.calculate_statistics_for_baseline` as an example.

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,137 @@
...
+        else:
+            return SecretClassToPrint.FALSE_POSITIVE        
+
+
+def print_report(
+    baseline_file: str,
+    class_to_print: SecretClassToPrint = None
+) -> None:
+    baseline = json.load(codecs.open(baseline_file, encoding='utf-8'))
+    details = get_secrets_details_from_baseline(baseline)
```
domanchi, line 56: v1 actually makes this a lot easier. This looks more like:

```python
for filename, secret in get_baseline_from_file(baseline_file):
    ...
```

Check out `detect_secrets.audit.analytics` for examples.

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def get_prioritary(self, secret_class: str) -> Enum:
+        try:
+            to_compare = SecretClass[secret_class]
+        except:
+            return self
+        if to_compare.value < self.value:
+            return secret_class
+        else:
+            return self
```
domanchi, line 37: Let's use `detect_secrets.constants.VerifiedResult` instead of declaring another enum that pretty much does the same thing.

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,137 @@
...
+    filename: str,
+    line_number: int
+) -> str:
+    """
+    :returns: Line content from filename by line number.
+    """
+    content = codecs.open(filename, encoding='utf-8').read()
+    if not content:
+        return None
+    return content.splitlines()[line_number - 1]
```
domanchi, line 137: There's actually already prior art for this, in `detect_secrets.audit.common.get_raw_secret_from_file`. Let's reuse this instead, as there are optimizations and edge case handling already built into that.

It may need to be refactored a little, but that should (hopefully) be pretty straightforward. e.g.

```python
def get_raw_secret_from_file(secret: PotentialSecret) -> str:
    line = get_raw_secret_line_from_file(secret)
    ...


def get_raw_secret_line_from_file(secret: PotentialSecret) -> str:
    ...
```

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,137 @@
...
+                secrets[identifier]['class'] = secret_class.get_prioritary(secrets[identifier]['class']).to_string()
+            else:
+                finding = {}
+                finding['secret'] = detection.secret_value
+                finding['filename'] = filename
+                finding['lines'] = {}
+                finding['lines'][detection.line_number] = get_line_content(filename, detection.line_number)
+                finding['types'] = [secret_type]
+                finding['class'] = secret_class.to_string()
+                secrets[identifier] = finding
```
domanchi, line 82: Can we make this more like this style:

```python
secrets[identifier] = {
    'secret': detection.secret_value,
    'filename': filename,
    ...
}
```

detect_secrets/core/usage/audit.py
```diff
@@ -46,6 +47,34 @@ def _add_mode_parser(parser: argparse.ArgumentParser) -> None:
...
+        )
+    )
+
+    parser.add_argument(
+        '--only-false',
+        action='store_true',
+        help=(
+            'Only includes false positives in the report'
+        )
+    )
```
domanchi, line 73: Might be worth using https://docs.python.org/3/library/argparse.html#mutual-exclusion, since it doesn't really make sense for `--only-real --only-false` to be used together.

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,137 @@
...
+                    secrets[identifier]['types'].append(secret_type)
+                secrets[identifier]['class'] = secret_class.get_prioritary(secrets[identifier]['class']).to_string()
+            else:
+                finding = {}
+                finding['secret'] = detection.secret_value
+                finding['filename'] = filename
+                finding['lines'] = {}
+                finding['lines'][detection.line_number] = get_line_content(filename, detection.line_number)
+                finding['types'] = [secret_type]
+                finding['class'] = secret_class.to_string()
```
domanchi, line 81: Let's use `disposition` or `category` rather than `class`, as it more accurately represents this field.

domanchi (commented):
Overall, an interesting concept. Thank you for adding the screenshot -- that helps with understanding the intent of this feature much clearer.

There are a lot of developer improvements that will be released in v1: most of my comments revolve around this fact. You should check them out (and the (much more thorough) [docs/](https://github.com/Yelp/detect-secrets/tree/pre-v1-launch/docs)) and use them, rather than rewriting a lot of shared concepts.

Please also write tests if you intend to merge this. It also looks like linting may not have happened in this branch, so make sure you setup your development environment, and run the pre-commit hooks configured.

Also, I am curious -- my assumption when writing this tool is that the secret value is rarely used multiple times within the same file. This is written as if that may not be the case. Have you encountered real secrets declared multiple times within the same file?

pablosantiagolopez:
> Have you encountered real secrets declared multiple times within the same file?

Yes, this situation can be possible in large projects. With the information about the occurrences of a secret (line list) we can monitor the utilization of secrets for multiple systems and help the developers to retire all occurrences from the repository.

Thank you for all your comments! I'm working in them

detect_secrets/audit/report.py
pablosantiagolopez: Done!

detect_secrets/core/usage/audit.py
pablosantiagolopez: Done!

detect_secrets/audit/report.py
pablosantiagolopez: Done!

detect_secrets/audit/report.py
pablosantiagolopez: Done!

detect_secrets/audit/report.py
pablosantiagolopez: Done!

detect_secrets/audit/report.py
pablosantiagolopez: Done!

Note that we include in the `report.py` file the methods `get_prioritary_verified_result` and `get_verified_result_from_boolean` to make comparations and conversions related to the `VerifiedResult` enum. This methods were specifically implemented to the reporting feature, but if you prefer to include them in the `VerifiedResult` object, I can do it.

detect_secrets/audit/report.py
pablosantiagolopez: I have some doubts about use the function `get_raw_secret_file` in the reporting feature.

I have reviewed the function `get_raw_secret_from_file` and I have seen that returns only the secret value in plaintext and only his first occurrence. I think that could be very interesting the inclusion of all the occurrences of one secret in the report, so I had implemented the `get_potential_secrets` function to get all of them.

Moreover, I have seen that the function `get_raw_secret_from_file` is based in the field `line_number` of the `PotentialSecret` (in this case, it is obtained from the baseline file), so if the baseline file is not updated when the user generates the report, this line number could be invalid and this function will raise the `SecretNotFoundOnSpecifiedLineError` exception. 

I think that the approach that I implemented can get all the required details about the secrets to generate a complete report. 

What do you think? Maybe could we implement a similar function in the `audit.common.py` file that obtain all the ocurrences of a secret? I will wait to know your opinion before apply any modification. 

Thank you very much!

pablosantiagolopez: I'm also working in the report generation test, but maybe the resolution of this comment could have implications in the test implementation. I will push the test implementation after the resolution of this comment.

Thank you!

detect_secrets/audit/common.py
```diff
@@ -95,6 +95,43 @@ def get_raw_secret_from_file(
     raise SecretNotFoundOnSpecifiedLineError(secret.line_number)
 
 
+def get_all_secrets_from_file(
```
pablosantiagolopez, line 98: This method is very similar to `get_raw_secret_from_file` with the difference that it search all the ocurrences of one secret in one file with one plugin. It allows to offer a very complete report of all secrets found with detect-secrets. The method `get_all_secrets_from_file` has been added to the `detect_secrets/audit/common.py` file, so it can be reused by other features in the future.

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,83 @@
...
+    line_getter_factory: Callable[[str], 'LineGetter'] = open_file,
+) -> None:
+    secrets = {}
+    for filename, secret in get_baseline_from_file(baseline_file):
+        verified_result = get_verified_result_from_boolean(secret.is_secret)
+        if class_to_print is not None and SecretClassToPrint.from_class(verified_result) != class_to_print:  # noqa: E501
+            continue
+        detections = get_all_secrets_from_file(secret)
+        identifier = hashlib.sha512((secret.secret_hash + filename).encode('utf-8')).hexdigest()
+        line_getter = line_getter_factory(filename)
```
pablosantiagolopez, line 35: The method `generate_report` uses the new `LineGetter` class to manage all the file lines.

docs/audit.md
```diff
@@ -140,3 +140,126 @@ There are times you want to extract the raw secret values to run further analysi
 so with the `--raw` flag.
 
 TODO: Example when this feature is written up.
+
+## Report generation
```
pablosantiagolopez, line 144: The audit documentation has been updated to include the new reporting feature information. It also includes some examples.

README.md
```diff
@@ -357,29 +357,34 @@ const secret = "hunter2";
 
 ```bash
 $ detect-secrets audit --help
```
pablosantiagolopez, line 359: The audit help output has been updated to include the new report generation options.

pablosantiagolopez:
Hi @domanchi, I think that the new implementation of the reporting feature solves your previous comments. With this implementation, the report feature is simpler and supports the code reutilization. I hope that you like it, can you review it?

pablosantiagolopez:
Hi @domanchi, do you think this implementation is good enough to be merged? Please, tell me any improvement you want to apply. Thank you very much!

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,83 @@
...
+
+def get_verified_result_from_boolean(
+    is_secret: bool,
+) -> VerifiedResult:
+    if is_secret is None:
+        return VerifiedResult.UNVERIFIED
+    elif is_secret:
+        return VerifiedResult.VERIFIED_TRUE
+    else:
+        return VerifiedResult.VERIFIED_FALSE
```
domanchi, line 83: You probably can move this into the `VerifiedResult` class, like:

```
class VerifiedResult(Enum):
    @staticmethod
    def from_secret(secret: 'PotentialSecret') -> 'VerifiedResult':
        ...
```

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,83 @@
...
+
+def generate_report(
+    baseline_file: str,
+    class_to_print: SecretClassToPrint = None,
+    line_getter_factory: Callable[[str], 'LineGetter'] = open_file,
+) -> None:
+    secrets = {}
+    for filename, secret in get_baseline_from_file(baseline_file):
+        verified_result = get_verified_result_from_boolean(secret.is_secret)
+        if class_to_print is not None and SecretClassToPrint.from_class(verified_result) != class_to_print:  # noqa: E501
```
domanchi, line 31: FWIW, you can avoid the super long line by splitting this up. e.g.

```suggestion
        if class_to_print is not None and SecretClassToPrint.from_class(verified_result) != class_to_print:  # noqa: E501
        if (
            class_to_print is not None
            and SecretClassToPrint.from_class(verified_result) != class_to_print
        ):
```

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,83 @@
...
+            continue
+        detections = get_all_secrets_from_file(secret)
+        identifier = hashlib.sha512((secret.secret_hash + filename).encode('utf-8')).hexdigest()
```
domanchi, line 34: It looks like you just use this to deduplicate entries in your output dictionary. FWIW, you can also just use a `tuple` as a dictionary key:

```python
>>> data = {}
>>> data[(1,2,)] = 'a'
>>> print((1,2,) in data)
True
```

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,83 @@
...
+                        secret.type,
+                    ],
+                    'category': verified_result.name,
+                }
+
+    output = []
+    for identifier in secrets:
+        output.append(secrets[identifier])
+
+    return output
```
domanchi, line 62:
```suggestion
    return list(secrets.values())
```

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def get_prioritary_verified_result(
```
domanchi, line 65: Did you mean:

```suggestion
def get_prioritized_verified_result(
```

Sorry, I couldn't find what `prioritary` means.

tests/audit/report_test.py
```diff
@@ -0,0 +1,128 @@
...
+            line_number=2,
+            filename=second_file,
+        ),
+    }
+
+    f = tempfile.NamedTemporaryFile(delete=False)
+    baseline.save_to_file(secrets, f.name)
+    f.seek(0)
+    CREATED_FILES.append(f.name)
+    return f.name
```
domanchi, line 114: You can actually use the built-in context manager to handle this for you. Combined with `pytest.fixture`, this will actually make your code a lot easier to manage.

e.g.

```python
@pytest.fixture
def baseline_filename():
    # create your file contents
    ...

    # then, you can write your files
    with create_file_with_content(...) as first_file, \
            create_file_with_content(...) as second_file, \
            tempfile.NamedTemporaryFile() as baseline_file:
        ...
        baseline.save_to_file(...)
        yield baseline_file.name

# powered by
@contextmanager
def create_file_with_content(content):
    with tempfile.NamedTemporaryFile() as f:
        f.write(content)
        f.seek(0)

        yield f.name

# then
def test_generate_report(..., baseline_filename):
    output = generate_report(baseline_filename, class_to_print)
```

Check out https://github.com/Yelp/detect-secrets/blob/pre-v1-launch/tests/pre_commit_hook_test.py#L53-L55 for an example. This way, you don't have to handle global variables, nor worry about cleaning up after yourself for tests.

tests/audit/report_test.py
```diff
@@ -0,0 +1,128 @@
...
+import pytest
+
+from detect_secrets.audit.report import generate_report
+from detect_secrets.audit.report import SecretClassToPrint
+from detect_secrets.constants import VerifiedResult
+from detect_secrets.core import baseline
+from detect_secrets.core.secrets_collection import SecretsCollection
+from detect_secrets.plugins.basic_auth import BasicAuthDetector
+from detect_secrets.plugins.jwt import JwtTokenDetector
+from testing.factories import potential_secret_factory as original_potential_secret_factory
```
domanchi, line 15: Curious. Normally, I see this if we also redefine `potential_secret_factory` in the current file. However, I don't see this redefined -- is there any reason why you decided to rename it?

tests/audit/report_test.py
```diff
@@ -0,0 +1,128 @@
...
+
+def baseline_file():
+    # Create our own SecretsCollection manually, so that we have fine-tuned control.
+    url_format = 'http://username:{}@www.example.com/auth'
+    first_secret = 'value1'
+    second_secret = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ'  # noqa: E501
+    random_secret = ''.join(random.choice(string.ascii_letters) for _ in range(8))
+    file_content = 'url = ' + url_format.format(first_secret) \
+        + '\nexample = ' + url_format.format(random_secret) \
+        + '\nlink = ' + url_format.format(first_secret)
```
domanchi, line 57: One of my favorite, not-as-well known Python module is `textwrap`. Combined with Python's f-strings, this allows you to write inline strings with much more clarity. e.g.

```suggestion
    file_content = textwrap.dedent(f"""
        url = {url_format.format(first_secret)}
        example = {url_format.format(second_secret)}
        link = {url_format.format(first_secret)} 
    """)[1:]
```

tests/audit/report_test.py
```diff
@@ -0,0 +1,128 @@
...
+            filename=second_file,
+        ),
+        original_potential_secret_factory(
+            type=BasicAuthDetector.secret_type,
+            secret=random_secret,
+            is_secret=False,
+            line_number=2,
+            filename=second_file,
+        ),
+    }
```
domanchi, line 108: Interesting. Why not just do:

```python
secrets.scan_file(first_file)
secrets.scan_file(second_file)
```

if you need to label them later, you can do:

```python
labels = [True, False, None, ...]
for label, item in zip(labels, secrets):
    _, secret = item
    secret.is_secret = label
```

Probably more resilient (and more realistic test) than hard-coding, especially since you already went through the trouble of crafting secrets that would be caught by the plugins.

tests/audit/report_test.py
```diff
@@ -0,0 +1,128 @@
...
+    'class_to_print, expected_real, expected_false',
+    [
+        (None, 2, 2),
+        (SecretClassToPrint.REAL_SECRET, 2, 0),
+        (SecretClassToPrint.FALSE_POSITIVE, 0, 3),
+    ],
+)
+def test_generate_report(class_to_print, expected_real, expected_false):
+    filename = baseline_file()
+    output = generate_report(filename, class_to_print)
```
domanchi, line 31: Maybe also add a test for format? e.g.

```python
assert output[0] == {...}
```

I think you crafted good test cases to test different situations (e.g. if a secret is marked both as false positive, and true positive, secrets that appear on multiple lines) -- it would be good to assert them too.

On that note, should we also test for cases when `secret.is_secret == None`?

domanchi (commented):
This is looking a lot better!

Many of my comments are style-related; I think the core functionality for this report is sound. Tests look sane, with nice edge cases considered. I hope my code examples are illustrative and helpful, if Python isn't your primary development language.

Sorry it's took me so long to get to this! We've been addressing other issues internally, but finally able to get back to this effort and bring this tool to a new shiny future.

pablosantiagolopez:
Hi @domanchi, thank you very much for the comments and suggestions, I think that everything is done! 

PD: Contratulations for the new release of detect-secrets 💯

detect_secrets/audit/common.py
```diff
@@ -50,49 +50,70 @@ def get_raw_secret_from_file(
...
-        # consistent transformed file.
-        #
-        # NOTE: This is defensive coding. If we assume that this is only run on valid baselines,
-        # then the baseline wouldn't record secrets that were both found with and without an eager
-        # transformer, in the same file.
-        if is_first_time_opening_file and not line_getter.use_eager_transformers:
+        if secret.line_number:
+            try:
+                lines_to_scan = [line_getter.lines[secret.line_number - 1]]
+                line_numbers = [secret.line_number]
```
domanchi, line 86: You might have an off-by-one error here.

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,70 @@
...
+) -> None:
+    secrets = {}
+    for filename, secret in get_baseline_from_file(baseline_file):
+        verified_result = VerifiedResult.from_secret(secret)
+        if (
+            class_to_print is not None and
+            SecretClassToPrint.from_class(verified_result) != class_to_print
+        ):
+            continue
+        secret.line_number = 0
```
domanchi, line 35: Interesting. Can you add a comment as to why this is explicitly set?

detect_secrets/audit/report.py
```diff
@@ -0,0 +1,70 @@
...
+from .common import get_raw_secrets_from_file
+from .common import LineGetter
+from .common import open_file
+
+
+class SecretClassToPrint(Enum):
+    REAL_SECRET = 1
+    FALSE_POSITIVE = 2
+
+    def from_class(secret_class: VerifiedResult) -> Enum:
```
domanchi, line 15:
```suggestion
    def from_class(secret_class: VerifiedResult) -> 'SecretClassToPrint':
```

domanchi (approved):
Minor comments, but LGTM overall. Let's get this out in our next feature release (waiting for any other patch fixes that may need to be applied to `1.0.X` lol)

detect_secrets/audit/common.py
```diff
@@ -50,49 +50,70 @@ def get_raw_secret_from_file(
...
+        else:
+            lines_to_scan = line_getter.lines
+            line_numbers = range(len(lines_to_scan))
+
+        for line_number, line in zip(line_numbers, lines_to_scan):
+            identified_secrets = call_function_with_arguments(
+                plugin.analyze_line,
+                filename=secret.filename,
+                line=line,
+                line_number=line_number + 1,
```
pablosantiagolopez, line 98:
```
if secret.line_number:
             try:
                 lines_to_scan = [line_getter.lines[secret.line_number - 1]]
                 line_numbers = [secret.line_number]
```

I think that this code is correct, note that in the line 98 the line_number value is increased in one.

domanchi: Right, but `secret.line_number` is already 1-based. As such, when we obtain the `target_line`, we subtract it by 1 (to obtain from the `line_getter.lines` array).

So, if we add one to it again, I think we'll get an off-by-one error.

pablosantiagolopez: Okay, I get it. I have runned the tests and I think that it's fine.

domanchi:
@pablosantiagolopez, can you run `make test` and fix the surfaced issues? Looks like there's some `mypy` issues, and `detect-secrets` findings too.

detect_secrets/audit/common.py
```diff
@@ -44,7 +45,7 @@ def open_file(filename: str) -> 'LineGetter':
 def get_raw_secret_from_file(
     secret: PotentialSecret,
     line_getter_factory: Callable[[str], 'LineGetter'] = open_file,
-) -> str:
+) -> Any:
```
domanchi, line 48: Uh, is there a reason why this is `Any`? `str` looks sound to me.

syn-4ck: Hi @domanchi, we made this change because the mypy recommended it. I'm going to check it right now. When I have something else I will reply this comment.

syn-4ck: Oh, I think that I got it... We should change `str` to `Optional[str]'. I will test again, run the pre-commit hook and push this change. What do you think @domanchi?

syn-4ck:
Hi again @domanchi, with the last change (https://github.com/Yelp/detect-secrets/pull/387#discussion_r612527845) I think the mypy issues are fixed. Thanks for your feedback!

detect_secrets/audit/common.py
domanchi: Ohhh, I see the issue. It's because of https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/potential_secret.py#L66, so `mypy` (rightly) expects this to return the same type.

Yeah, `Optional[str]` works for me.

PR #393, opened 2021-01-15
Adding support for custom filters and disabling filters through CLI

domanchi:
## Summary

This introduces two new flags:

- `--filter`: allows the use of custom filters with the `detect-secrets` engine
- `--disable-filter`: allows users to disable built-in filters, if necessary

There's quite a bit of parsing logic to grok, but essentially, it allows for both custom modules to be used as custom filters, as well as local files. For example:

```
detect-secrets scan --filter detect_secrets_internal_filters.ml_model.is_secret
```

Might be YAGNI for now, but the intention is to accommodate new filters that may not be auto added to the default set (aka any new filters created after v1)

### DIFF
--- a/detect_secrets/core/usage/filters.py
+++ b/detect_secrets/core/usage/filters.py
@@ -6,6 +6,7 @@ from urllib.parse import urlparse
 
 from ... import filters
 from ...constants import VerifiedResult
+from ...core.log import log
 from ...exceptions import InvalidFile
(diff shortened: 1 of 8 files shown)


tanx16 (approved):
LGTM, could add a test case though

domanchi:
Merged to `pre-v1-launch`.

PR #394, opened 2021-01-25
Adds --slim flag for baseline merge conflict reduction

domanchi:
## Summary

Adds `--slim` flag during `scan` to create a baseline that reduces merge conflicts. Specifically, it removes line numbers and the `generated_at` field, as these are the main causes of observed merge conflicts.

A few additional points:
- When the pre-commit hook is used with a slimmed baseline, and it makes changes, it will also maintain the slimmed nature of the baseline.
- Since `audit` requires line numbers (to find the secret, without scanning the whole file), we raise an error if there's no line numbers to be found. While technically we can (automatically) re-scan the baseline upon auditing, I decided to leave that up to the user to update their baseline with line numbers (explicit local caching is a good thing).

Also, this PR also includes a fix for a pretty bad bug where the pre-commit hook would remove labels on secrets (e.g. `is_secret`). Good thing we found it before we released the new version!

## Reviewer's Note

It looks like this commit history includes one from https://github.com/Yelp/detect-secrets/pull/393, so I'll merge that first, then merge this one.

### DIFF
--- a/detect_secrets/core/secrets_collection.py
+++ b/detect_secrets/core/secrets_collection.py
@@ -188,7 +188,7 @@ class SecretsCollection:
         for filename in sorted(self.files):
             secrets = self[filename]
 
-            # TODO: Handle cases when line numbers are not supplied
+            # NOTE: If line numbers aren't supplied, they will default to 0.
             for secret in sorted(secrets, key=lambda x: (x.line_number, x.secret_hash, x.type)):
                 yield filename, secret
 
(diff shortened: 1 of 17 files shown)


detect_secrets/core/secrets_collection.py
```diff
@@ -176,7 +188,7 @@ def __iter__(self) -> Generator[Tuple[str, PotentialSecret], None, None]:
         for filename in sorted(self.files):
             secrets = self[filename]
 
-            # TODO: Handle cases when line numbers are not supplied
+            # NOTE: If line numbers aren't supplied, they will default to 0.
```
calvinli, line 191: would prefer `None` to be the line number value if not provided but not a big deal

calvinli (approved):

calvinli:
That's a ✅ on ada03e6 through e868384, to be clear :)

PR #395, opened 2021-01-27
Integrating with bump2version

domanchi:
This integrates with `bump2version`, and adds two helper deploy scripts:

- scripts/bump-version
- scripts/upload-to-pypi

After this, I will run:

```
$ scripts/bump-version major
$ scripts/upload-to-pypi
$ scripts/upload-to-pypi --prod
```

once v1 is *finally* merged!


docs/upgrades.md
```diff
@@ -57,12 +57,24 @@ assumption of the upgrade script will break.
 
 ### Changing the Version
 
-The version file is specified in `detect_secrets.__version__`. Just modify this value to the
-desired version, based on your change that is made.
+`detect-secrets` integrates with `bump2version` to make this process easier. Here's a handy
```
calvinli, line 60: I'd rather this work like all the other Yelp packages (merge to master, bump version number, tag the new version, let CI do the release)

If you want to write a convenience script for doing version bumping/changelog generation and tagging go ahead, but the actual release should be CI

scripts/bump-version
```diff
@@ -0,0 +1,106 @@
...
+        git tag "v$version"
+    else
+        echo "setting version to $version"
+    fi
+}
+
+function bumpVersion() {
+    local part="$1"
+    if [[ $DRY_RUN == 0 ]]; then
+        venv/bin/bump2version "$part"
```
calvinli, line 98: Using bump2version here seems kind of excessive, shouldn't changing the version in `detect_secrets/__version__.py` then tagging it be sufficient?

docs/upgrades.md
domanchi: I'd love to get a working CI again (after travis-ci.com stopped working for us), but that's going to have to wait for when we have the bandwidth to do infra improvements. Until then, scrappy solutions will need to suffice.

scripts/bump-version
domanchi: As with other helper scripts, they merely add nice-to-have features, that can always be done manually. `bump2version` includes standardized commit messages, and a workflow improvement. Furthermore, adding this "dependency" is a non-issue IMO, as it:

1. Does not inflate the size of the `detect-secrets` package, when others install it
2. Does not increase the time for the developer virtualenv to build (since it's only during deploy time)

PR #396, opened 2021-01-28
Keyword detector optimization

### LINKED ISSUE #148: Should 'secret_key' be added to the keyword plugin?

### LINKED ISSUE #270: Support the XML file format
In Yelp/detect-secrets-server#40 we got a report that a password in an XML tag called "password" was not detected. We can consider adding a regex to the keyword detector to detect values in such tags.

I'm a bit dubious as to whether this is a good idea since XML tags and inner values often spans multiple lines, but I figure I'd open up the issue for discussion.

pablosantiagolopez:
This pull request includes the following improvements to the `Keyword Detector` plugin:

- Specifically support for the secret detection in new file types: `INI`, `YAML`, `properties`, `C`, `C++`, `C#`, `bash`, `PowerShell`. 
- Secret detection by keyword combined with any other value. For example: the keyword `secret` will detect as secret the following lines `secret = value` and `my_special_secret_for_my_db = value`. Closes #148 
- Consideration of new situations where a secret could be hardcoded. For example: comparations between a variable and a hardcoded value `if (password == 'value')` or `if ('value' == password)`. 
- Improvement of the secret selection based on the regex group.
- New items in `DENYLIST` and `FALSE_POSITIVES` lists.

This pull request includes a new plugin called `Keyword XML Detector` to analyze all the situations where a secret could be hardcoded in a XML file. Closes #270 

The new plugins have the associated tests and all the implementation pass the pre-commit-hook established by the open source project.


detect_secrets/plugins/keyword.py
```diff
@@ -144,17 +172,20 @@
 }
 # Includes ], ', " as closing
 CLOSING = r'[]\'"]{0,2}'
-DENYLIST_REGEX = r'|'.join(DENYLIST)
+AFFIX_REGEX = r'[a-zA-Z0-9_-]*'
+DENYLIST_REGEX = r'(' + r'|'.join(DENYLIST) + r')({suffix})?'.format(suffix=AFFIX_REGEX)
+DENYLIST_REGEX_WITH_PREV = r'({prefix})('.format(prefix=AFFIX_REGEX) \
```
domanchi, line 177: Did you mean `DENYLIST_REGEX_WITH_PREFIX`?

detect_secrets/plugins/keyword.py
```diff
@@ -144,17 +172,20 @@
...
+    + r'|'.join(DENYLIST) + r')({suffix})'.format(suffix=AFFIX_REGEX)
 # Non-greedy match
 OPTIONAL_WHITESPACE = r'\s*?'
 OPTIONAL_NON_WHITESPACE = r'[^\s]{0,50}?'
 QUOTE = r'[\'"]'
-SECRET = r'[^\s]+'
+SECRET = r'""|\'\'|[^\'"\s]+'
```
domanchi, line 183: Why do we consider `""` and `''` a secret?

tests/plugins/keyword_test.py
```diff
@@ -329,3 +369,19 @@ def test_analyze_example_negatives(self, file_content):
             # Make it start with `<`, (and end with `>`) so it hits our false-positive check
             line=file_content.replace('m{', '<').replace('}', '>'),
         )
+
+    @pytest.mark.skip(
+        reason='TODO: false positive heuristics need to be migrated over to filters/*',
+    )
```
domanchi, line 287: Did you mean to skip this test? All filters have been migrated already.

domanchi (commented):
This PR introduces a variety of different, distinct changes, that I have comments on individually. As such, I figured it might be better to write them here, rather than at specific lines.

### Increasing "filetype" intelligence

All for this. Yay! I was originally worried about the compatibility with the config transformer, but it looks like the config transformer just attempts all files anyway, so we should be fine.

### Keyword Detector Improvements

TBH, this file continues to mutate, and becomes increasingly hard to understand -- so it takes me a while to understand your changes. I like your `FOLLOWED_BY_COMPARATION_QUOTES_REQUIRED_REGEX` and `FOLLOWED_BY_REV_COMPARATION_QUOTES_REQUIRED_REGEX`, but I'm uncertain of the implications of modifying the core `DENYLIST_REGEX` and `SECRET` regexes.

It *should* be the case that secret values don't have quotes in them: but I need to think more deeply about whether we can assert this.

### KeywordXML

Ok, this absolutely needs to change. Along the same lines of why the KeywordDetector is in need for some serious love (to increase code flexibility), I'm against a direct copy-paste of the `KeywordDetector` to make a version for XML files.

Have you considered writing an XML [transformer](https://github.com/Yelp/detect-secrets/blob/pre-v1-launch/docs/design.md#transformers) instead? This will allow it to take advantage of all the other features that v1 provides. For prior art, check out [YAMLTransformer](https://github.com/Yelp/detect-secrets/blob/pre-v1-launch/detect_secrets/transformers/yaml.py#L25).

detect_secrets/plugins/keyword.py
```diff
@@ -205,49 +235,103 @@
...
+        denylist=DENYLIST_REGEX_WITH_PREV,
+        quote=QUOTE,
+        whitespace=OPTIONAL_WHITESPACE,
+        secret=SECRET,
+    ),
+)
 FOLLOWED_BY_QUOTES_AND_SEMICOLON_REGEX = re.compile(
     # e.g. private_key "something";
-    r'({denylist}){nonWhitespace}{whitespace}({quote})({secret})(\2);'.format(
+    r'({denylist}){nonWhitespace}{whitespace}({quote})({secret})({quote});'.format(
```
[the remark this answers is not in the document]
pablosantiagolopez: This regex was updated to be sure that the last quote won't be included as a secret in the audit code snippet. The original `SECRET` regex allows the `'` and `"` characters, so the last quote will be included in the secret value. I have modified the `SECRET` regex to match only with empty values (`""` or `''`, commented below) or whatever value without `'`, `"` and whitespaces. With this changes, I have included the last quote as a new item in this regex to make it more specific.

detect_secrets/plugins/keyword.py
pablosantiagolopez: When you analyze an other person's repository, you don't known the context of each secret, so I think that it is important to detect secrets in the mosts situations as possible and then mark them as false positive if it is needed. In some environments, as the local environment, developers usually configure empty passwords for some resources, so I think that detect-secrets should detect it.

tests/plugins/keyword_test.py
pablosantiagolopez: I'm sorry, when I have implemented this feature, the filters haven't been migrated already. I will commit it comming soon.

detect_secrets/plugins/keyword.py
pablosantiagolopez: Moreover, there are a lot of common technologies that use blank passwords by default. [Here](https://github.com/danielmiessler/SecLists/blob/master/Passwords/Default-Credentials/default-passwords.csv) you can see some.

pablosantiagolopez:
> I'm uncertain of the implications of modifying the core `DENYLIST_REGEX` and `SECRET` regexes.

All the modifications in this variables are based on our experience analyzing multiples repositories. With this modifications we note an improvement in the results obtained with detect-secrets.

pablosantiagolopez:
> Ok, this absolutely needs to change

Okey, I will work in your proposal comming soon. If you like the other changes included in this Pull Request, we can remove the `KeywordXMLDetector` code from it and contributing this feature later.

detect_secrets/plugins/keyword.py
pablosantiagolopez: Done!

tests/plugins/keyword_test.py
pablosantiagolopez: Hi @domanchi, I have reviewed the default filters of detect-secrets:

```JSON
"filters_used": [
    {
      "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
    },
    {
      "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
      "min_level": 2
    },
    {
      "path": "detect_secrets.filters.heuristic.is_likely_id_string"
    },
    {
      "path": "detect_secrets.filters.heuristic.is_potential_uuid"
    },
    {
      "path": "detect_secrets.filters.heuristic.is_sequential_string"
    }
  ],
```
In the default filters, we haven't found any adequate to our test's needs. I think that the most useful filter in this case, can be `heuristic.is_templated_secret`. We haven't found any references to this filter in the [documentation](https://github.com/Yelp/detect-secrets/blob/pre-v1-launch/docs/filters.md#user-content-built-in-filters) and we think that it should be established by default. Are this tasks in your roadmap?

domanchi: Good catch. I must have missed this during my migration effort. I'll cut a ticket for this.

domanchi: https://github.com/Yelp/detect-secrets/pull/405

pablosantiagolopez:
Hi @domanchi, I close this Pull Request. @syn-4ck and I will contribute all this features adapted to the new version of detect-secrets comming soon. Here is our first contribution #418 related to this Pull Request, I think that it will be easier to discuss every change in keyword plugin using smaller Pull Requests. Thank you very much!

Issue #406, opened 2021-02-17
Security Slack Channel should be a parameter


sfc-gh-nsfard:
one of the mitigation suggestions that are printed when detect-secrets fails a scan states:

```
- For information about putting your secrets in a safer place,
    please ask in #security
```

This is a bit confusing for developers since we don't have a #security channel and we wouldn't use that channel for this type of question anyway. I think this should be a custom value that's handled as a commandline parameter (or via a config file or an environment variable)

I notice it's currently being hardcoded here: https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/pre_commit_hook.py#L210

I also notice there that `DETECT_SECRETS_SECURITY_TEAM` is already an environment variable we can set. I think it'd be pretty easy to just create another env var called `DETECT_SECRETS_SLACK_CHANNEL` with `#security` being the default value. What do you think?

domanchi:
Pull requests welcome.

sfc-gh-nsfard:
@domanchi PR here: https://github.com/Yelp/detect-secrets/pull/413

BlueCanary-DM:
You can also trigger a slackbot via email.

domanchi:
From the linked PR, it looks like this functionality is already supported. i.e. you can do

```
DETECT_SECRETS_SECURITY_TEAM='in #blah' detect-secrets scan ...
```

which will result in:

```
For information about putting your secrets in a safer place, please ask in #blah.
```

PR #413, opened 2021-02-26
security slack channel as a env var

sfc-gh-nsfard:
Addressing https://github.com/Yelp/detect-secrets/issues/406

### DIFF
--- a/detect_secrets/pre_commit_hook.py
+++ b/detect_secrets/pre_commit_hook.py
@@ -119,6 +119,8 @@ def should_update_baseline(
 
 
 def pretty_print_diagnostics(secrets: SecretsCollection, width: int = 80) -> None:
+    SECURITY_SLACK_CHANNEL = os.environ.get('DETECT_SECRETS_SECURITY_SLACK_CHANNEL') or '#security'
+
     # Header
     print(
         textwrap.fill(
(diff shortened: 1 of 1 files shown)


detect_secrets/pre_commit_hook.py
```diff
@@ -144,7 +146,7 @@ def pretty_print_diagnostics(secrets: SecretsCollection, width: int = 80) -> Non
     )
     for suggestion in [
         'For information about putting your secrets in a safer place, '
-        f'please ask {os.environ.get("DETECT_SECRETS_SECURITY_TEAM", "in #security")}',
+        f'please ask {os.environ.get("DETECT_SECRETS_SECURITY_TEAM")} in {SECURITY_SLACK_CHANNEL}',
```
domanchi, line 149: The bad news is that this isn't going to work.

```python
>>> print(f'team: {os.environ.get("blah")}')
team: None
```

The good news is that it looks like this PR isn't needed? That is, you can just do:

```bash
DETECT_SECRETS_SECURITY_TEAM='in #blah' detect-secrets scan ...
```

and this code will already do what you're looking for it to do:

```
For information about putting your secrets in a safer place, please ask in #blah.
```

domanchi (changes requested):

domanchi:
I'm closing this PR, given the conversation thread above.

@sfc-gh-nsfard : if you think the proposed solution doesn't address your needs, feel free to reopen.

PR #417, opened 2021-03-04
adding is_lock_file filter

domanchi:
## Summary

Lock files typically have hashes in them, and this triggers a lot of false positives for our high entropy string plugins. We've seen success in just ignoring them, so let's formalize it.

### DIFF
--- a/detect_secrets/filters/heuristic.py
+++ b/detect_secrets/filters/heuristic.py
@@ -165,3 +165,12 @@ def is_indirect_reference(secret: str) -> bool:
             continue
 
     return output
+
+
+def is_lock_file(filename: str) -> bool:
+    return os.path.basename(filename) in {
+        'Brewfile.lock.json',
+        'composer.lock',
+        'package.lock',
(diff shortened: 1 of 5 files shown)


detect_secrets/filters/heuristic.py
```diff
@@ -165,3 +165,12 @@ def is_indirect_reference(secret: str) -> bool:
...
 
     return output
+
+
+def is_lock_file(filename: str) -> bool:
+    return os.path.basename(filename) in {
+        'Brewfile.lock.json',
+        'composer.lock',
+        'package.lock',
+        'yarn.lock',
```
calvinli, line 175: I've also seen `Podfile.lock`, `Cartfile.resolved`, and `Package.resolved`

is `package.lock` a typo for `package-lock.json`?

domanchi: TIL about those other entries. I've updated this, and added tests too.

calvinli (approved):

PR #418, opened 2021-03-04
Keyword regexes improvement

pablosantiagolopez:
This Pull Request solves some issues detected in #414 and includes some improvements described in #396 to the latest version of detect-secrets.

This code includes:

- Support to secret detection in comparisons. You can see it in `FOLLOWED_BY_EQUAL_SIGNS_REGEX` and `FOLLOWED_BY_EQUAL_SIGNS_QUOTES_REQUIRED_REGEX` regexes. For example: 
```JAVASCRIPT
if (password == "value") {}
else if (password === "value") {}
else if (password != "value") {}
else if (password !== "value") {}
```
- Support to secret detection in reverse comparisons. You can see it in the `PRECEDED_BY_EQUAL_COMPARISON_SIGNS_QUOTES_REQUIRED_REGEX` regex. For example:
```JAVASCRIPT
if ("value" == password) {}
else if ("value" === password) {}
else if ("value" != password) {}
else if ("value" !== password) {}
```
- Corrections in the `SECRET` regex to avoid some problems detected in #414 and reduce the false positive percentage. The regex can be decomposed in the following sections:

   -  `[^\r\n]*`: this section match with every character except line breaks. This allows to find secrets that starts with symbols or alphanumeric characters.
   - `[a-zA-Z0-9]+`: this section match only with alphanumeric characters, and at least one is required. This allows to reduce the false positives number.
   - `[^\r\n]*`:  this section match with every character except line breaks. This allows to find secrets with symbols at the end. 
   - `[^\r\n,\'"]`: this section match with the last secret character that can be everything except line breaks, comma or quotes. This allows to reduce the false positives number and to prevent errors in the code snippet highlighting. Next you can see an example:  
last_comma
The following image shows a test with this new regex:

regex_example

Of course, `keyword_test.py` has been updated to check all this changes.

PR #420, opened 2021-03-08
Secret regex bugfix & suffix and prefix support

pablosantiagolopez:
## Secret regex bugfix

The performance bug described by @domanchi in #418 has been fixed. The bug happens when the PRECEDED_BY_EQUAL_COMPARISON_SIGNS_QUOTES_REQUIRED_REGEX regex is checked in files with long lines that includes many quotes. The regex engine must check a lot of combinations to determine the line section that matches the secret regex. To solve this issue the quote before the secret (if it exists) has been excluded in the secret regex, improving the regex engine performance.

The secret regex has been replaced by **[^\v\\g{quote_group}]*\w+[^\v\\g{quote_group}]*[^\v,\'"`]**. The main changes in the regex are the following:

- Replace of `\r\n` by `\v` because they are equivalent. This change simplifies the regex.
- Exclusion of `\g{quote_group}` in all the regexes. `quote_group` is the regex group number of the quote before the secret value (if it exists) and it is configured in each specific regex. This change prevents the bug to occur and reduce the false positives number.
- Replace of `[a-zA-Z0-9]+` by `\w+`because they are equivalent. This change simplifies the regex.

Next you can see the result of the time command:

time

## Suffix and prefix support

The original plugin KeywordDetector doesn't detect secrets in lines like:

```JAVA
String my_password_secure = "value"

if ("value" == my_password_secure) {
```

This problem happens because the plugin searches for variable names that ends in any keyword, but it doesn't take into account variable names that contains keywords in the middle. So we add the following lines to allow the presence of words arround the keywords:

```PYTHON
AFFIX_REGEX = r'\w*'
DENYLIST_REGEX = r'|'.join(DENYLIST)
# Support for suffix after keyword i.e. password_secure = "value"
DENYLIST_REGEX = r'({denylist}){suffix}'.format(
    denylist=DENYLIST_REGEX,
    suffix=AFFIX_REGEX,
)
# Support for prefix and suffix with keyword, needed for reverse comparisons
# i.e. if ("value" == my_password_secure) {}
DENYLIST_REGEX_WITH_PREFIX = r'{prefix}{denylist}'.format(
    prefix=AFFIX_REGEX,
    denylist=DENYLIST_REGEX,
)
```

We add new test cases with suffix and prefix. Thank you!

### DIFF
--- a/detect_secrets/filters/heuristic.py
+++ b/detect_secrets/filters/heuristic.py
@@ -165,6 +165,11 @@ def is_indirect_reference(line: str) -> bool:
 
         secret = request.headers['apikey']
     """
+    return bool(_get_indirect_reference_regex().search(line))
+
+
+@lru_cache(maxsize=1)
+def _get_indirect_reference_regex() -> Pattern:
     # Regex details:
     #   ([^\v=!:]*)     ->  Something before the assignment or comparison
     #   \s*             ->  Some optional whitespaces
@@ -176,8 +181,7 @@ def is_indirect_reference(line: str) -> bool:
     #       [^\v]*      ->  Something except line breaks
     #       [\]\)]      ->  End of indirect reference: ] or )
     #   )
-    regex = re.compile(r'([^\v=!:]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
-    return bool(regex.search(line))
+    return re.compile(r'([^\v=!:]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
 
 
 def is_lock_file(filename: str) -> bool:
--- a/detect_secrets/filters/heuristic.py
+++ b/detect_secrets/filters/heuristic.py
@@ -155,7 +155,7 @@ def is_prefixed_with_dollar_sign(secret: str) -> bool:
     return secret[0] == '$'
 
 
-def is_indirect_reference(secret: str) -> bool:
+def is_indirect_reference(line: str) -> bool:
     """
     Filters secrets that take the form of:
 
@@ -165,19 +165,19 @@ def is_indirect_reference(secret: str) -> bool:
 
         secret = request.headers['apikey']
     """
-    output = False
-    for start, end in (
-        list('()'),
-        list('[]'),
-    ):
-        try:
-            output = secret.index(start) < secret.index(end)
-            if output:
-                return output
-        except ValueError:
-            continue
-
-    return output
+    # Regex details:
+    #   ([^\v=!:]*)     ->  Something before the assignment or comparison
+    #   \s*             ->  Some optional whitespaces
+    #   (:=?|[!=]{1,3}) ->  Assignment or comparison: :=, =, ==, ===, !=, !==
+    #   \s*             ->  Some optional whitespaces
+    #   (
+    #       [\w.-]+     ->  Some alphanumeric character, dot or -
+    #       [\[\(]      ->  Start of indirect reference: [ or (
+    #       [^\v]*      ->  Something except line breaks
+    #       [\]\)]      ->  End of indirect reference: ] or )
+    #   )
+    regex = re.compile(r'([^\v=!:]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
+    return bool(regex.search(line))
 
 
 def is_lock_file(filename: str) -> bool:
--- a/detect_secrets/plugins/keyword.py
+++ b/detect_secrets/plugins/keyword.py
@@ -58,23 +58,23 @@ OPTIONAL_WHITESPACE = r'\s*'
 OPTIONAL_NON_WHITESPACE = r'[^\s]{0,50}?'
 QUOTE = r'[\'"`]'
 # Secret regex details:
-#    [^\v\\g<{quote_group}>]*   -> this section match with every character except line breaks
-#                                  and the previous quote if exists. This allows to find 
-#                                  secrets that starts with symbols or alphanumeric characters.
+#    [^\v\g{quote_group}]*  -> this section match with every character except line breaks
+#                              and the previous quote if exists. This allows to find 
+#                              secrets that starts with symbols or alphanumeric characters.
 #
-#    \w+                        -> this section match only with words (letters, numbers or _ 
-#                                  are allowed), and at least one character is required. This 
-#                                  allows to reduce the false positives number.
+#    \w+                    -> this section match only with words (letters, numbers or _ 
+#                              are allowed), and at least one character is required. This 
+#                              allows to reduce the false positives number.
 #
-#    [^\v\\g<{quote_group}>]*   -> this section match with every character except line breaks
-#                                  and the previous quote if exists. This allows to find secrets 
-#                                  with symbols at the end.
+#    [^\v\g{quote_group}]*  -> this section match with every character except line breaks
+#                              and the previous quote if exists. This allows to find secrets 
+#                              with symbols at the end.
 #
-#    [^\v,\'"`]                 -> this section match with the last secret character that can be
-#                                  everything except line breaks, comma, backticks or quotes. This
-#                                  allows to reduce the false positives number and to prevent
-#                                  errors in the code snippet highlighting.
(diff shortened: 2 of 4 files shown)


tests/plugins/keyword_test.py
```diff
@@ -51,6 +55,7 @@
     ('if ("{}" === private_key) {{'.format(COMMON_SECRET), COMMON_SECRET),
     ('if ("{}" != secret) {{'.format(COMMON_SECRET), COMMON_SECRET),
     ('if ("{}" !== password) {{'.format(COMMON_SECRET), COMMON_SECRET),
+    ('if ("{}" !== my_password_sec) {{'.format(COMMON_SECRET), COMMON_SECRET),    # Preffix/suffix
```
domanchi, line 58:
```suggestion
    ('if ("{}" !== my_password_sec) {{'.format(COMMON_SECRET), COMMON_SECRET),    # Prefix/suffix
```

detect_secrets/plugins/keyword.py
```diff
@@ -52,69 +52,85 @@
 )
 # Includes ], ', " as closing
 CLOSING = r'[]\'"]{0,2}'
+AFFIX_REGEX = r'\w*'
 DENYLIST_REGEX = r'|'.join(DENYLIST)
+# Support for suffix after keyword i.e. password_secure = "value"
+DENYLIST_REGEX = r'({denylist}){suffix}'.format(
```
domanchi, line 58: I notice that this adds "suffix support" by default to all capturing regexes; however, "prefix support" is only for `PRECEDED_BY_EQUAL_COMPARISON_SIGNS_QUOTES_REQUIRED_REGEX`. What's the reason for this?

detect_secrets/plugins/keyword.py
```diff
@@ -52,69 +52,85 @@
...
+#
+#    [^\v\g{quote_group}]*  -> this section match with every character except line breaks
+#                              and the previous quote if exists. This allows to find secrets
+#                              with symbols at the end.
+#
+#    [^\v,\'"`]             -> this section match with the last secret character that can be
+#                              everything except line breaks, comma, backticks or quotes. This
+#                              allows to reduce the false positives number and to prevent
+#                              errors in the code snippet highlighting.
+SECRET = r'[^\v\\g{quote_group}]*\w+[^\v\\g{quote_group}]*[^\v,\'"`]'
```
domanchi, line 89: Interesting.

If I'm understanding this right, technically speaking, we can replace the "quote group" with just quotes in general (i.e. `QUOTE`), right? This assumes that secrets don't have quote like characters in them.

detect_secrets/plugins/keyword.py
pablosantiagolopez: This is because in the other regex the secret value is always after the keyword, so the prefix is not important to us. For example: in the line `my_password = "value"` the plugin will detect `password = "value"`, even if `my_` is not included in the regex.

In the `PRECEDE_BY_EQUAL_COMPARISON_SIGNS_QUOTES_REQUIRED_REGEX` the secret value is always before the keyword, so the prefix is needed to detect the secret. For example: in the line `if ("value" == my_password) {` if the  `my_` isn't included in the line, the secret won't be detected.

detect_secrets/plugins/keyword.py
pablosantiagolopez: Yes, we are trying to exclude the text that matches `QUOTE` but it isn't working as we expect. So we replace `\\g{quote_group}` by `\'"` in the `SECRET` regex. We think that this quotes won't be included in secrets because it is unusual in the most programming languages.

pablosantiagolopez:
Hi @domanchi, we have finished all the tests and changes related to this Pull Request. I think that everything is okay. What do you think?

domanchi (approved):

domanchi (changes requested):
@pablosantiagolopez , did you run `make test` to make sure your change is safe? I'm getting test failures on your branch.

pablosantiagolopez:
Moreover of the @syn-4ck comment, if we add more complexity to the `SECRET` regex we can introduce new bugs or performance issues like the comment in #418. I think that it is large enought yet.

In the case of the filters, I think that it is interesting to maintain both filters `gibberish.should_exclude_secret` and `heuristic.is_indirect_reference` independently because they can be disabled separately.

domanchi:
I agree that those filters should be different. For example, the `heuristic.is_indirect_reference` filters out `secret = get_secret_value()` (good), but with the latest change to not include quotes in the findings of `KeywordDetector`, I'm proposing that `heuristic.is_indirect_reference` just becomes (arguably) *less* useful.

In other words, I'm struggling to understand why we want to add additional complexity to the `SECRET` regex, for the mere purposes of filtering it out later. If the additional complexity actually catches secrets that we may want to care about (e.g. secrets with whitespace, suffixed variable names), this makes sense. Otherwise, [plugins should find stuff](https://github.com/Yelp/detect-secrets/blob/master/docs/design.md#plugins) and [filters should exclude them](https://github.com/Yelp/detect-secrets/blob/master/docs/design.md#filters), no?

pablosantiagolopez:
Okay, I think that we didn't explain it good enough. If we don't add this additional complexity to the `SECRET` regex, in the line `secret = request.headers['apikey']` the plugin will get the value `request.headers[` as a secret and the filter `heuristic.is_indirect_reference` (or the other filters) won't be able to detect it as a false positive. So we modify the `SECRET` regex to detect the full "secret" value in this cases: `request.headers['apikey']` and the filter `heuristic.is_indirect_reference` can filter this value because it includes the characters `[` and `]` in that order. However, in the value `request.headers[` it can't do that because it only has one `[` and the "secret" will be included as a output of detect-secrets scan.

In a previous comment, you commented that we can modify the `SECRET` regex to don't detect any secrets in lines like `secret = request.headers['apikey']`. This makes sense and I think that it could be the best approach, but I also think that it will be very difficult to implement and the `SECRET` regex will be too large and complex, increasing the matching combinations in large lines and maybe reducing the detect-secrets performance. Moreover, I think that it could introduce some bug in the tool, because in the regex world, there aren't many differences between this line `secret = request.headers['apikey']` and this other `secret = <real secret value without quotes>`. 

The solution described in the first paragraph is implemented in this Pull Request right now. If you think that the solution described in the second paragraph is better (I understood that it is your proposal) we can try to implement it and comment in this Pull Request the results obtained. Please tell us your opinion to work in this problem.

I hope that I have explained the problem well this time, I know that it is a very specific situation and it's hard to think in all the other situations that can be affected by this changes. Thank you very much for all your comments, they are very helpful!

pablosantiagolopez:
Well, we implement a new approach to solve the problem without change the `SECRET` regex. We change the filter `heuristic.is_indirect_reference` to check lines instead of secrets, so it can detect indirect references before scan the line with each plugin. 
I think that it is the best solution to this problem because the keyword plugin isn't affected at all and the filter works fine. What do you think?
Thank you!

detect_secrets/filters/heuristic.py
```diff
@@ -165,19 +165,19 @@ def is_indirect_reference(secret: str) -> bool:
...
+    #   \s*             ->  Some optional whitespaces
+    #   (:=?|[!=]{1,3}) ->  Assignment or comparison: :=, =, ==, ===, !=, !==
+    #   \s*             ->  Some optional whitespaces
+    #   (
+    #       [\w.-]+     ->  Some alphanumeric character, dot or -
+    #       [\[\(]      ->  Start of indirect reference: [ or (
+    #       [^\v]*      ->  Something except line breaks
+    #       [\]\)]      ->  End of indirect reference: ] or )
+    #   )
+    regex = re.compile(r'([^\v=!:]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
```
domanchi, line 179: Let's cache this regex, as described in [Best Practices for Writing Filters](https://github.com/Yelp/detect-secrets/blob/master/docs/filters.md#best-practices). An example is found in that documentation too (search `Example #2`)

domanchi (commented):
This approach makes a lot more sense. Good thinking!

detect_secrets/filters/heuristic.py
pablosantiagolopez: Okay, perfect

pablosantiagolopez:
> This approach makes a lot more sense. Good thinking!

Thank you very much for your comments and ideas!

domanchi (approved):

PR #422, opened 2021-03-10
Preserve binary values for YAMLTransformer

### LINKED ISSUE #421: Bug auditing test_data/config.yaml with binary values
We found the following bug scanning detect-secrets project when the file `test_data/config.yaml` is audited:

bug_audit

We tried to solve this bug in #414, and we get it in the most cases, but this file includes the following line: `low_entropy_binary_secret: !!binary MjNjcnh1IDJieXJpdXYyeXJpaTJidnl1MnI4OXkyb3UwMg==`. We try to parse this file as YAML from python and we get the following behaviour:

bug_config_yaml

As can you see, the secret value changes when the file is parsed as YAML, so I think that this bug can be introduced by the YAML transformer. I'm not sure because I don't know how this transformer works exactly. 

I also think that this bug is due to the specific value included in this file and it isn't a generic problem. I don't know how to fix it, so I open this issue to report it.

domanchi:
Fixes #421.

## Testing Done

```bash
$ detect-secrets scan test_data/config.yaml > blah.baseline
$ detect-secrets audit blah.baseline
```


calvinli (approved):
sgtm

PR #423, opened 2021-03-11
Filter of swagger files

pablosantiagolopez:
This Pull Request introduce the `heuristic.is_swagger_file` filter that excludes files like `swagger-ui.html` or files with the keyword `swagger` in its path. This files should be excluded because in the most cases only contain example values that are marked as false positives.


detect_secrets/filters/heuristic.py
```diff
@@ -191,3 +191,11 @@ def is_lock_file(filename: str) -> bool:
         'Podfile.lock',
         'yarn.lock',
     }
+
+
+def is_swagger_file(filename: str) -> bool:
+    """
+    Filters swagger files and paths, like swagger-ui.html or /swagger/.
+    """
+    regex = re.compile(r'.*swagger.*')
```
domanchi, line 200: Let's cache this regex, as described in [Best Practices for Writing Filters](https://github.com/Yelp/detect-secrets/blob/master/docs/filters.md#best-practices). An example is found in that documentation too (search `Example #2`)

pablosantiagolopez: Okay, It's done!

domanchi (approved):

detect_secrets/filters/heuristic.py
```diff
@@ -191,3 +191,15 @@ def is_lock_file(filename: str) -> bool:
...
+def is_swagger_file(filename: str) -> bool:
+    """
+    Filters swagger files and paths, like swagger-ui.html or /swagger/.
+    """
+    return bool(_get_swagger_regex().search(filename))
+
+
+@lru_cache(maxsize=1)
+def _get_swagger_regex() -> Pattern:
+    return re.compile(r'.*swagger.*')
```
domanchi, line 205: I think this is probably too aggressive? Looks like it introduces a huge slowdown again with the `rfc.model`.

domanchi (changes requested):

detect_secrets/filters/heuristic.py
pablosantiagolopez: No, I think there is no bug here. The branch `filter/is-swagger-file` has been created from the detect-secrets master before you merge #420. Now #420 is merged, and the bug detected in the `rfc.model` scanning is solved. I have updated the Pull Request with the latest changes in master, so it should work fine now.
Please, check it again. Thank you very much!

domanchi: ohh, gotcha. You're probably right.

domanchi (approved):

domanchi:
This has been merged: https://github.com/Yelp/detect-secrets/commit/22326d5bbbb4320a9296729be3c5e4d5c48013bc

PR #424, opened 2021-03-11
Filter of i18n files

pablosantiagolopez:
This Pull Request introduce the `heuristic.is_i18n_file` filter that excludes some internationalization files from the plugins scan. This files should be excluded because only content text that will be printed in the frontends, so all the findings in these files will be false positives.


domanchi:
This is a great idea, but I think this should exist as a [custom filter](https://github.com/Yelp/detect-secrets/blob/master/docs/filters.md#using-your-own-filters) instead, given that there's no standard layout for i18n files.

If I'm wrong, please let me know.

PR #427, opened 2021-03-11
Filter compiled folders

pablosantiagolopez:
This Pull Request introduce the `heuristic.is_compiled_file` filter that excludes compiled folders like:

- `dist/`: generated after python setup
- `build/`: generated after python setup
- `/__pycache__/`: generated after python execution
- `/.*egg-info/`: generated after python setup
- `/node_modules/`: generated by npm install (Node JS)
- `/target/`: generated by mvn install (Apache Maven)

### DIFF
--- a/detect_secrets/filters/heuristic.py
+++ b/detect_secrets/filters/heuristic.py
@@ -198,14 +198,13 @@ def is_compiled_file(filename: str) -> bool:
         'package.json'
     }:
         return True
-    regexes = [re.compile(r) for r in [
-            r'^dist{}.*'.format(os.path.sep),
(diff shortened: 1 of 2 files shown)


domanchi:
@pablosantiagolopez : `detect-secrets` already runs on git tracked files by default already. In our experience, it does not trigger on these types of files since they are not tracked by git.

Given this, in your experience, is this still needed?

pablosantiagolopez:
Oh, you are right! I'm sorry, I didn't notice that because I usually use the `--all-files` option. This code will stay in this branch of my fork, so it can be merged if needed. In this moment I think that this Pull Request is unuseful so I close it.
Thank you!

PR #432, opened 2021-03-17
don't override the root logger

domanchi:
## Summary

With this change, the `detect-secrets` logger will not use the root logger. I've tested this by running:

```
detect-secrets -vv scan
```

and the logs seem to still work. This way, when we modify the debug level, it won't cause all other packages (that use the root logger) to show debug logs too.


calvinli (approved):
lgtm

PR #441, opened 2021-04-01
feature: adding multiprocessing support

domanchi:
I never knew parallel processing could be so easy. ❤️   python

## Summary

This increases the efficiency of scans by throwing some parallel processing to it.

Since scanning files are independent of each other, it seemed like a prime candidate for this. I considered parallelizing line scanning, but then realized that this may not be such a good idea, since lines are somewhat dependent on each other (e.g. context-based filters).

## Testing Done

I ran this on our internal monolith, and these were the stats I got:

```bash
# Some stats on the repository itself
$ git ls-files | wc -l
35519
$ cloc $(git ls-files)
...
--------------------------------------------------------------------------------
Language                      files          blank        comment           code
--------------------------------------------------------------------------------
...
--------------------------------------------------------------------------------
SUM:                          22948         660907         422368        3680522
--------------------------------------------------------------------------------
$ lscpu
...
CPU(s):              16

# Running with single process
$ time detect-secrets scan > /tmp/single.baseline
real    23m41.917s
user    23m20.518s
sys     0m3.527s

# Running with multiprocessor
$ time detect-secrets scan > /tmp/mp.baseline
real    6m4.887s
user    37m4.611s
sys     0m8.304s

# Comparing results
$ diff /tmp/single.baseline /tmp/mp.baseline
4961c4961
<   "generated_at": "2021-04-01T14:42:50Z"
---
>   "generated_at": "2021-04-01T15:05:30Z"
```

That's like a 74% optimization. Heck yeah!

### DIFF
--- a/detect_secrets/core/secrets_collection.py
+++ b/detect_secrets/core/secrets_collection.py
@@ -1,3 +1,4 @@
+import multiprocessing as mp
 import os
 from collections import defaultdict
 from typing import Any
@@ -44,6 +45,16 @@ class SecretsCollection:
     @property
     def files(self) -> Set[str]:
         return set(self.data.keys())
+
+    def scan_files(self, *filenames: str, num_processors: int = mp.cpu_count()) -> None:
+        """Just like scan_file, but optimized through parallel processing."""
+        with mp.Pool(processes=num_processors) as p:
+            for secrets in p.imap_unordered(
+                _scan_file_and_serialize,
+                [os.path.join(self.root, filename) for filename in filenames],
+            ):
+                for secret in secrets:
+                    self[os.path.relpath(secret.filename, self.root)].add(secret)
 
     def scan_file(self, filename: str) -> None:
         for secret in scan.scan_file(os.path.join(self.root, filename)):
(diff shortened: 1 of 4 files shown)


detect_secrets/core/secrets_collection.py
```diff
@@ -45,6 +46,20 @@ def load_from_baseline(cls, baseline: Dict[str, Any]) -> 'SecretsCollection':
     def files(self) -> Set[str]:
         return set(self.data.keys())
 
+    def scan_files(self, *filenames: str, num_processors: int = mp.cpu_count()) -> None:
+        """Just like scan_file, but optimized through parallel processing."""
+        if len(filenames) == 1:
+            self.scan_file(filenames[0])
+            return
+
+        with mp.Pool(processes=num_processors) as p:
```
calvinli, line 55: setting `processes` to `mp.cpu_count()` is probably going to be too high if running on a shared machine/CI etc. (especially if it's going to be running for several minutes clock time), maybe let this be configured?

calvinli (approved):
I took a quick peek and it does look like `scan.scan_file` is independent/doesn't rely on any shared state so this should work.

detect_secrets/core/secrets_collection.py
domanchi: Added `--cores` flag to configure this.

PR #442, opened 2021-04-01
Secret regex correction

pablosantiagolopez:
This Pull Request solves the bug commented in #418.

I also add a test case with a long line proposed by @domanchi to check the `KeywordDetector` performance in the unit tests and prevent future performance bugs in this plugin. Note that this test case introduces a little delay in the test execution, but I think that it would be very useful.

### DIFF
--- a/tests/plugins/keyword_test.py
+++ b/tests/plugins/keyword_test.py
@@ -1,3 +1,6 @@
+import base64
+from random import randint
+
 import pytest
 
 from detect_secrets.core.scan import scan_line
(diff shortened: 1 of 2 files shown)


tests/plugins/keyword_test.py
```diff
@@ -10,6 +13,8 @@
 LETTER_SECRET = 'A,.:-¨@*¿?!'
 SYMBOL_SECRET = ',.:-¨@*¿?!'
 
+LONG_LINE = '<img src="data:image/png;base64,{}">'.format(base64.b64encode((str(randint(0, 9)) * 30500).encode()))  # noqa: E501
```
domanchi, line 16:
```suggestion
LONG_LINE = '<img src="data:image/png;base64,{}\n"\n>'.format(base64.b64encode((str(randint(0, 9)) * 30500).encode()))  # noqa: E501
```

The new lines *shouldn't* do anything, but it will screw up regexes that assume properly terminated quotes and HTML tags.

domanchi (approved):
small modification, but otherwise, lgtm! I can also confirm that tests pass, so can merge once addressed.

tests/plugins/keyword_test.py
pablosantiagolopez: I'm sorry. I didn't see your comment the last day. Done!

domanchi (approved):

PR #444, opened 2021-04-03
bug fix: clearing filter caches upon loading new settings

domanchi:
Shout out to @calvinli for finding this one. When you're swapping between `Settings` objects to run more advanced server side analysis, filters' caches do not become invalidated, and cause unexpected results.

This fixes it.


calvinli (approved):
seems kinda magic but sure i guess

PR #448, opened 2021-04-14
Keyword plugin: default regex modification

syn-4ck:
This Pull Request includes the improvements described in #430.

### DIFF
--- a/detect_secrets/util/filetype.py
+++ b/detect_secrets/util/filetype.py
@@ -17,7 +17,11 @@ class FileType(Enum):
     C_SHARP = 11
     C = 12
     C_PLUS_PLUS = 13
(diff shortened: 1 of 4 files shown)


detect_secrets/util/filetype.py
```diff
@@ -17,7 +17,11 @@ class FileType(Enum):
     C_SHARP = 11
     C = 12
     C_PLUS_PLUS = 13
-    OTHER = 14
+    CONFIG = 14
+    INI = 15
+    PROPERTIES = 16
```
domanchi, line 22: Not sure the practical difference between them, but sure.

domanchi (approved):

PR #449, opened 2021-04-15
bug fix: better support for YAML flow style

### LINKED ISSUE #374: yaml tranformer fails on a certain case where the file ends with a blank line
This happens with the pre-v1-launch branch, not reproducible with the master branch.

When a YAML file ends with a block style list of flow style maps followed by a blank line, scanning fails with "list index out of range" at `line=lines[item['__line__'] - 1]` because `__line__` is n but the `lines` list has only n-1 items, no element for the final blank line.

Example:

```
foo:
  - { bar: 'baz' }

```
When it fails, `__line__` is 3, but `lines` has only two items: `['foo:', '   - { bar: "baz" }']`:

```
list index out of range
  File "/Users/gpflaum/ws/detect-secrets-pre-v1-launch/detect_secrets/transformers/yaml.py", line 193, in __iter__
    line=lines[item['__line__'] - 1],
  File "/Users/gpflaum/ws/detect-secrets-pre-v1-launch/detect_secrets/transformers/yaml.py", line 33, in parse_file
    items = sorted(YAMLFileParser(file), key=lambda x: x.line_number)
  File "/Users/gpflaum/ws/detect-secrets-pre-v1-launch/detect_secrets/core/scan.py", line 130, in _get_transformed_file
    return transformer.parse_file(file)
  File "/Users/gpflaum/ws/detect-secrets-pre-v1-launch/detect_secrets/core/scan.py", line 50, in scan_file
    lines = _get_transformed_file(f)
  File "/Users/gpflaum/ws/detect-secrets-pre-v1-launch/detect_secrets/core/secrets_collection.py", line 42, in scan_file
    for secret in scan.scan_file(filename):
  File "/Users/gpflaum/ws/detect-secrets-pre-v1-launch/detect_secrets/core/baseline.py", line 30, in create
    secrets.scan_file(filename)
  File "/Users/gpflaum/ws/detect-secrets-pre-v1-launch/detect_secrets/main.py", line 47, in handle_scan_action
    secrets = baseline.create(*args.path, should_scan_all_files=args.all_files)
  File "/Users/gpflaum/ws/detect-secrets-pre-v1-launch/detect_secrets/main.py", line 26, in main
    handle_scan_action(args)
  File "/Users/gpflaum/ws/detect-secrets-pre-v1-launch/detect_secrets/__main__.py", line 7, in <module>
    sys.exit(main())
```

Parsing is successful when the file doesn't end with a blank line:
```
foo:
  - { bar: 'baz' }]
```
Here are several cases that don't fail when the file ends with a blank line.

Block style map of block style list:
```
foo: 
  - 
    bar: baz

```
Flow style map that isn't in a list:
```
foo: { bar: 'baz' }

```
Block style list of flow style lists:
```
foo:
  - [ "bar", "baz" ]

```
Flow style list of flow style maps:
```
foo: [ { bar: 'baz' } ]

```

domanchi:
## Summary

Fixes #374.

Turns out, YAML parsing is hard. Oof.

## Details

So I spent about two hours digging **deep** into the bowels of YAML parsing. This is what I learned:

**Flow style** is the terminology for using `[ ]` to define sequences, and `{ }` to define mappings. It is still valid YAML, but the underlying parsing engine handles them a little different. Thankfully, we don't need to worry about sequences currently, since we assume secrets don't appear in arrays.

When flow style is encountered, the YAML parser needs to read the entire payload before processing it. Otherwise, it won't know when the mapping ends, and it could be an invalid YAML file (makes sense).

Here's where it gets wacky: for some reason, on the termination of the flow mapping character (`}`), it continues to read characters until it next valid character ([source](https://github.com/yaml/pyyaml/blob/5.4.1/lib3/yaml/reader.py#L108)). In this example:

```yaml
{a: "asdf"}
```

the end-of-file marker is next valid character. However, if we have a new line before that, the `line` variable will be **incremented before parsing**, leading to an off-by-one error causing this traceback.

Apparently, this issue is "well-known" for the YAML parsing library? This is just my hypothesis, given that there's a function dedicated to handling this edge case ([source](https://github.com/yaml/pyyaml/blob/5.4.1/lib3/yaml/parser.py#L537)). Within this function, the parser conveniently stores the start of the FlowMapping character under `self.marks`, which allows us to reference this value.

HOWEVER, we should only reference this value in the case where the FlowMapping character is on the same line as the first key. In other cases, `self.loader.line` should suffice (for bizarre reasons I didn't dig too deeply into).

### DIFF
--- a/detect_secrets/transformers/yaml.py
+++ b/detect_secrets/transformers/yaml.py
@@ -152,6 +152,9 @@ class YAMLFileParser:
         self.loader = yaml.SafeLoader(self.content)
         self.loader.compose_node = self._compose_node_shim  # type: ignore
 
+        self.is_flow_mapping_first_key = False
+        self.loader.parse_flow_mapping_key = self._parse_flow_mapping_key_shim  # type: ignore
+
     def json(self) -> Dict[str, Any]:
         return cast(Dict[str, Any], self.loader.get_single_data())
 
@@ -206,7 +209,11 @@ class YAMLFileParser:
         parent: Optional[yaml.nodes.Node],
         index: Optional[yaml.nodes.Node],
     ) -> yaml.nodes.Node:
-        line = self.loader.line
+        line = (
+            self.loader.marks[-1].line
+            if self.is_flow_mapping_first_key
+            else self.loader.line
+        )
 
         node = yaml.composer.Composer.compose_node(self.loader, parent, index)
         node.__line__ = line + 1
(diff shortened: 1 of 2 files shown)


streichsbaer:
Is there any ETA on shipping this fix? Is support needed?

domanchi:
Closed due to linked PR.

PR #463, opened 2021-06-14
Add a plugin for SendGrid

ninoseki:
Add a plugin for SendGrid API key detection (#462)

### DIFF
--- a/detect_secrets/plugins/sendgrid.py
+++ b/detect_secrets/plugins/sendgrid.py
@@ -1,0 +1,17 @@
+"""
+This plugin searches for SendGrid API keys
+"""
+import re
+
+from detect_secrets.plugins.base import RegexBasedDetector
+
+
+class SendGridDetector(RegexBasedDetector):
+    """Scans for SendGrid API keys."""
+    secret_type = 'SendGrid API key'
+
+    denylist = [
+        # SendGrid API key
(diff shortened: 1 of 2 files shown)


detect_secrets/plugins/sendgrid.py
```diff
@@ -0,0 +1,17 @@
...
+from detect_secrets.plugins.base import RegexBasedDetector
+
+
+class SendGridDetector(RegexBasedDetector):
+    """Scans for SendGrid API keys."""
+    secret_type = 'SendGrid API key'
+
+    denylist = [
+        # SendGrid API key
+        # ref. https://d2w67tjf43xwdp.cloudfront.net/Classroom/Basics/API/what_is_my_api_key.html
```
KevinHock, line 15: In case `https://d2w67tjf43xwdp.cloudfront.net` looks odd to someone reading this, I tried Google-foo'ing and got the same. 

https://docs.sendgrid.com/ui/account-and-settings/api-keys doesn't give an example.

KevinHock (approved):
lgtm, thanks for making this!

ninoseki:
@KevinHock Sorry for providing the URL of non-official website. 
Just a note, here is an evidence to validate the pattern for SendGrid API key. (I masked most of the part. So it might be not persuadable :wink:)

SendGrid

PR #470, opened 2021-07-16
Hook currently fails to interpret filenames with spaces

psedge:
I noticed that `detect-secrets-hook -v --baseline .secrets.baseline $(git ls-files)` interprets filenames containing spaces as separate arguments, preventing them from being scanned. It seems like with the current README.md command it is possible to evade secrets detection by naming files in this way.

To prevent that, it would be better to use the `-z` argument for `git diff` and `git ls-files`, splitting by NUL instead, and passing the arglist to `detect-secrets-hook` via xargs -0.

I'm not sure if the change needs to be made elsewhere as well, possibly in the .pre-commit hook?

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,13 +77,13 @@ For baselines older than version 0.9, just recreate it.
 **Scanning Staged Files Only:**
 
 ```bash
-$ detect-secrets-hook --baseline .secrets.baseline $(git diff --staged --name-only)
+$ git diff --staged --name-only -z | xargs -0 detect-secrets-hook --baseline .secrets.baseline
 ```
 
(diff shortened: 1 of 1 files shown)


KevinHock (approved):
TIL 
```
git ls-files --help
....
-z
           \0 line termination on output and do not quote filenames. See OUTPUT below for more information.
```
and
```
man xargs
....
     -0      Change xargs to expect NUL (``\0'') characters as separators, instead of spaces and newlines.  This is expected to be
             used in concert with the -print0 function in find(1).
```

Thanks @psedge!

PR #359, opened 2020-11-11
Azure Storage Key Detector plugin

DariuszPorowski:
This plugin searches for Azure Storage Account access keys.

Co-authored-by: Dariusz Porowski <Dariusz.Porowski@microsoft.com>

PR #347, opened 2020-09-26
Add npm detector

ninoseki:
Add a new detector which searches for NPM auth tokens.

It detects NPM auth tokens which are probably written in .npmrc.

```
//registry.npmjs.org/:_authToken=foobar
```

PR #474, opened 2021-08-01
Capitalize GitHub and SendGrid Secret Types

KevinHock:
I am not editing the plugins merged in PRs #347 and #359 because of how disruptive that would be to users that already made baselines.

I just merged the GitHub and SendGrid PRs a few minutes ago and have not released a new version, so this should be minimally disruptive to anyone.


KevinHock:
Since this is just a 1-line change for 2 PRs I just reviewed, I'm going to go ahead and merge, as I would have had the original author change this if they didn't make their PRs so long ago.

PR #472, opened 2021-07-27
[#471] Safely get line_number from PotentialSecret

js-truework:
This is a relatively naive approach to solving the issue in #471. I looked at the PotentialSecret class and it very clearly has a default value of 0 for `line_number`. But alas, intermittently it does not. So this should safely fall back to 0 if `line_number` isn't present.

Closes #471 (I think)

PR #476, opened 2021-08-01
Fix #472

KevinHock:
Mainly fixes #472, my bad for merging that, didn't realize Travis CI was gone.

```
    :snake:Prettify a few lines
    
    Add start keyword to enumerate argument
    set([]) -> set()
```
...
```
    :bug:Fix .get to be getattr
    
    PotentialSecret is not a dictionary, so do getattr and not .get
    Fixes #472
```


js-truework:
Sorry for the `.get` usage! I actually realized much later that night that I probably needed to use `getattr` and then I got caught up with work and didn't circle back in time to catch it before the merge.

PR #477, opened 2021-08-04
Fix Unescaped anychar in Slack URL RegEx

turbo:
The two `.` in the regex for slack are unescaped, which means they match more hostnames than `hooks.slack.com` (for example `hooks-slack.com`). This PR adds escape characters to prevent this.


domanchi (approved):
Nice find!

domanchi:
@KevinHock : wanna merge this?

KevinHock:
Oof, that is a good 🐛  fix!

KevinHock:
Thanks for the ping! Always happy to merge @domanchi  approved PRs ;)

KevinHock:
_I'll try and bump the version Thursday or sometime this weekend_ 🤜🏼

PR #478, opened 2021-08-11
Yielding submatches when findall() returns tuples

sinKettu:
Hello!
I assume such changes can be helpful for several users. Let me explain.
For instance, some user writes its own plugin like the following:

``` python
import re
from .base import RegexBasedDetector


class SomeKeyDetector(RegexBasedDetector):
    secret_type = 'Some Detector'

    denylist = [
        re.compile(r'^(key|token|password)=(prefix_)?.*$'),
    ]

```

As you can see, the only regex of the detector contains two match groups: `(key|token|password)` and `(prefix_)` and the function `re.findall()` can return a tuple(s). For exmaple:

``` python
>>> import re
>>> regex = r"^(key|token|password)=(prefix_)?.*$"
>>> string = "key=prefix_45guh245ghihfrgjdsfguhrwth"
>>> print(re.findall(regex, string))
[('key', 'prefix_')]
>>> 

```

In this case, just yielding a `match` with such a plugin will lead to exception in the `detect_secrets/core/potential_secret.py:71`, because it will try to call `.encode()` method for the instance of a `tuple`.

### DIFF
--- a/detect_secrets/plugins/base.py
+++ b/detect_secrets/plugins/base.py
@@ -146,7 +146,7 @@ class RegexBasedDetector(BasePlugin, metaclass=ABCMeta):
         for regex in self.denylist:
             for match in regex.findall(string):
                 if isinstance(match, tuple):
-                    for submatch in filter(len, tuple):
+                    for submatch in filter(bool, tuple):
                         # It might make sense to paste break after yielding
                         yield submatch
                 else:
--- a/detect_secrets/plugins/base.py
+++ b/detect_secrets/plugins/base.py
@@ -145,7 +145,12 @@ class RegexBasedDetector(BasePlugin, metaclass=ABCMeta):
     def analyze_string(self, string: str) -> Generator[str, None, None]:
         for regex in self.denylist:
             for match in regex.findall(string):
-                yield match
+                if isinstance(match, tuple):
+                    for submatch in filter(len, tuple):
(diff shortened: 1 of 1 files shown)


detect_secrets/plugins/base.py
```diff
@@ -145,7 +145,12 @@ def denylist(self) -> Iterable[Pattern]:
     def analyze_string(self, string: str) -> Generator[str, None, None]:
         for regex in self.denylist:
             for match in regex.findall(string):
-                yield match
+                if isinstance(match, tuple):
+                    for submatch in filter(len, tuple):
```
domanchi, line 149:
```suggestion
                    for submatch in filter(bool, tuple):
```

I think this is what you were going for? Do let me know if it isn't. I think this makes a little more sense in terms of readability.

domanchi (approved):
Great find! Thanks for your contribution.

detect_secrets/plugins/base.py
sinKettu: Hello! Thanks! Yes, I think, this more suitable change.

domanchi: Great. Once you merge this fix in, @KevinHock can help push this through to master.

sinKettu: Done.

PR #482, opened 2021-09-16
Fix 'no plugins found' bug on python > 3.7

Ian288:
Fixes the 'no plugins found' bug on python > 3.7 by explicitly passing
the current settings object to an initialization function that runs
before the multiprocessing job is started. This initialization creates a
new settings object for that process and sets its properties equal to
that of the settings object passed from the parent process.

### DIFF
--- a/detect_secrets/core/secrets_collection.py
+++ b/detect_secrets/core/secrets_collection.py
@@ -10,11 +10,44 @@ from typing import Set
 from typing import Tuple
 
 from . import scan
+from ..settings import get_settings
+from ..settings import Settings
 from .potential_secret import PotentialSecret
+
+
+def initialize_settings(settings: Settings) -> None:
+    new_settings = get_settings()
+    new_settings.set(settings)
+
+    if 'detect_secrets.filters.wordlist.should_exclude_secret' in settings.filters:
+        config = settings.filters[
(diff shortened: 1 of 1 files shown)


benjivesterby (approved):

detect_secrets/core/secrets_collection.py
```diff
@@ -55,7 +90,11 @@ def scan_files(self, *filenames: str, num_processors: Optional[int] = None) -> N
         if not num_processors:
             num_processors = mp.cpu_count()
 
-        with mp.Pool(processes=num_processors) as p:
+        with mp.Pool(
+            processes=num_processors,
+            initializer=initialize_settings,
+            initargs=(get_settings(),),
```
domanchi, line 96:
```suggestion
            initializer=configure_settings_from_baseline,
            initargs=(get_settings().json(),),
```

Instead of passing around Python objects (and repeating the logic found in `detect_secrets.settings.configure_settings_from_baseline`), does this work instead?

Or, if we wanted to optimize it even more, we can create the JSON representation once, and pass it into `initargs`.

domanchi (commented):
After *many* months of this bug being live in production, thank you @Ian288 for finally contributing a PR to fix this!

jpdakran:
Closing as this was merged as part of #505

PR #486, opened 2021-10-22
Update jwt.py

knadt:
Added an is_time_valid method to check jwt tokens are not expired

### DIFF
--- a/detect_secrets/plugins/jwt.py
+++ b/detect_secrets/plugins/jwt.py
@@ -2,8 +2,11 @@
 This plugin finds JWT tokens
 """
 import base64
+import jwt
 import json
 import re
+import datetime  
+from datetime import datetime
 from typing import Generator
 
 from .base import RegexBasedDetector
@@ -19,6 +22,7 @@ class JwtTokenDetector(RegexBasedDetector):
     def analyze_string(self, string: str) -> Generator[str, None, None]:
         yield from filter(
             self.is_formally_valid,
+            self.is_time_valid,
             super().analyze_string(string),
         )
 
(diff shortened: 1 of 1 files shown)


detect_secrets/plugins/jwt.py
```diff
@@ -19,6 +22,7 @@ class JwtTokenDetector(RegexBasedDetector):
     def analyze_string(self, string: str) -> Generator[str, None, None]:
         yield from filter(
             self.is_formally_valid,
+            self.is_time_valid,
```
domanchi, line 25: Uh...does this work?

Based on the [python documentation](https://docs.python.org/3/library/functions.html#filter), `filter` takes in a function, and an iterable. Based on this, I'm surprised that three arguments would actually pass tests...

detect_secrets/plugins/jwt.py
```diff
@@ -2,8 +2,11 @@
 This plugin finds JWT tokens
 """
 import base64
+import jwt
```
domanchi, line 5: Historically, I've attempted to reduce the number of dependencies that `detect-secrets` needs to install. This is probably the reason why `is_formally_valid` doesn't use the library. Furthermore, if we require this, then it would require all users to install `pyjwt`, even if they don't use this plugin.

If we're not verifying signatures, this should be an easier fix: please try to do this manually, rather than use a library. The good news is that you can use a lot of the same logic as already defined in `is_formally_valid`.

detect_secrets/plugins/jwt.py
```diff
@@ -2,8 +2,11 @@
...
 import json
 import re
+import datetime
```
domanchi, line 8: This line is not needed since you redeclare the `datetime` variable in the line right after this.

detect_secrets/plugins/jwt.py
```diff
@@ -42,5 +46,15 @@ def is_formally_valid(token: str) -> bool:
                     _ = json.loads(b64_decoded.decode('utf-8'))
             except (TypeError, ValueError, UnicodeDecodeError):
                 return False
+            
+    @staticmethod
+    def is_time_valid(token: str) -> bool:
+        try:
+            json_data = jwt.decode(token,options={"verify_signature": False})
+            json_exp_date_time = datetime.fromtimestamp( json_data['exp']  )
```
domanchi, line 54: What about `KeyError`s if the payload doesn't have an expiry time?

domanchi (changes requested):

PR #501, opened 2022-01-07
Bug Fix: Fix Inline Dictionary For YAML Files

jpdakran:
## Problem
- When running detect-secrets on YAML files - there is a miscalculation that occurs when calculating the line number of an inline dictionary (FlowMapping). 
- This is an issue with the PyYAML library. 
- This issue occurs when there is an empty line after the inline dictionary.
- An example of where the miscalculation occurs:
```
dictionary: 
     - {keyA: valueA, keyB: valueB, keyC: valueC}
     
```
## Solution

- This solution was inspired by this [PR](https://github.com/Yelp/detect-secrets/pull/449) but was expanded on to fix a few other use-cases
- The parser keeps track of the the start of the FlowMapping character under `self.marks`
- We need to use this values line number when the inline dictionary has a key on the same line as the start of the FlowMapping.
- The PR above handles the case only when there is a SINGLE key value pair followed by an empty line. Ex. `{key: value}`
- To handle the cases where there are multiple key value pairs on a single line - we must check that the next upcoming tokens are of type `FlowEntryToken, KeyToken` (Comma, new Key). This will tell us if this dictionary has more values on the same line. Ex. `{keyA: valueA, keyB: valueB, keyC: valueC}`

## Verification
- A suite of unit tests were created to validate the approach.

### DIFF
--- a/detect_secrets/transformers/yaml.py
+++ b/detect_secrets/transformers/yaml.py
@@ -164,7 +164,7 @@ class YAMLFileParser:
 
         to_search = deque([self.json()])
         while to_search:
-            item: Any = to_search.pop()
+            item: Any = to_search.popleft()
 
             if not item:
                 # mainly for base case (e.g. if file is all comments)
--- a/detect_secrets/transformers/yaml.py
+++ b/detect_secrets/transformers/yaml.py
@@ -14,6 +14,8 @@ from typing import Tuple
 from typing import Union
 
 import yaml
+from yaml.tokens import FlowEntryToken
+from yaml.tokens import KeyToken
 
 from ..types import NamedIO
 from ..util.filetype import determine_file_type
@@ -152,6 +154,9 @@ class YAMLFileParser:
         self.loader = yaml.SafeLoader(self.content)
         self.loader.compose_node = self._compose_node_shim  # type: ignore
 
+        self.is_inline_flow_mapping_key = False
+        self.loader.parse_flow_mapping_key = self._parse_flow_mapping_key_shim  # type: ignore
+
     def json(self) -> Dict[str, Any]:
         return cast(Dict[str, Any], self.loader.get_single_data())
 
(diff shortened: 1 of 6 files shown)


detect_secrets/transformers/yaml.py
```diff
@@ -218,6 +227,41 @@ def _compose_node_shim(
 
         return cast(yaml.nodes.Node, node)
 
+    def _parse_flow_mapping_key_shim(
```
calvinli, line 230: I trust you that this works, but could you write some comments here explaining what this code does?

jpdakran: Updated. Please let me know if that makes sense.

calvinli (approved):

KevinHock:
Great stuff @jpdakran!

detect_secrets/transformers/yaml.py
```diff
@@ -218,6 +227,49 @@ def _compose_node_shim(
...
+        # mapping, we will use this line number when we are processing:
+        # A) The first key in an inline dictionary where the flow mapping start character is on the
+        # same line as the key
+        # B) The n key of an inline dictionary that is followed by a FlowEntryToken (',') and
+        # KeyToken ('key:')
+        if (
+            first
+            and self.loader.marks[-1].line == self.loader.peek_token().start_mark.line
+            or self._check_next_tokens_shim(FlowEntryToken, KeyToken)
+        ):
```
KevinHock, line 246: Super nit, but you could assign the bool, might be more pythonic

detect_secrets/transformers/yaml.py
```diff
@@ -218,6 +227,49 @@ def _compose_node_shim(
...
+        else:
+            self.is_inline_flow_mapping_key = False
+
+        return cast(yaml.nodes.Node, yaml.parser.Parser.parse_flow_mapping_key(self.loader, first))
+
+    def _check_next_tokens_shim(
+        self,
+        *choices: Any,
+    ) -> bool:
+        # Check the next tokens type match the argument list of token types
```
KevinHock, line 257: Repo Convention more uses `""""""`, e.g. https://github.com/Yelp/detect-secrets/blob/f295adadb7b5b40f3a24d574a7a3d0988d943860/detect_secrets/core/secrets_collection.py#L17

detect_secrets/transformers/yaml.py
jpdakran: Got it. Please see update PR: https://github.com/Yelp/detect-secrets/pull/504

detect_secrets/transformers/yaml.py
jpdakran: Got it. Please see updated PR: https://github.com/Yelp/detect-secrets/pull/504

PR #505, opened 2022-01-18
Bug Fix: Plugins Fail To Load on Python 3.8+

jpdakran:
# Problem
- Python 3.8+ multiprocessing library has switched to `spawn` where before it was `fork`
- This is an issue since `spawn` does not share any memory between the parent and child processes. 
- The plugins/filters are stored as singleton global variable in the Settings class

# Solution
- Default the multiprocessing to `fork` when creating a pool of processes
- Python does have a shared memory library now however I find that it is only useful when sharing simple data types like strings, ints, list but not custom classes

### DIFF
--- a/detect_secrets/core/secrets_collection.py
+++ b/detect_secrets/core/secrets_collection.py
@@ -11,6 +11,8 @@ from typing import Tuple
 
 from . import scan
 from .potential_secret import PotentialSecret
+from detect_secrets.settings import configure_settings_from_baseline
+from detect_secrets.settings import get_settings
 
 
 class PatchedFile:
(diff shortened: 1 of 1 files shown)


calvinli (commented):
nice find, let's hold off on merging this until we get CI set up again though, I'm interested to see whether py38/py39 tests pass on CI

KevinHock:
Hi @jpdakran, I also saw #482, as a possible alternative solution.

@domanchi  can you review this one? I think you have the most context.

detect_secrets/core/secrets_collection.py
```diff
@@ -55,7 +55,8 @@ def scan_files(self, *filenames: str, num_processors: Optional[int] = None) -> N
         if not num_processors:
             num_processors = mp.cpu_count()
 
-        with mp.Pool(processes=num_processors) as p:
+        # Default to fork multiprocessing. Python 3.8+ defaults to spawn which doesn't share memory.
+        with mp.get_context('fork').Pool(processes=num_processors) as p:
```
calvinli, line 59: I did find on https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

that it says `fork` should not be used on macOS, and that it's not available on Windows

so it may be better to resolve this some other way

domanchi:
I think it may be better to use the approach with #482 -- essentially, initializing a new settings object per process, rather than sharing memory. It's been a while since I looked at it, but the `Settings` object was designed for easy serialization/deserialization, so you could pass in a serialized object to the spawned process, and hydrate it when starting it up.

This way, we'd be less dependent on OS level details since this library is used on multiple platforms.

detect_secrets/core/secrets_collection.py
jpdakran: Updated to pass the settings object to the child processes.

calvinli (approved):

PR #509, opened 2022-02-01
Fix Catastophic Backtracking for Indirect Reference

jpdakran:
# Problem
- I've noticed that there is a hang when running detect-secrets.
- This is due to the regex that the Indirect-Reference filter utilizes.
- The problem here is catastrophic backtracking. This occurs when we are trying to match a line that is typically more than 25k characters against the regex. 

# Solution
- We have a few options here 
1) Add conditions on when to run the regex for a line - 1k characters or less (or some agreed up length)
2) Update the regex 
- I resorted to option 1. The intention of the heuristic should be for short and simple lines.

### DIFF
--- a/detect_secrets/filters/heuristic.py
+++ b/detect_secrets/filters/heuristic.py
@@ -165,6 +165,8 @@ def is_indirect_reference(line: str) -> bool:
 
         secret = request.headers['apikey']
     """
+    if len(line) > 1000:
+        return False
     return bool(_get_indirect_reference_regex().search(line))
 
 
@@ -181,7 +183,7 @@ def _get_indirect_reference_regex() -> Pattern:
     #       [^\v]*      ->  Something except line breaks
     #       [\]\)]      ->  End of indirect reference: ] or )
     #   )
-    return re.compile(r'(\w+)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
+    return re.compile(r'([^\v=!:]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
 
 
 def is_lock_file(filename: str) -> bool:
--- a/detect_secrets/filters/heuristic.py
+++ b/detect_secrets/filters/heuristic.py
@@ -181,7 +181,7 @@ def _get_indirect_reference_regex() -> Pattern:
     #       [^\v]*      ->  Something except line breaks
     #       [\]\)]      ->  End of indirect reference: ] or )
     #   )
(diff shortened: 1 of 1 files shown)


detect_secrets/filters/heuristic.py
```diff
@@ -181,7 +181,7 @@ def _get_indirect_reference_regex() -> Pattern:
     #       [^\v]*      ->  Something except line breaks
     #       [\]\)]      ->  End of indirect reference: ] or )
     #   )
-    return re.compile(r'([^\v=!:]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
+    return re.compile(r'(\w+)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
```
calvinli, line 184: nice catch  :)

comment on line 174 needs updating

do you know if there are unit tests covering this regex? I'd be curious what practical effect this change has (lines that it matches now that it didn't before or vice versa)

calvinli: also given how this regex is used, I feel like option 1 is an attractive idea, I can't imagine why you'd want this heuristic to apply to lines that long, and since it's a heuristic having false negatives doesn't matter too much

calvinli: I'm also trying this on regex101.com and it seems to still have just as much backtracking as it did before

jpdakran: There are unit tests but they do not have much coverage - only 3 cases:
`    'line, result',
    (
        ('secret = get_secret_key()', True),
        ('secret = request.headers["apikey"]', True),
        ('secret = hunter2', False),
    ),
)`

jpdakran: I'm wondering if we apply both approaches. Regex update and a cap on line length - maybe 20k characters?

calvinli: It sounds like the kinds of lines this is trying to catch should be pretty short (similar to `secret = get_secret_key()`, function calls and the like), so I'd put an even lower limit, 1K or so (and in practice I wouldn't expect to see anything much above 100 chars)

jpdakran: Updated

calvinli (approved):

detect_secrets/filters/heuristic.py
```diff
@@ -165,6 +165,8 @@ def is_indirect_reference(line: str) -> bool:
 
         secret = request.headers['apikey']
     """
+    if len(line) > 1000:
```
domanchi, line 168: @jpdakran : we should document design rationales in code to assist future developers in understanding the codebase.

jpdakran: Understood. Please see https://github.com/Yelp/detect-secrets/pull/516 for update.

PR #510, opened 2022-02-01
fix: wrong quotes used in code snippets

paulo-sampaio:
the code examples on 'Report generation' section had the wrong triple quotes.


KevinHock (approved):
Thank you very much @paulo-sampaio

PR #512, opened 2022-02-04
Bug Fix: Yaml Line Numbering Empty Mapping Fix

jpdakran:
# Problem
- When we have an empty mapping (dictionary) included in a file - the line numbering of the file is off
- It is specifically off by `actual line number - 1`
- The issue here is that the only time we parse a flow mapping key for an empty mapping is once. (At function `_parse_flow_mapping_key_shim`)
- When we parse a flow mapping key we determine if this is an inline mapping or not. 
- We determine that this is an inline mapping (correct) - however since it is empty mapping, we never come back into parsing the flow mapping to change the `is_inline_flow_mapping_key` to False. 

# Solution
- Set the `is_inline_flow_mapping_key` to False when the mapping has ended.

### DIFF
--- a/detect_secrets/transformers/yaml.py
+++ b/detect_secrets/transformers/yaml.py
@@ -221,6 +221,7 @@ class YAMLFileParser:
         node.__line__ = line + 1
 
         if node.tag.endswith(':map'):
+            self.is_inline_flow_mapping_key = False
             return _tag_dict_values(node)
 
         # TODO: Not sure if need to do :seq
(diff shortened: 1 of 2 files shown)


detect_secrets/transformers/yaml.py
```diff
@@ -221,6 +221,7 @@ def _compose_node_shim(
         node.__line__ = line + 1
 
         if node.tag.endswith(':map'):
+            self.is_inline_flow_mapping_key = False
```
calvinli, line 224: Can you put some of your explanation from your PR description into here as a code comment?

jpdakran: Updated

calvinli (approved):

PR #516, opened 2022-02-07
Document Design Rationales for Heuristic Constraint

jpdakran:
- Added comment for the length constraint on the indirect_reference heuristic constraint

### DIFF
--- a/detect_secrets/filters/heuristic.py
+++ b/detect_secrets/filters/heuristic.py
@@ -165,8 +165,8 @@ def is_indirect_reference(line: str) -> bool:
 
         secret = request.headers['apikey']
     """
-    # Constain lines as the heuristics intention is to target relatively short function like strings
-    # This constraint also avoids catastrophic backtracking failures of the regex.
+    # Constrain line length as the heuristic's intention is to target lines that resemble
+    # function calls. The constraint avoids catastrophic backtracking failures of the regex.
     if len(line) > 1000:
         return False
     return bool(_get_indirect_reference_regex().search(line))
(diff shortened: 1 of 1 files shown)


detect_secrets/filters/heuristic.py
```diff
@@ -165,6 +165,8 @@ def is_indirect_reference(line: str) -> bool:
 
         secret = request.headers['apikey']
     """
+    # Constain lines as the heuristics intention is to target relatively short function like strings
```
calvinli, line 168: `Constain` -> `Constrain`, `heuristics` -> `heuristic's` and `function like strings` -> something like `lines like function calls`

and `also` in the next line sounds weird, isn't that the primary reason for this constraint?

calvinli: also not that i have any problem at all with code comments :)  but IMO the reasoning here would have been pretty clear from git blame to anyone interested, maybe i'm just spoiled by tooling though

jpdakran: Yep. That makes sense. Updated.

detect_secrets/filters/heuristic.py
```diff
@@ -165,6 +165,8 @@ def is_indirect_reference(line: str) -> bool:
 
         secret = request.headers['apikey']
     """
+    # Constrain line length as the heuristic's intention
+    # is to target lines that resemble function calls
```
calvinli, line 169: lol sorry for the back and forth but IMO this now misses the important part which is that this constraint was added to prevent the regex from taking a long time

jpdakran: No worries. That's the whole point of reviews :)

calvinli (approved):

Issue #489, opened 2021-11-09
is_likely_id_string filter creates false negatives


fproulx-boostsecurity:
This filter will eliminate true positives by mistake
https://github.com/Yelp/detect-secrets/blob/b206dc29f274839ebd3c522fed19448c9b8f3f70/detect_secrets/filters/heuristic.py#L71

The following line in a scanned file:
```
DATABASE_URL = "postgres://david:ThisIsNotdavid8789albino@126.234.54.3/somedatabase";
```

Some snippet to test:
```
import re
secret = 'ThisIsNotdavid8789albino'
line = f"DATABASE_URL = 'postgres://david:{secret}@126.234.54.3/somedatabase';"
bool(re.compile(r'id[^a-z0-9]', re.IGNORECASE).search(line, pos=0, endpos=line.index(secret)))
>> True
```

domanchi:
@fproulx-boostsecurity : great find! Thanks for reporting this. When we resolve our current issue with the lack of authorized contributors, hopefully we can resolve this issue.

jpdakran:
Closing as fixed in [526](https://github.com/Yelp/detect-secrets/pull/526)

PR #526, opened 2022-03-30
Make is_likely_id_string More Strict

jpdakran:
Issue: [489](https://github.com/Yelp/detect-secrets/issues/489)

Problem:
- The `is_likely_id_string` regex can eliminate true positives by mistake.
- Ex. Anything ending in `id` like the name `david` followed by a non-letter/non-numeric character will match

Solution:
- Make the regex more strict to identify id strings.
- This can be achieved by enforcing the string starts with the `id` identifier however we lose out on the other typical id_string identifiers like `myid`, `userid` etc.
- We can enforce these additional identifiers
- Also enforce the `_id` identifier and this can exist in the middle of a string

### DIFF
--- a/detect_secrets/filters/heuristic.py
+++ b/detect_secrets/filters/heuristic.py
@@ -68,7 +68,7 @@ def is_likely_id_string(secret: str, line: str) -> bool:
 
 @lru_cache(maxsize=1)
 def _get_id_detector_regex() -> Pattern:
-    return re.compile(r'id[^a-z0-9]', re.IGNORECASE)
+    return re.compile(r'^(id|myid|userid)|_id[^a-z0-9]', re.IGNORECASE)
 
 
 def is_non_text_file(filename: str) -> bool:
--- a/tests/filters/heuristic_filter_test.py
+++ b/tests/filters/heuristic_filter_test.py
@@ -63,6 +63,12 @@ class TestIsLikelyIdString:
             ('RANDOM_STRING', 'myid: RANDOM_STRING'),
             ('RANDOM_STRING', 'myid=RANDOM_STRING'),
             ('RANDOM_STRING', 'myid = RANDOM_STRING'),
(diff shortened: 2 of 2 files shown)


detect_secrets/filters/heuristic.py
```diff
@@ -68,7 +68,7 @@ def is_likely_id_string(secret: str, line: str) -> bool:
 
 @lru_cache(maxsize=1)
 def _get_id_detector_regex() -> Pattern:
-    return re.compile(r'id[^a-z0-9]', re.IGNORECASE)
+    return re.compile(r'^(id|myid|userid)|_id[^a-z0-9]', re.IGNORECASE)
```
lorenzodb1, line 71: It looks like it doesn't enforce strings that end with `_id` (ex. my_id). It also doesn't enforce strings where id is used in its plural form (ids). Could it be worth including both?

jpdakran: So this would identify the following string `my_id: POTENTIAL_SECRET` as an id_string. The idea is to identify strings that follow the format `id: POTENTIAL_SECRET`. We want to do this to filter them out as potential secrets (We detected the secret but are filtering it out of the result set since it is an id string).

lorenzodb1: Oh okay, so it relies on the fact that all names will always be followed by `:`. What about the plural form of id? Would something like `my_ids: POTENTIAL_SECRET` be flagged?

jpdakran: Right. Not necessarily only `:`. We are trying to mimic some sort of assignment operation. So it can be followed by any non-letter/non-numeric character. Ex. `space, =` etc. Currently the plural of ids wouldn't be flagged and would not be filtered out. We can add the plural incase of a series of ids possibly.

lorenzodb1 (approved):
LGTM 🎈

Issue #493, opened 2021-11-19
detect-secrets not identifying all the secret occurrences in a file


BhaskiBuzz:
If the secret written into a file at multiple locations, only the first one being identified by detect-secrets. Remaining instances are not getting reported by detect-secrets.

Example code in demo.txt
1. secret="abcd"
2. random code
3. secret="abcd"

Now run the detect-secrets scan on the above file and observe that the issue will be reported only at the line 1 and it is not reported the line number 3.

BhaskiBuzz:
Hi @lorenzodb1 any update on this?

geekNero:
Hey, this is mentioned in the docs, to avoid this you can do the following: 
go to detect-secrets/detect_secrets/core/potential_secret.py on line number 53, you'll find a list. In that list add another parameter "line_number". So it should look something like this: self.fields_to_compare = ['filename', 'secret_hash', 'type', 'line_number']

a-weiss-programmer:
I'm running into this issue as well. 

Would it be worth adding a command line flag/programmatic setting of this behavior?

geekNero:
I'll give it a go but the devs have been inactive for a while, I doubt they'll accept it. Still I'll drop the link to my forked repository.

lorenzodb1:
Hello everyone, apologies for the delayed reply. I've gathered some context around this issue and it turns out that _this is actually a feature_ ™. Jokes aside, the idea behind this is that if `detect-secrets` flags a secret in a file, a developer would look for other instances of the same secret in it. Now, I personally do like the idea of flagging all instances, but the change is not trivial. #527 would help a lot, and I'm pushing for it to be merged as soon as possible. In the meantime, forgive us and stay tuned for more updates :)

reddybhaskarvengala:
Hi @lorenzodb1 thank you for the update.

PR #527, opened 2022-04-01
Add Occurences to Potential Secrets

jpdakran:
A potential secret is uniquely identified by three main properties: `filename, secret_hash, type`. Since `line_number` is not included this means we will only track a single potential secret in a file although it may appear on multiple lines. This is a fundamental design decision. 

The idea behind this pull request is to give users some insight to how many occurences a potential secret is showing up in their files. This does not violate the fundamental design above but rather serves as an improvement hopefully giving better insight to developers on the overall scope of their potential secrets. 

One of the key changes is changing the `SecretsCollection` data structure from a `Set` to a `List`. The reasoning behind this decision is when we are scanning through the file - we do not only want to deny duplicate secrets to the data structure but alter the existing potential secrets occurrences. `List` gives us a lot of flexibility with changing properties in the data structure while `Set` is a little restricted. The new `List` will operate almost identically to the previous `Set` in terms of not allowing duplicate secrets but with the addition of tracking occurrences. 

The `occurrences` property will operate very similar to the `line_number` property. They are similar in the way that we will not alert on the change of this property but rather only notify the user that the baseline file changed with updated information on their secrets.

### DIFF
--- a/detect_secrets/core/secrets_collection.py
+++ b/detect_secrets/core/secrets_collection.py
@@ -31,7 +31,7 @@ class SecretsCollection:
             relative to root, since we're running as if it was in a different directory,
             rather than scanning a different directory.
         """
-        self.data: Dict[str, Set[PotentialSecret]] = defaultdict(set)
+        self.data: Dict[str, List[PotentialSecret]] = defaultdict(list)
         self.root = root
 
     @classmethod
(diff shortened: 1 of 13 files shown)


detect_secrets/core/secrets_collection.py
```diff
@@ -180,7 +187,11 @@ def trim(
                     # Only update line numbers if we're tracking them.
                     existing_secret.line_number = secret.line_number
```
lorenzodb1, line 188: Looking at your implementation, it looks like it could be fairly trivial to track the line number of each of the occurrences found and report them as it was mentioned in #493.

KBiru:
Hi, @lorenzodb1 Is this feature already added, or it is still in pending review state?

BhaskiBuzz:
Hi,

Any update on this feature? it seems open for very long time. 

Thanks,
Bhaskar.

PR #575, opened 2022-06-28
Persist '# pragma: allowlist nextline secret' in IniFileParser For ConfigFileTransformers

jpdakran:
# Problem
- `# pragma: allowlist nextline secret` are not filtered out and secrets on the proceeding line are alerted on
- This is an issue with the `ConfigFileTransformer` which leverages the `IniFileParser`
- The transformer and the parser does not track comments in files - This is why our logic is alerting on the secrets which should be whitelisted

# Solution
- When parsing files in the `IniFileParser` persist comments that match the allowlist_nextline_secret comment
- We do not want to keep all comments and we do not want to keep allowlist_secret comment (Which occur on the same line as a secret) since this is working properly


lorenzodb1 (approved):
LGTM

PR #578, opened 2022-07-06
Fix is_verified not stored in PotentialSecret

abdelrahman-thafeer:
Problem:

- `is_verified` flag does not get stored in PotentialSecret which results in its value being always `false`. 
- This problem didn't affect `--only-verified` flag because verification was running but not persisted.

Solution:

- When a secret is found, run `verify` method and store the result within `PotentialSecret`.
-  pass down `CodeSnippet` to `analyze_line` as it is required to verify secrets in some plugins.
- `verify` will not run in case `--no-verify` flag was passed.

### DIFF
--- a/detect_secrets/core/scan.py
+++ b/detect_secrets/core/scan.py
@@ -17,6 +17,7 @@ from ..transformers import get_transformed_file
 from ..types import NamedIO
 from ..types import SelfAwareCallable
 from ..util import git
+from ..util.code_snippet import CodeSnippet
 from ..util.code_snippet import get_code_snippet
 from ..util.inject import call_function_with_arguments
 from ..util.path import get_relative_path
@@ -112,6 +113,7 @@ def scan_line(line: str) -> Generator[PotentialSecret, None, None]:
         'detect_secrets.filters.common.is_invalid_file',
     )
     get_filters.cache_clear()
+    context = get_code_snippet(lines=[line], line_number=1)
 
     yield from (
         secret
@@ -122,6 +124,7 @@ def scan_line(line: str) -> Generator[PotentialSecret, None, None]:
             line=line,
             line_number=0,
             enable_eager_search=True,
+            context=context,
         )
         if not _is_filtered_out(
             required_filter_parameters=['context'],
@@ -129,10 +132,7 @@ def scan_line(line: str) -> Generator[PotentialSecret, None, None]:
             secret=secret.secret_value,
             plugin=plugin,
             line=line,
-            context=get_code_snippet(
-                lines=[line],
-                line_number=1,
-            ),
+            context=context,
         )
     )
 
@@ -225,10 +225,11 @@ def _scan_for_allowlisted_secrets_in_lines(
     line_numbers, lines = zip(*lines)
     line_content = [line.rstrip() for line in lines]
     for line_number, line in zip(line_numbers, line_content):
+        context = get_code_snippet(line_content, line_number)
         if not is_line_allowlisted(
             filename,
             line,
-            context=get_code_snippet(line_content, line_number),
+            context=context,
         ):
             continue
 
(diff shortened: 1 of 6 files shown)


detect_secrets/core/scan.py
```diff
@@ -225,18 +225,19 @@ def _scan_for_allowlisted_secrets_in_lines(
     line_numbers, lines = zip(*lines)
     line_content = [line.rstrip() for line in lines]
     for line_number, line in zip(line_numbers, line_content):
+        context = get_code_snippet(line_content, line_number)
         if not is_line_allowlisted(
             filename,
             line,
-            context=get_code_snippet(line_content, line_number),
+            context=context,
```
lorenzodb1, line 232: It'd be nice to have all parameters specified similarly to what you did here (e.g. `line=line`)

lorenzodb1 (commented):
I re-ran the tests as a couple were running for over an hour. Hopefully they'll all pass this time.

jpdakran (changes requested):
I think we need to mock out the verify for the `detect_secrets.plugins.aws.verify_aws_secret_access_key` in `main_test.py::TestSlimScan::test_basic`

abdelrahman-thafeer:
It looks like this is an old [issue](https://github.com/python/cpython/issues/74570) with macOS that occurs when calling `requests` API within a forked process. It didn't happen before because the secret was being filtered by `gibberish` before it reaches to the verification filter. 

This currently impacts few tests (`main_test.py::TestSlimScan::test_basic ` and any test in `audit/audit_test.py` that uses `test_baseline` fixture). I've added an environment variable `NO_PROXY` as a workaround.

tox.ini
```diff
@@ -7,6 +7,10 @@ tox_pip_extensions_ext_venv_update = true
 
 [testenv]
 passenv = SSH_AUTH_SOCK
+# NO_PROXY is needed to call requests API within a forked process
+# when using macOS and python version 3.6/3.7
+setenv =
+    NO_PROXY = '*'
```
jpdakran, line 13: Is this only a testing issue or do we need to somehow enable this in a production mode as well?

jpdakran: Can we verify that this was existing in some sort of integration test / brute force testing. I would just like to ensure we are not introducing any new issues to the repository.

lorenzodb1: If this is caused by `requests`, could it be the case that a newer version doesn't have this issue anymore? I see we're using `2.26.0`, while the latest one is `2.28.1`.

abdelrahman-thafeer: Seems like [requests v2.28.0](https://requests.readthedocs.io/en/latest/community/updates/#id2) dropped support for python 3.6, I tried `requests` v2.27.1 with no luck. 

It's worth noting that the issue can be fixed if we use `spawn` context for multiprocessing but I think that would come at a performance cost.

jpdakran: Right. So python 3.8+ multiprocessing library has switched to spawn where before it was fork. This is an issue since spawn does not share any memory between the parent and child processes. So we default the multiprocessing to fork when creating a pool of processes

jpdakran: Can we try and use `v2.28.1`. Python 3.6 is end of life. We can consider dropping support for it.

abdelrahman-thafeer: I tried `v2.28.1`, still the same problem. I think this is more of a python issue than requests, the default context was changed to spawn because of crashes similar to this. 

Since we are already using default multiprocessing (spawn) for python 3.8+ on macOS, what do you think of checking the current OS and if it's macOS we change to spawn?

jpdakran: Sorry, that was the initial attempt to fix it. We decided to try and stay away from being OS specific so instead of changing the default multiprocessing, we instead passed the shared settings to child processes from the parent. So in python 3.8+ we are using spawn which is the default.

jpdakran (approved):

PR #579, opened 2022-07-09
add py.typed file to enable type hints for package consumers

gruebel:
Just added the missing `py.typed` file. So other projects, which add the project as a library will benefit from the inline type hints. Without this file type checkers, like `mypy`, think the whole project is untyped. And also added the `Typed` classifier to make it searchable in `pypi` 🙂 great work by the way 🚀

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,9 @@ setup(
     ],
     include_package_data=True,
     package_data={
-        "detect_secrets": ["py.typed"],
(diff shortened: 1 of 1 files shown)


gruebel:
@jpdakran fixed the lint issues. Didn't expect you are checking the `setup.py` 😄

jpdakran (approved):
LGTM 👍 Thanks for your contribution.

PR #607, opened 2022-08-28
Add feature to scan line

karankohli13:


lorenzodb1:
Hi @karankohli13, thank you for your contribution! Could you please expand on what issue you're trying to solve with this PR?

karankohli13:
> Hi @karankohli13, thank you for your contribution! Could you please expand on what issue you're trying to solve with this PR?

Hi, this function is useful when using the python library in third party apps, sometimes the use case of scanning a particular line/string for secrets comes up and this function creates a wrapper for that purpose.

0x123456789:
I faced the same needs, please add something to scan from general string without reading file. Example is some data which comes from HTTP response from server

lorenzodb1:
@0x123456789 unfortunately it appears that @karankohli13 has abandoned this PR. If you need this, feel free to pick it up so we can run CI and merge it!

Issue #588, opened 2022-07-28
Json output format returns an array


ArisBee:
That's minor, but when I execute something like:

```
detect-secrets audit .secrets.baseline --json --report
```

It will typically returns json objects within an array, for instance if there is no finding it will simply output `[]`. Couldn't it instead return a json object in between curly brackets `{}`?

jpdakran:
Hello @ArisBee. Is there a reason you want this enhancement? The functionality of the audit report is to display a list of secret values which is why it returns an empty list.

ArisBee:
That makes it easier for automated parsing. Now I have to define an exception when nothing is found.

lorenzodb1:
As I mentioned internally, I fully support this enhancement 😄

jpdakran:
@ArisBee Do you want it only in the nothing found case Ex. `{}`. Or just in general?

lorenzodb1:
I would add it in general for consistency. Something like
```
{
    results: [ ... ]
}
```

PR #619, opened 2022-09-22
#588: Update Report Output for Easier Parsing

jpdakran:
Problem:
- It is difficult to parse things as a list versus object

Solution:
- Update report output to be a object instead of list with the following format

```
{
    results: [ ... ]
}
```

### DIFF
--- a/detect_secrets/audit/report.py
+++ b/detect_secrets/audit/report.py
@@ -28,7 +28,7 @@ def generate_report(
     baseline_file: str,
     class_to_print: SecretClassToPrint = None,
     line_getter_factory: Callable[[str], 'LineGetter'] = open_file,
-) -> Dict[str, List[Any]]:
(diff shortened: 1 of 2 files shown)


detect_secrets/audit/report.py
```diff
@@ -28,7 +28,7 @@ def generate_report(
     baseline_file: str,
     class_to_print: SecretClassToPrint = None,
     line_getter_factory: Callable[[str], 'LineGetter'] = open_file,
-) -> List[Dict[str, Any]]:
+) -> Dict[str, List[Any]]:
```
lorenzodb1, line 31: I wonder if we can change it to `Dict[str, List[Dict[str, Any]]]` to maintain the typing benefits

jpdakran: Yea, nice!

lorenzodb1 (approved):

Issue #611, opened 2022-09-09
audit fails with custom plugin


steamraven:
When running detect-secrets audit against a baseline created with a custom filter, I get a python error

```
@ detect-secrets  audit  .secrets.baseline
Traceback (most recent call last):
  File "/home/matthewh/.local/lib/python3.10/site-packages/detect_secrets/core/plugins/initialize.py", line 19, in from_secret_type
    plugin_type = get_mapping_from_secret_type_to_class()[secret_type]
KeyError: 'IP Addresses'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/matthewh/.local/bin/detect-secrets", line 8, in <module>
    sys.exit(main())
  File "/home/matthewh/.local/lib/python3.10/site-packages/detect_secrets/main.py", line 32, in main
    handle_audit_action(args)
  File "/home/matthewh/.local/lib/python3.10/site-packages/detect_secrets/main.py", line 152, in handle_audit_action
    audit.audit_baseline(args.filename[0])
  File "/home/matthewh/.local/lib/python3.10/site-packages/detect_secrets/audit/audit.py", line 25, in audit_baseline
    if _classify_secrets(get_secret_iterator(secrets)):
  File "/home/matthewh/.local/lib/python3.10/site-packages/detect_secrets/audit/audit.py", line 40, in _classify_secrets
    secret.secret_value = get_raw_secret_from_file(secret)
  File "/home/matthewh/.local/lib/python3.10/site-packages/detect_secrets/audit/common.py", line 59, in get_raw_secret_from_file
    for item in get_raw_secrets_from_file(secret, line_getter_factory):
  File "/home/matthewh/.local/lib/python3.10/site-packages/detect_secrets/audit/common.py", line 78, in get_raw_secrets_from_file
    plugin = cast(BasePlugin, plugins.initialize.from_secret_type(secret.type))
  File "/home/matthewh/.local/lib/python3.10/site-packages/detect_secrets/core/plugins/initialize.py", line 21, in from_secret_type
    raise TypeError
TypeError
Return 1
```
```
$ detect-secrets  --version
1.3.0
```

Relevant sections of .secrets-baseline 
```
{
  "version": "1.3.0",
  "plugins_used": [
...
    {
      "name": "IPAddressDetector",
      "path": "file:///home/matthewh/no-snap/mvfiler_body/ip_detector.py"
    },
...
  ],
  "results": {
...
    "OPTIONS.txt": [
      {
[...]

PR #620, opened 2022-09-22
Fix audit report that verifies a secret

jpdakran:
Problem:
- When running an audit report on a baseline file - we reverse engineer the detect-secrets starting from secrets hashes, filenames, and how it was detected to the actual secret itself. 
- After a recent change - we fixed a bug with respect to detect-secrets not persisting the verify result
- As part of this - we were required to pass context in to the process where we analyze a line of code
- Since we are reverse engineering - we must also update the audit to include the context as part of the analyze_line

Solution:
- Add context (code_snippet) when we analyze a line in the audit process.

### DIFF
--- a/detect_secrets/audit/common.py
+++ b/detect_secrets/audit/common.py
@@ -19,6 +19,7 @@ from ..plugins.base import BasePlugin
 from ..transformers import get_transformed_file
 from ..types import NamedIO
 from ..util.inject import call_function_with_arguments
+from detect_secrets.util.code_snippet import get_code_snippet
 
 
 def get_baseline_from_file(filename: str) -> SecretsCollection:
@@ -91,6 +92,7 @@ def get_raw_secrets_from_file(
             line_numbers = list(range(len(lines_to_scan)))
 
         for line_number, line in zip(line_numbers, lines_to_scan):
+            context = get_code_snippet(lines=lines_to_scan, line_number=line_number)
             identified_secrets = call_function_with_arguments(
                 plugin.analyze_line,
                 filename=secret.filename,
@@ -100,6 +102,7 @@ def get_raw_secrets_from_file(
                 # We enable eager search, because we *know* there's a secret here -- the baseline
                 # flagged it after all.
                 enable_eager_search=bool(secret.line_number),
+                context=context,
             )
 
             for identified_secret in (identified_secrets or []):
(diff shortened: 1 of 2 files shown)


lorenzodb1:
So just to clarify, if this doesn't fix #611, what does it fix?

jpdakran:
@lorenzodb1 This was uncovered while I was investigating #611. When I was investigating #611 I had some AWS secrets in my baseline. These secrets have a verify on them (We can reach out to an external to verify if they are truly secrets). After a recent change in the scan functionality for a regular scan - we pass in our context to lines we analyze so we can verify them. 

Since we are reverse engineering a scan - we should pass the context in to analyze lines for an audit as well. This was not a known issue because context is a parameter for the analyze function with None default. So it worked fine until we hit the verify which occasionally requires a context (depending on the dependency injection). 

So when it tried to audit a AWS secret - it was receiving a None context - causing an error.

lorenzodb1 (approved):

detect_secrets/audit/common.py
```diff
@@ -91,6 +92,7 @@ def get_raw_secrets_from_file(
             line_numbers = list(range(len(lines_to_scan)))
 
         for line_number, line in zip(line_numbers, lines_to_scan):
+            context = get_code_snippet(lines=lines_to_scan, line_number=line_number)
```
abdelrahman-thafeer, line 95: I'd use `line_getter.lines` for the context here because verify method on some plugins search around the secret for other needed variables (e.g. username), and lines_to_scan can potentially be the secret line alone (if secrets.line_number yields True). 

I think -I could be wrong- we should increment line_number by one similar to the plugin.analyze_line call below? I think the value of line_number here refers to the index in the list and no the actual line number in the file.

abdelrahman-thafeer:
> @lorenzodb1 This was uncovered while I was investigating #611. When I was investigating #611 I had some AWS secrets in my baseline. These secrets have a verify on them (We can reach out to an external to verify if they are truly secrets). After a recent change in the scan functionality for a regular scan - we pass in our context to lines we analyze so we can verify them.
> 
> Since we are reverse engineering a scan - we should pass the context in to analyze lines for an audit as well. This was not a known issue because context is a parameter for the analyze function with None default. So it worked fine until we hit the verify which occasionally requires a context (depending on the dependency injection).
> 
> So when it tried to audit a AWS secret - it was receiving a None context - causing an error.

Good catch there! I'm wondering if we should add or modify a test case to better detect similar issues in the future?

detect_secrets/audit/common.py
jpdakran: Yes you are correct. Nice catch.

abdelrahman-thafeer (approved):

PR #621, opened 2022-09-22
Fixe #611: audit fails with custom plugin

steamraven:
Clear a lru_cache when creating default_settings

### DIFF
--- a/detect_secrets/settings.py
+++ b/detect_secrets/settings.py
@@ -71,6 +71,7 @@ def default_settings() -> Generator['Settings', None, None]:
             for plugin_type in get_mapping_from_secret_type_to_class().values()
         ],
     }) as settings:
+        get_mapping_from_secret_type_to_class.cache_clear()
         yield settings
 
 


detect_secrets/settings.py
```diff
@@ -71,6 +71,7 @@ def default_settings() -> Generator['Settings', None, None]:
             for plugin_type in get_mapping_from_secret_type_to_class().values()
         ],
     }) as settings:
+        get_mapping_from_secret_type_to_class.cache_clear()
```
lorenzodb1, line 74: How much does the performance get affected by adding this?

steamraven: `settings.defailt_settings` is called usually only once to setup the parser or run tests, so performance impact should be minimal

steamraven:
Appears the filters.common_filter_test.py makes use of testing.pugins.register_plugin.   This contextmanager relies specifically on get_mapping_from_secret_type_to_class  having an lru_cache.  It actually inserts plugins into the cache.  These of course get purged by my code.

I would say this is a problem with the test framework.  I will try to find a solution

lorenzodb1:
@steamraven have you manually tested your repo besides running the test suite we have?

steamraven:
I have manually tested with the minimal plugin and baseline that I have and ran pytest with  "time python -m pytest tests" .  I have not been able to do a full test with tox due to environment setup issues.  

It appears I am failing on autopep8 at the moment. 

I can try to write up some full tests to test my issue, but that will take bit.

lorenzodb1:
That's okay, I'd rather have coverage of this. Thank you for your contribution!

PR #628, opened 2022-10-20
Improve `DiscordBotTokenDetector` and its tests.

### LINKED ISSUE #627: DiscordBotTokenDetector failing to detect some Discord bot tokens
Hello! I was trying out the `DiscordBotTokenDetector` and noticed that it wasn't flagging some tokens (i.e. it's producing **false negatives**).

## Examples:

- `OTUyNED5MDk2MTMxNzc2MkEz.YjESug.UNf-1GhsIG8zWT409q2C7Bh_zWQ`
- `MTAyOTQ4MTN5OTU5MTDwMEcxNg.GSwJyi.sbaw8msOR3Wi6vPUzeIWy_P0vJbB0UuRVjH8l8`

(These example tokens are slightly fudged and I've also invalidated the unfudged versions, so there's no harm sharing them here. 😉)

### Expected Behavior:

These Discord bot tokens are flagged as such by `detect-secrets`.

### Actual Behavior:

`detect-secrets` does not recognize them as Discord bot tokens.

## Root Cause:

https://github.com/Yelp/detect-secrets/blob/0dcd54ce1bbe6417f5667dc7c807de20baf4912c/detect_secrets/plugins/discord.py#L14-L16

**This regex is a bit too restrictive**. Specifically, it only recognizes `M` or `N` as a valid first character, and it limits the following substring to `23` characters. From what I've observed recently, Discord bot tokens can sometimes begin with `O`, and can have a substring of up to length `25` following the first character.

## Proposed Solution:

At the **bare minimum**, I would suggest widening the regex as follows, to handle the specific false negatives mentioned above:

```
re.compile(r'[MNO][a-zA-Z\d_-]{23,25}\.[a-zA-Z\d_-]{6}\.[a-zA-Z\d_-]{27}')
```

(**Note:** The segment of the token matched by `[a-zA-Z\d_-]{27}` may actually be longer than `27` characters, as is the case in the second example token above. It isn't *strictly necessary* to account for this in the regex, since it's sufficient to match a substring of that segment.)

Additional test cases should also be included to properly capture the correct behavior.

### Optional Enhancements:

1. **Improving Readability:** Add the [`re.ASCII`](https://docs.python.org/3/library/re.html#re.A) flag to the regex and replace each instance of `[a-zA-Z\d_-]` with `[\w-]`.
2. **Future-Proofing:** Remove the `[MNO]` restriction to prevent false negatives if/when Discord ["runs out"](https://cdn.discordapp.com/attachments/881309496385884180/927924165254737970/7WdehGn.png) of tokens that begin with `O`.
3. **Fuzz Testing:** Generate random token values each test run, to simulate "user inputs" and ensure that they're properly detected.

Here are some examples from my own Discord bot token library to illustrate what the [regex](https://github.com/nuztalgia/botstrap/blob/c4113bc68f70fcb3dbc81637bbefda36a3d6a11c/botstrap/internal/tokens.py#L13-L14) and [token randomizer](https://github.com/nuztalgia/botstrap/blob/c4113bc68f70fcb3dbc81637bbefda36a3d6a11c/tests/conftest.py#L107-L112) could look like.

## Action Items

- [x] Create test cases to expose current incorrect behavior
- [ ] Open PR containing new test cases + "bare minimum" fix
- [ ] Ensure all CI checks are green so that the PR can be merged
- [ ] Determine if "optional enhancements" are appropriate for this project, and open a new PR if so

nuztalgia:
Fixes #627.

See the linked issue for more context. As mentioned, this is the "bare minimum" fix for the implementation. Lots of test cases have also been added and/or clarified.

### DIFF
--- a/detect_secrets/plugins/discord.py
+++ b/detect_secrets/plugins/discord.py
@@ -11,7 +11,8 @@ class DiscordBotTokenDetector(RegexBasedDetector):
     secret_type = 'Discord Bot Token'
 
     denylist = [
-        # Discord Bot Token ([M|N]XXXXXXXXXXXXXXXXXXXXXXX.XXXXXX.XXXXXXXXXXXXXXXXXXXXXXXXXXX)
+        # Discord Bot Token ([M|N|O]XXXXXXXXXXXXXXXXXXXXXXX[XX].XXXXXX.XXXXXXXXXXXXXXXXXXXXXXXXXXX)
         # Reference: https://discord.com/developers/docs/reference#authentication
-        re.compile(r'[MN][a-zA-Z\d_-]{23}\.[a-zA-Z\d_-]{6}\.[a-zA-Z\d_-]{27}'),
+        # Also see: https://github.com/Yelp/detect-secrets/issues/627
+        re.compile(r'[MNO][a-zA-Z\d_-]{23,25}\.[a-zA-Z\d_-]{6}\.[a-zA-Z\d_-]{27}'),
     ]
--- a/tests/plugins/discord_test.py
+++ b/tests/plugins/discord_test.py
@@ -22,17 +22,67 @@ class TestDiscordBotTokenDetector:
                 'MZ1yGvKTjE0rY0cV8i47CjAa.uRHQPq.Xb1Mk2nEhe-4iUcrGOuegj57zMC',
                 True,
             ),
-            # Random values to fail
+            # From https://github.com/Yelp/detect-secrets/issues/627
+            (
+                'OTUyNED5MDk2MTMxNzc2MkEz.YjESug.UNf-1GhsIG8zWT409q2C7Bh_zWQ',
+                True,
+            ),
+            (
+                'OTUyNED5MDk2MTMxNzc2MkEz.GSroKE.g2MTwve8OnUAAByz8KV_ZTV1Ipzg4o_NmQWUMs',
+                True,
+            ),
+            (
+                'MTAyOTQ4MTN5OTU5MTDwMEcxNg.GSwJyi.sbaw8msOR3Wi6vPUzeIWy_P0vJbB0UuRVjH8l8',
+                True,
+            ),
+            # Pass - token starts on the 3rd character (first segment is 24 characters)
+            (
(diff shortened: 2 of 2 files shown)


lirantal:
I've reached out to ChatGPT to explain the regex if it helps getting this PR merged quicker:

This is a regular expression used to match a specific pattern in a string. It is being passed to the `re.compile()` function, which compiles the regular expression into a pattern object that can be used for matching.

Here is a breakdown of the regular expression:

-   `r'`: This indicates that the string is a raw string, which means that backslashes are treated literally and do not need to be escaped.
    
-   `[MNO]`: This matches any single character that is either M, N, or O.
    
-   `[a-zA-Z\d_-]`: This matches any single character that is either a lowercase letter (a-z), an uppercase letter (A-Z), a digit (\d), or a hyphen or underscore (_).
    
-   `{23,25}`: This indicates that the preceding character or character group (in this case, `[a-zA-Z\d_-]`) must be repeated 23 to 25 times.
    
-   `\.`: This matches a literal period (.) character.
    
-   `[a-zA-Z\d_-]{6}\.`: This matches any sequence of 6 characters that are either a letter, a digit, a hyphen, or an underscore, followed by a period.
    
-   `[a-zA-Z\d_-]{27}`: This matches any sequence of 27 characters that are either a letter, a digit, a hyphen, or an underscore.
    

Overall, this regular expression is used to match strings that have the following form:

-   The first character is either M, N, or O.
-   The next 23 to 25 characters are either letters, digits, hyphens, or underscores.
-   The next character is a period.
-   The next 6 characters are either letters, digits, hyphens, or underscores.
-   The next character is a period.
-   The final 27 characters are either letters, digits, hyphens, or underscores.

For example, the string "Mabcdefghijklmnopqrstuvwxy.123456.abcdefghijklmnopqrstuvwxyz123" would match this regular expression.

---

The regular expression itself has been changed only to support an additional `O` character (optional) and an optional longer string of up to 25 characters in length.

---

This PR also includes more tests and is passing CI.
@syn-4ck it's a relatively small change to the regex to extend it to support other token formats. think we can land this?

PR #634, opened 2022-11-06
Improve b261af9fbd8bfcc3f412ab48f283c1970e5841c0 (#568)

clonejo:
- move try-except up the stack, so we actually have a line number and not only the column
- also catch IndexError (eg. if the file has fewer lines than secret.line_number)


clonejo:
i apologize for the formatting, i have my editor configured to run black on save.

PR #656, opened 2023-01-16
Extend keywords

nimrodkor:

- [x] All CI checks are green

Added 2 new keywords to the keyword detector that we found to be rampant:
1. stuff that ends with `_pass`
2. `token`

secrets are not being flagged

these secrets are being flagged

No

* **Other information**:


nimrodkor:
Had to remove python 3.6 as it isn't supported any longer by GHA. Now I see it is failing on finding secrets in my changes :(
Help would be appreciated, I'm sure I didn't add any secret...

PR #676, opened 2023-03-28
Calculate line number correctly in multi-line strings

nimrodkor:

- [x] All CI checks are green

This is a bugfix for a case where a multi-line yaml string has secrets in a line which is not the very first one.

Today, the error is pointed at the beginning of the string, even if the secret itself is in some row down the file.

It will point to the actual line with the secret in it

No

* **Other information**:
Already working on our (bridgecrewio) fork :)

### DIFF
--- a/detect_secrets/core/scan.py
+++ b/detect_secrets/core/scan.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+from functools import lru_cache
 from typing import Any
 from typing import cast
 from typing import Generator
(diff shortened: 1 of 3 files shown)


jpdakran:
Hi. Please merge master to your branch to re-run CI.

detect_secrets/core/scan.py
```diff
@@ -327,25 +340,33 @@ def _process_line_based_plugins(
...
+            ):
+                if not _is_filtered_out(
+                        required_filter_parameters=['context'],
+                        filename=secret.filename,
+                        secret=secret.secret_value,
+                        plugin=plugin,
+                        line=line,
+                        context=code_snippet,
+                ):
+                    if determine_file_type(filename) == FileType.YAML and secret.secret_value:
```
jpdakran, line 359: Have you looked at [yaml transformer](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/transformers/yaml.py)? Yaml files are transformed according to this transformer. So any changes should be done here.

nimrodkor: Hi @jpdakran !
Yes I have, it seems it merges the lines into 1 which makes sense from scanning perspective, but not from finding perspective. Otherwise, the secret's line is the first one, even though the secret is on the third line.
The yaml transformer doesn't retain a mapping of this...

lorenzodb1:
hi @nimrodkor 😄 could you please merge master into your branch?

PR #692, opened 2023-05-16
Add a plugin for IPAddress (ipv4)

perryzjc:
- [X] Tests for the changes have been added

- [X] Docs have been added / updated

- [X] All CI checks are green

This PR introduces a new feature in the form of an IP address detector. This detector scans for public IP addresses (IPv4) and ignores certain non-public IPv4 addresses. The ignored addresses include 127.x.x.x, 10.x.x.x, 172.16-31.x.x and 192.168.x.x.

Currently, there is no specific functionality in our system that allows for the detection of any IP addresses.
With this feature change, the system can now detect public IPv4 addresses. It uses a regex-based detector to identify valid public IPv4 addresses and ignore certain non-public ones.

This PR does not introduce a breaking change. The new detector class IPAddressDetector is a standalone addition and does not alter or interfere with the existing code base.

* **Other information**:
In addition to the implementation of the IP Address Detector, this PR also includes comprehensive tests to ensure the correct functionality of the detector. It tests the detector against valid public and non-public IPv4 addresses, as well as invalid IP addresses

The reference for the definition of public and non-public IPv4 addresses can be found in the following links:
[IPv4 Address Space](https://www.iana.org/assignments/ipv4-address-space/ipv4-address-space.xhtml)
[Private Network - Wikipedia](https://en.wikipedia.org/wiki/Private_network)

The code documentation has been updated to reflect these changes. The new feature is expected to enhance the system's capabilities in handling IP address-related data.


riverma:
FYI @KevinHock - thoughts?

lorenzodb1:
Hi @perryzjc, thank you for opening this PR. I noticed our PR checks did not run, so I'll have to figure what happened there first. Rest assured I'll get to this as soon as possible.

lorenzodb1:
@perryzjc hi again 😄 could you please merge master into your branch?

perryzjc:
> @perryzjc hi again 😄 could you please merge master into your branch?

@lorenzodb1 hi again, I've merged the latest updates from the master branch into my feature branch as requested. Please let me know if there's anything else needed for this pull request. Thanks!

lorenzodb1:
@perryzjc looks like there's some pre-commit checks failing, please take a look

perryzjc:
> @perryzjc looks like there's some pre-commit checks failing, please take a look

@lorenzodb1  Thanks for letting me know. I've resolved the issues and all checks are now passing in the workflow.

tests/plugins/ip_public_test.py
```diff
@@ -0,0 +1,54 @@
...
+                # Valid IPv4 addresses, Non-public
+                ('127.0.0.1', False),
+                ('10.0.0.1', False),
+                ('172.16.0.1', False),
+                ('192.168.0.1', False),
+                # Invalid IPv4 addresses
+                ('256.256.256.256', False),
+                ('1.2.3', False),
+                ('1.2.3.4.5.6', True),
+                ('1.2.3.4.5.6.7.8', '2'),
```
lorenzodb1, line 44: These two aren't valid IPv4 addresses, so they shouldn't get flagged (similarly to `1.2.3`)

tests/plugins/ip_public_test.py
```diff
@@ -0,0 +1,54 @@
...
+                ('1.2.3.04', True),
```
lorenzodb1, line 45: This one is also an invalid IPv4 address, but differently from the two cases above it follows the generic format correctly. This case (I think) should return true, but I'd like to ask for @jpdakran for his opinion on this

tests/plugins/ip_public_test.py
perryzjc: Hi @lorenzodb1 , thank you for pointing it out. I agree with this and just resolved the issue.

tests/plugins/ip_public_test.py
lorenzodb1: @perryzjc after thinking about this a bit more, I think an invalid IP should always return `False`, therefore this case should return `False` too (this in an attempt to avoid false positives)

perryzjc: > @perryzjc after thinking about this a bit more, I think an invalid IP should always return `False`, therefore this case should return `False` too (this in an attempt to avoid false positives)

@lorenzodb1 I feel it makes sense. Just adjusted the test case and the implementation.

riverma:
Hey @lorenzodb1 @jpdakran - just wanted to follow-up here. It would be great to have these plugins part of the main codebase when you get a chance! Great job @perryzjc.

lorenzodb1:
Hi @riverma, I have a bit of a backlog, but I promise you I'll get to this (and other great PRs) as soon as possible.

riverma:
> Hi @riverma, I have a bit of a backlog, but I promise you I'll get to this (and other great PRs) as soon as possible.

Thank you @lorenzodb1 for your support!

lorenzodb1 (approved):
@perryzjc thank you for your contribution! These changes will be included in the next release coming up at the end of this week 😄

Issue #734, opened 2023-11-20
UI improvement: Wording of audit prompt is confusing


direvus:
* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request

* **What is the current behavior?**

When running a `detect-secrets audit`, the tool displays the following prompt:

> "Is this a secret that should be committed to this repository? (y)es, (n)o, (s)kip, (q)uit:"

This prompt is worded in a way that only makes sense if the detection *is actually a secret*. The wording doesn't make sense when the detection is not a secret (false positive).

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Not a bug.

* **What is the expected behavior?**

The prompt should be worded to make sense for false positives, as well as for genuine detections.

* **What is the motivation / use case for changing the behavior?**

False positives are common -- 100% of detections have been false positives so far in my experience. The prompt wording should make sense in the case of false positives.

If the user is taking the question in the prompt literally, what is the correct response for a false positive? Should we answer (y)es because it should be committed, or (n)o because it's not a secret? It's unclear.

I would suggest an alternative and simpler wording that works in both cases, such as:

> Should this item be committed to this repository?

* **Please tell us about your environment:**
  - detect-secrets Version: 1.4.0
  - Python Version: 3.11.4
  - OS Version: MacOS Sonoma 14.1.1
  - File type (if applicable):

* **Other information**

direvus:
As further supporting evidence, the code in https://github.com/Yelp/detect-secrets/blob/81da50d60c428ed3d95a25a2405c190238735e26/detect_secrets/audit/audit.py#L81 already agrees with my proposed wording :D

lorenzodb1:
Hi @direvus, thank you for raising this issue! Do you think you can turn this into a PR where we can discuss a more appropriate wording for that question?

direvus:
Yeah, no worries.

PR #738, opened 2023-11-20
Audit prompt wording

direvus:

- [x] All CI checks are green

Modifies the wording of the audit prompt, so that it makes sense for false positives as well as genuine detections.

This was initially raised as issue #734 

The current audit prompt is confusing for false positives.

The prompt now omits any reference to the detection being a secret (since we don't know whether it is or isn't at this point)

No

* **Other information**:

### DIFF
--- a/detect_secrets/audit/io.py
+++ b/detect_secrets/audit/io.py
@@ -130,7 +130,7 @@ class UserPrompt:
 
     def __str__(self) -> str:
         if 'Y' in self.valid_input:
-            output = 'Should this be committed to the repository?'
+            output = 'Should this string be committed to the repository?'
         else:
             output = 'What would you like to do?'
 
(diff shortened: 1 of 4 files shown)


detect_secrets/audit/io.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, allow_labelling: bool, allow_backstep: bool) -> None:
 
     def __str__(self) -> str:
         if 'Y' in self.valid_input:
-            output = 'Is this a secret that should be committed to this repository?'
+            output = 'Should this be committed to the repository?'
```
lorenzodb1, line 133: Could we change it to the following (in all occurrences)?
```
Should this string be committed to the repository?
```

direvus: Sure, that's fine by me. I'll update the branch.

lorenzodb1 (approved):

PR #774, opened 2023-12-22
fix not being able to scan from subdirectories

Ferada:
- [ ] Docs have been added / updated (which one?)
- [x] All CI checks are green

bug fix

When running `detect-secrets scan` from a subdirectory of the git root, no results will be returned (unless `--all-files` was specified). This doesn't seem to be intentional, as the path handling implies that the current directory may well be in some other place.

Compare `detect-secrets scan docs` in this repository with `cd docs && detect-secrets scan`, the former returns hits for `docs/design.md` and `docs/filters.md`, the latter doesn't.

Why is it desirable to have these behave the same? Well for example you might have a monorepo and just scan a subdirectory. When `detect-secrets` is being called from another tool, you might not necessarily be aware where the git root is and would have to first navigate there, then pass the path to the subdirectory explicitly, instead of just running `detect-secrets scan` from where you're standing without any further action.

Not a new feature, you'll just be able to run `detect-secrets scan` from a subdirectory and things will still work.

Unless someone accidentally relied on no results being returned, this is not a breaking change and might surface some secrets that were previously ignored if the program was called like described above.


lorenzodb1 (dismissed):

Ferada:
@lorenzodb1 hi, since I've got you here, any thoughts on the problem statement and / or my attempt at fixing it? I've only ran the test cases on Linux, if the failed run there on Mac OS is real, I'll see if I can find a machine to replicate it.

lorenzodb1:
@Ferada so are you saying the test passes on Linux but fails on MacOS?

Ferada:
> @Ferada so are you saying the test passes on Linux but fails on MacOS?

@lorenzodb1 I was saying that the GitHub actions output here has two failed runs and several cancelled ones and since I can't restart them I couldn't tell if they were real. However I found a machine to test it on and added one more change to fix up a path problem that, yes, only happens on Mac. I get zero test failures now.

Issue #715, opened 2023-10-23
Custom Filters cannot be passed in from the command line


a-weiss-programmer:
* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request

* **What is the current behavior?**
Adding a custom filter by file when doing a command line scan is not working. When trying to pass a custom filter as described in the document, a command line usage error comes back saying "detect-secrets scan: error: argument -f/--filter: Did not specify function name for imported file."

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Install detect-secrets
2. Create a custom_filter.py script with the given filter in [the documentation](https://github.com/Yelp/detect-secrets/blob/master/docs/filters.md#using-your-own-filters)
3. Invoke the command below the example script provided [in the filter documentation](https://github.com/Yelp/detect-secrets/blob/master/docs/filters.md#using-your-own-filters)
![9ozQvsEVtE](https://github.com/Yelp/detect-secrets/assets/17444278/3fa9787e-dc8f-4a41-9da1-762bcc702f4a)

* **What is the expected behavior?**
The expected behavior is that the filter would be picked up properly. This functionality works when invoking detect secrets from a separate python program.

* **What is the motivation / use case for changing the behavior?**
Allows for proper usage of the command line behavior provided

* **Please tell us about your environment:**
  - detect-secrets Version: 1.4.30
  - Python Version: 3.11.5
  - OS Version: Ubuntu 22.04.3 LTS (Windows Subsystem for Linux)
  - File type (if applicable):

* **Other information**
This is most likely an issue with urllib parsing of the file in the [filters usage file](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/core/usage/filters.py), as urlparse returns a ParseResult where the netloc property is populated with the filepath instead of the path property, since the scheme is file based.

![WindowsTerminal_eZtkAdZvR6](https://github.com/Yelp/detect-secrets/assets/17444278/66a747cc-218f-4504-b245-e0c7a89f4853)

PR #803, opened 2024-04-04
Check netloc in urlparse result if path is empty

a-weiss-programmer:

- [x] All CI checks are green

This fixes a bug where certain versions of urllib will put the results of a parsed file (e.g. file://testing/custom_filters.py::is_invalid_secret) into the netloc property of urlparse instead of the path property.

https://github.com/Yelp/detect-secrets/issues/715

The custom filter parser will now check netloc if path is not populated

Users should have to make no changes, as it prefers path over netloc

* **Other information**:


lorenzodb1:
@a-weiss-programmer thank you for your contribution! Could you advice me on which version have one or the other behaviour?

Issue #768, opened 2023-12-13
Detection of telegram bot API-keys


StuffyMonkey:
* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request

* **What is the current behavior?**
Library doesn't support detection of telegram bot tokens.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Not a bug.

* **What is the expected behavior?**
Library should support detection of telegram bot tokens.

* **What is the motivation / use case for changing the behavior?**
Due to popularity of creating telegram bots it would be better to add option to detect telegram bot API keys.

* **Please tell us about your environment:**
  - detect-secrets Version: 1.4.0
  - Python Version: 3.12
  - OS Version: Manjaro KDE 2023
  - File type (if applicable):

* **Other information**
add two plugins for detecting telegram bot tokens and some tests: telegram_token.py and telegram_token_test.py

PR #769, opened 2023-12-13
Added TelegramTokenDetector and tester

StuffyMonkey:
* **Please check if the PR fulfills these requirements**
- [x] Tests for the changes have been added
- [x] Docs have been added / updated
- [ ] All CI checks are green

* **What kind of change does this PR introduce?**
feature

* **What is the current behavior?**
Current version of library doesn't support detection of telegram bot API keys.

* **What is the new behavior (if this is a feature change)?**
Added two plugins to support detection of telegram bot API keys.

* **Does this PR introduce a breaking change?**
No breaking changes.

* **Other information**:

PR #808, opened 2024-04-29
GH-768: Plugin to detect Telegram bot tokens

Chandra158:
## Please check if the PR fulfills these requirements

- [x] All CI checks are green

## What kind of change does this PR introduce?

- Feature: new plugin for Telegram bot tokens
- GH-768

## Tests
- New test added for plugin
- Tested manually in local with newly created token
```sh
{
        "type": "Telegram Bot Token",
        "filename": "test_data/test-bot.yaml",
        "hashed_secret": "c429079ee6a839ece2e7b0a6770198d8fe5ed438",
        "is_verified": true,
        "line_number": 11
},
```

### DIFF
--- a/detect_secrets/plugins/telegram_token.py
+++ b/detect_secrets/plugins/telegram_token.py
@@ -1,0 +1,31 @@
+"""
+This plugin searches for Telegram bot tokens
+"""
+import re
+
+import requests
+
+from ..constants import VerifiedResult
+from detect_secrets.plugins.base import RegexBasedDetector
+
+
+class TelegramBotTokenDetector(RegexBasedDetector):
+    """Scans for Telegram bot tokens."""
+    secret_type = 'Telegram Bot Token'
+
+    denylist = [
+        # refs https://core.telegram.org/bots/api#authorizing-your-bot
+        re.compile(r'\d{8,10}:[0-9A-Za-z_-]{35}'),
+    ]
+
+    def verify(self, secret: str) -> VerifiedResult:  # pragma: no cover
+        response = requests.get(
+            'https://api.telegram.org/bot{}/getMe'.format(
+                secret,
+            ),
+        )
+        return (
+            VerifiedResult.VERIFIED_TRUE
+            if response.status_code == 200
+            else VerifiedResult.VERIFIED_FALSE
+        )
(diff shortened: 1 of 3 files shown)


detect_secrets/plugins/telegram_token.py
```diff
@@ -0,0 +1,31 @@
...
+class TelegramBotTokenDetector(RegexBasedDetector):
+    """Scans for Telegram bot tokens."""
+    secret_type = 'Telegram Bot Token'
+
+    denylist = [
+        # refs https://core.telegram.org/bots/api#authorizing-your-bot
+        re.compile(r'\d{8,10}:[0-9A-Za-z_-]{35}'),
+    ]
+
+    def verify(self, secret: str) -> VerifiedResult:  # pragma: no cover
```
lorenzodb1, line 21: Looking at other `RegexBasedDetector, it doesn't look like you need a `verify` method like you would in a `BasePlugin` (see `github_token.py` for reference).

lorenzodb1 (commented):
Is this the same as https://github.com/Yelp/detect-secrets/pull/769?

detect_secrets/plugins/telegram_token.py
Chandra158: [plugins.md#verified-secrets](https://github.com/Yelp/detect-secrets/blob/master/docs/plugins.md#verified-secrets)
> provides the ability to decrease false positives by making an API call to a server to check whether a PotentialSecret is indeed real. This can be extremely useful for specific secrets (e.g. RegexBasedDetector subclasses) since we are able to determine which server to contact.


`verify` method is optional but really good in decreasing `False-positives`. Others `RegexBasedDetector` plugins also has this method implemented where there is a way to verify whether the token is still active. ([stripe.py](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/stripe.py), [slack.py](https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/slack.py))

In my test in local:
- without verify: will report if any string matches the plugin regex ("is_verified": false) 
- with verify: will only report if the token is still active ("is_verified": true)

lorenzodb1:
@Chandra158 I don't see any update on that PR, so I'd go with this one.

jpdakran (approved):
LGTM

Killerbye (approved):

PR #819, opened 2024-05-01
Plugin for PyPI api tokens

Chandra158:
## Please check if the PR fulfills these requirements

- [x] All CI checks are green

## What kind of change does this PR introduce?

- Feature: new plugin for PyPI tokens
- [Token format](https://warehouse.pypa.io/development/token-scanning.html): 
> PyPI token is bound to start with:
> pypi-AgEIcHlwaS5vcmc[A-Za-z0-9-_]{70,}
> A token can be arbitrary long because we may add arbitrary many caveats.

## Tests
- New test added for plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ KeywordDetector
 MailchimpDetector
 NpmDetector
 PrivateKeyDetector
+PypiTokenDetector
 SendGridDetector
 SlackDetector
 SoftlayerDetector
--- a/detect_secrets/plugins/pypi_token.py
+++ b/detect_secrets/plugins/pypi_token.py
@@ -1,0 +1,16 @@
+"""
+This plugin searches for PyPI tokens
+"""
+import re
+
+from detect_secrets.plugins.base import RegexBasedDetector
+
+
+class PypiTokenDetector(RegexBasedDetector):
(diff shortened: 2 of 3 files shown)


lorenzodb1 (commented):
Two questions:
1) does the TestPypi token have the same format?
2) is there a way to check the validity of the token?

Chandra158:
> 1. does the TestPypi token have the same format?

TestPypi has different format: `pypi-AgENdGVzdC5weXBpLm9yZw[A-Za-z0-9-_]{70,}`

> 2. is there a way to check the validity of the token?

This is not supported yet. ([refs](https://github.com/pypi/warehouse/issues/6407))
There're some hacky way to do it but it'd also require other details like `username` in addition to the token.

lorenzodb1:
Could you add checks for Test Pypi too?

>There're some hacky way to do it but it'd also require other details like `username` in addition to the token.

Do you think is something we could add as optional? Maybe you could add it as a configurable option for the user in the baseline file.

Chandra158:
Added a new commit for the test.pypi tokens. 

> Do you think is something we could add as optional? 

IMO, there's no way to validate yet.

lorenzodb1 (approved):

Issue #857, opened 2024-06-20
Bypass any plugin with square brackets


tsigouris007:
* **I'm submitting a ...**
  - [x] bug report

**Current behavior**

A user is able to bypass the tool engine by using `[]` characters. This is also a security issue.

**Steps to reproduce**

To reproduce you can use the following `config.rb` file as is (the secrets are dummy / generated values):
```yaml
     1	client_1:
     2	  region: "eu-central-1"
     3	  access_key_id: "<%= ENV['CLIENT_ACCESS_KEY_ID'].presence || 'AKIA123456789ABCDEF1' %>"
     4	  secret_key: "<%= ENV['CLIENT_SECRET_KEY'].presence || 'izxN/tDSMytE0ZsmcG6JOts27fd1cIkkmzimFiL1' %>"
     5	client_2:
     6	  region: "eu-central-1"
     7	  access_key_id: "AKIA123456789ABCDEF2"
     8	  secret_key: "izxN/tDSMytE0ZsmcG6JOts27fd1cIkkmzimFiL2"
     9	client_3:
    10	  region: "eu-central-1"
    11	  access_key_id: "<%= ENV'CLIENT_ACCESS_KEY_ID'].presence || 'AKIA123456789ABCDEF3' %>"
    12	  secret_key: "<%= ENV'CLIENT_SECRET_KEY'].presence || 'izxN/tDSMytE0ZsmcG6JOts27fd1cIkkmzimFiL3' %>"
```
Or raw without the line numbers:
```yaml
client_1:
  region: "eu-central-1"
  access_key_id: "<%= ENV['CLIENT_ACCESS_KEY_ID'].presence || 'AKIA123456789ABCDEF1' %>"
  secret_key: "<%= ENV['CLIENT_SECRET_KEY'].presence || 'izxN/tDSMytE0ZsmcG6JOts27fd1cIkkmzimFiL1' %>"
client_2:
  region: "eu-central-1"
  access_key_id: "AKIA123456789ABCDEF2"
  secret_key: "izxN/tDSMytE0ZsmcG6JOts27fd1cIkkmzimFiL2"
client_3:
  region: "eu-central-1"
  access_key_id: "<%= ENV'CLIENT_ACCESS_KEY_ID'].presence || 'AKIA123456789ABCDEF3' %>"
[...]

PR #860, opened 2024-06-20
Fix check bypass via square brackets Issue #857

tsigouris007:

- [ ] All CI checks are green

Fixes behavior on missed secrets that have square brackets before their default value.
Example:
```
"<%= ENV['CLIENT_ACCESS_KEY_ID'].presence || 'AKIA123456789ABCDEF1' %>"
```

Issue: https://github.com/Yelp/detect-secrets/issues/857

Escapes the square brackets and catches secrets properly.

No breaking changes.

### DIFF
--- a/detect_secrets/core/scan.py
+++ b/detect_secrets/core/scan.py
@@ -316,7 +316,7 @@ def _process_line_based_plugins(
     # filters return True.
     for line_number, line in lines:
         log.debug(f'Processing {filename}:{line_number}')
-        line = line.rstrip()
+        line = line.rstrip().replace("[", "\[").replace("]", "\]")
         code_snippet = get_code_snippet(
             lines=line_content,
             line_number=line_number,


lorenzodb1 (commented):
Hi @tsigouris007, thank you for your PR!

Could you explain what the bug was?

tsigouris007:
> Hi @tsigouris007, thank you for your PR!
> 
> 
> 
> Could you explain what the bug was?

Hi,
You can visit the issue for the compete write-up.

lorenzodb1:
Hi @tsigouris007, #857 describes the issue, I was looking for an explanation around what's causing said issue in the code

tsigouris007:
> Hi @tsigouris007, #857 describes the issue, I was looking for an explanation around what's causing said issue in the code

I haven't quite had the time to find the exact location that brakes this.
The combined opening/closing square brackets `[]` seem to be causing the lines to not be fetched at all for checking.

For example I got `AWSKeyDetector`'s code into a separate standalone python file and the following secret was caught properly by the regex:
```
access_key_id: "<%= ENV['CLIENT_ACCESS_KEY_ID'].presence || 'AKIA123456789ABCDEF1' %>"
```

I am taking wild a guess here and I most probably think the problem is how the checked file lines are fetched (and that's why I thought of "escaping" the square brackets earlier).
As far as I can help atm you can go to `detect_secrets/core/scan.py` method `_process_line_based_plugins` and the following condition is returning `True` and is quitting early:
```python
if _is_filtered_out(
            required_filter_parameters=['line'],
            filename=filename,
            line=line,
            context=code_snippet,
):
```
If you remove it or add a `pass` instead of `continue` these secrets are caught.
I can take a closer look but it will take me at least a couple of days.

tsigouris007:
> Understanding what exactly is causing the bug will help us finding a good solution moving forward rather than a workaround that could potentially add another set of issues, hence why I'm asking this. Take your time and please let me know what you'll find!

It's understandable.
I 'll try to get back to this asap.

tsigouris007:
Hi @lorenzodb1 ,
so I found where the problem occurs.
In file `detect_secrets/filters/heuristic.py` there is `_get_indirect_reference_regex` with a regex of:
```python
return re.compile(r'([^\v=!:]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
```
So the choice is to either change this regex a bit or escape earlier the square brackets.
The latter could be done as is in my PR or we could do it also in `is_indirect_reference` function inside the `heuristic.py` file right before the boolean return as shown:
```python
def is_indirect_reference(line: str) -> bool:
    """
    Filters secrets that take the form of:

        secret = get_secret_key()

    or

    """
    # Constrain line length as the heuristic's intention is to target lines that resemble
    # function calls. The constraint avoids catastrophic backtracking failures of the regex.
    if len(line) > 1000:
        return False
    line = line.replace('[', '\[').replace(']', '\]')
    return bool(_get_indirect_reference_regex().search(line))
```
This produces the correct results.

After playing around with the regex I came up with the following that seems to be working (tested on a huge repo with ~1.2k secrets):
```python
def _get_indirect_reference_regex() -> Pattern:
    # Regex details:
    #   ([^\v=!:]*)     ->  Something before the assignment or comparison
    #   \s*             ->  Some optional whitespaces
    #   (:=?|[!=]{1,3}) ->  Assignment or comparison: :=, =, ==, ===, !=, !==
    #   \s*             ->  Some optional whitespaces
    #   (
    #       [\w.-]+     ->  Some alphanumeric character, dot or -
    #       [\[\(]      ->  Start of indirect reference: [ or (
    #       [^\v]*      ->  Something except line breaks
    #       [\]\)]      ->  End of indirect reference: ] or )
    #   )
    # return re.compile(r'([^\v=!:]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])') # Left the old one for reference
    return re.compile(r'([^\v=!:"<%>]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
```
How do you want to go through with this?
I have no strong opinion.

lorenzodb1:
Hi @tsigouris007, sorry for the delay in replying. I personally prefer the second option. Thanks for investigating this and coming up with a solution!

tsigouris007:
> Hi @tsigouris007, sorry for the delay in replying. I personally prefer the second option. Thanks for investigating this and coming up with a solution!

Hi @lorenzodb1 ,
I have updated the PR.
Give it a test if you want, for sanity reasons.

detect_secrets/filters/heuristic.py
```diff
@@ -197,7 +197,7 @@ def _get_indirect_reference_regex() -> Pattern:
     #       [^\v]*      ->  Something except line breaks
     #       [\]\)]      ->  End of indirect reference: ] or )
     #   )
-    return re.compile(r'([^\v=!:]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
+    return re.compile(r'([^\v=!:"<%>]*)\s*(:=?|[!=]{1,3})\s*([\w.-]+[\[\(][^\v]*[\]\)])')
```
lorenzodb1, line 200: @tsigouris007 could you add a test for this change?

tsigouris007: Sorry for the delay.
I added two test cases for this change.
Are we good to go?

tests/filters/heuristic_filter_test.py
```diff
@@ -139,6 +139,8 @@ def test_is_prefixed_with_dollar_sign(secret, result):
         ('secret = get_secret_key()', True),
         ('secret = request.headers["apikey"]', True),
         ('secret = hunter2', False),
+        ("<%= ENV['CLIENT_ACCESS_KEY_ID'].presence || 'AKIA123456789ABCDEF1' %>", True),
+        ("ENV['CLIENT_ACCESS_KEY_ID'].presence || 'AKIA123456789ABCDEF1'", False),
```
lorenzodb1, line 143: Why would this case return `False`?

tsigouris007: I'm not sure I understand the question. The tool does not catch this case anyways. The first case is the one we're fixing here. The second one is not caught either way. Do you also wanna catch this one?

detect_secrets/filters/heuristic.py
lorenzodb1: Isn't that what you were trying to fix with this PR?

tsigouris007: I was catching the first case as that was my initial issue as described above.
So, I beefed up the regex and now it catches both cases (broke it in multiple lines for readability, it's not that far from the original one).
I also added a few more test cases that should be caught following this logic and added a brief explanation for each case.
Let me know if this works for you now.

tsigouris007:
Hi @lorenzodb1 ,
it's been quite some time.
Is the latest update what we're looking for?

lorenzodb1:
Hi @tsigouris007, I'm sorry for the late reply. I'm no longer involved in this project as a maintainer, but I'll tag @jpdakran so he can take a look when it's most convenient for him. Changes do lgtm anyway 👍🏼

tsigouris007:
> Hi @tsigouris007, I'm sorry for the late reply. I'm no longer involved in this project as a maintainer, but I'll tag @jpdakran so he can take a look when it's most convenient for him. Changes do lgtm anyway 👍🏼 

Thank you very much!

PR #878, opened 2024-09-03
fix(telegram_bot_token): Regex must match just bot tokens

jfagoagas:
* **Please check if the PR fulfils these requirements**

- [x] All CI checks are green
```
(venv) ➜  time python -m pytest tests/plugins/telegram_token_test.py
================================================================================ test session starts =================================================================================
platform darwin -- Python 3.11.9, pytest-7.4.3, pluggy-1.5.0
rootdir: /private/tmp/detect-secrets
plugins: xdist-3.6.1
collected 6 items

tests/plugins/telegram_token_test.py ......                                                                                                                                    [100%]

================================================================================= 6 passed in 0.05s ==================================================================================
python -m pytest tests/plugins/telegram_token_test.py  0,21s user 0,04s system 96% cpu 0,262 total
```

This PR fixes a bug with the TelegramBotToken since some AWS ARNs were matched against it due to the regex configured in the `deny_list`.

* **What is the current behaviour?**

The following AWS ARN `arn:aws:sns:aaa:111122223333:aaaaaaaaaaaaaaaaaaassssssdddddddddddd` matches against the TelegramBotToken when the `verify` option is not enabled, but it should not be needed to be verified since it is not.

This match because the regex is not enough restrictive from the beginning and the end.

* **What is the new behaviour (if this is a feature change)?**
Add the start-of-line `^` and end-of-line `$` character to reduce false positives.

No.
* **Other information**:

### DIFF
--- a/detect_secrets/plugins/telegram_token.py
+++ b/detect_secrets/plugins/telegram_token.py
@@ -15,7 +15,7 @@ class TelegramBotTokenDetector(RegexBasedDetector):
 
     denylist = [
         # refs https://core.telegram.org/bots/api#authorizing-your-bot
-        re.compile(r'\d{8,10}:[0-9A-Za-z_-]{35}'),
+        re.compile(r'^\d{8,10}:[0-9A-Za-z_-]{35}$'),
     ]
 
     def verify(self, secret: str) -> VerifiedResult:  # pragma: no cover
--- a/tests/plugins/telegram_token_test.py
+++ b/tests/plugins/telegram_token_test.py
@@ -13,6 +13,7 @@ class TestTelegramTokenDetector:
             ('7213808860:AAH1bjqpKKW3maRSPAxzIU-0v6xNuq2-NjM', True),
             ('foo:AAH1bjqpKKW3maRSPAxzIU-0v6xNuq2-NjM', False),
             ('foo', False),
+            ('arn:aws:sns:aaa:111122223333:aaaaaaaaaaaaaaaaaaassssssddddddddddddd', False)
         ],
     )
     def test_analyze(self, payload, should_flag):
--- a/tests/plugins/telegram_token_test.py
+++ b/tests/plugins/telegram_token_test.py
@@ -8,7 +8,7 @@ class TestTelegramTokenDetector:
     @pytest.mark.parametrize(
         'payload, should_flag',
         [
-            ('bot110201543:AAHdqTcvCH1vGWJxfSe1ofSAs0K5PALDsaw', True),
+            ('110201543:AAHdqTcvCH1vGWJxfSe1ofSAs0K5PALDsaw', True),
             ('110201543:AAHdqTcvCH1vGWJxfSe1ofSAs0K5PALDsaw', True),
             ('7213808860:AAH1bjqpKKW3maRSPAxzIU-0v6xNuq2-NjM', True),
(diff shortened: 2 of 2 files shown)


tests/plugins/telegram_token_test.py
```diff
@@ -8,11 +8,12 @@ class TestTelegramTokenDetector:
     @pytest.mark.parametrize(
         'payload, should_flag',
         [
-            ('bot110201543:AAHdqTcvCH1vGWJxfSe1ofSAs0K5PALDsaw', True),
+            ('110201543:AAHdqTcvCH1vGWJxfSe1ofSAs0K5PALDsaw', True),
```
jfagoagas, line 11:
```suggestion
            ('bot110201543:AAHdqTcvCH1vGWJxfSe1ofSAs0K5PALDsaw', False),
```

jfagoagas:
Pinning @lorenzodb1 @KevinHock @domanchi to see if this can get a review soon. Thanks in advance!!

detect_secrets/plugins/telegram_token.py
```diff
@@ -15,7 +15,7 @@ class TelegramBotTokenDetector(RegexBasedDetector):
 
     denylist = [
         # refs https://core.telegram.org/bots/api#authorizing-your-bot
-        re.compile(r'\d{8,10}:[0-9A-Za-z_-]{35}'),
+        re.compile(r'^\d{8,10}:[0-9A-Za-z_-]{35}$'),
```
lorenzodb1, line 18: Reading the Telegram docs, I found that

>The token looks something like 123456:ABC-DEF1234ghIkl-zyx57W2v1u123ew11

which I assume means there's a specific format in terms of which characters will be upper case and which will be lower case. If that's correct, would you mind making the regex more accurate?

jfagoagas: Hello @lorenzodb1 I'm not sure if that's accurate since they also states this in their documentation

> The token is a string, like 110201543:AAHdqTcvCH1vGWJxfSeofSAs0K5PALDsaw, which is required to authorize the bot and send requests to the Bot API. Keep your token secure and store it safely, it can be used by anyone to control your bot.

https://core.telegram.org/bots/features#creating-a-new-bot

jfagoagas: The regex configured is the same as Trufflehog's https://github.com/trufflesecurity/trufflehog/blob/57802abf5276961426482ace808057016c714958/pkg/detectors/telegrambottoken/telegrambottoken.go#L27

lorenzodb1 (approved):
Thank you for your contribution!

jfagoagas:
Hello @lorenzodb1 is there any ETA to have this included in a new release? Is something we do really need and we are trying not to vendor/fork the repository.

Thank you!

jfagoagas:
@lorenzodb1 @KevinHock @domanchi Any updates for a new release? Thanks!

lorenzodb1:
@jfagoagas we're planning to make a release soon. Either @jpdakran or myself will make one. Thank you for your patience.

jfagoagas:
@lorenzodb1 @jpdakran any updates on the release?

jfagoagas:
@lorenzodb1 @jpdakran any updates on the release? It's been almost a year since the last one.

jfagoagas:
@lorenzodb1 @jpdakran any updates on the release? It's been more than a year since the last one. Please let us know what are the plans to keep maintaining this tool.

Issue #852, opened 2024-06-09
KeywordDetector plugin doesn't detect secrets which start with a symbol


gpflaum:
* **I'm submitting a ...**
  - [X] bug report
  - [ ] feature request

* **What is the current behavior?**

The KeywordDetector plugin doesn't detect secrets which start with a symbol. The [comments in keyword.py](https://github.com/Yelp/detect-secrets/blob/5ee30bfb93128b3d74d8993c0000122b20270081/detect_secrets/plugins/keyword.py#L84) say it will "find secrets that starts with symbols or alphanumeric characters". That used to work, but a [change](https://github.com/Yelp/detect-secrets/pull/442/files#diff-e6de82b29d51ff0d50bb9a4951479046f43f6dad3d40ba27e3dd7e2a9f6fda8b) made in #442 to address a [performance problem](https://github.com/Yelp/detect-secrets/pull/418#issuecomment-811980908) broke this. After that change it only finds secrets which start with word characters (`\w`): letters, digits, or underscore.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Create a file `leading_symbol.properties`:
```properties
# KeywordDetector does not find this secret:
a_password = +eTxoi3271x
# It does find this secret:
b_password = eTxoi3271x+
```
detect-secrets and detect-secrets-hook only find the second secret, don't find the secret with a leading symbol:
```
% detect-secrets-hook leading_symbol.properties
ERROR: Potential secrets about to be committed to git repo!

Secret Type: Secret Keyword
Location:    leading_symbol.properties:4

Possible mitigations:
  - For information about putting your secrets in a safer place, please ask in
    #security
  - Mark false positives with an inline `pragma: allowlist secret`
    comment

If a secret has already been committed, visit
https://help.github.com/articles/removing-sensitive-data-from-a-repository
```

* **What is the expected behavior?**

KeywordDetector should find secrets that start with a symbol.

* **Please tell us about your environment:**
  - detect-secrets Version: 1.5
  - Python Version: 3.11.2
  - OS Version: macOS Sonoma
  - File type (if applicable): various

nandakumar649:
Hi,

Could you please update the latest on this issue. I am facing similar issue when the passwords begin with special characters, the plugin is not detecting the secret.

clintonsteiner:
https://github.com/Yelp/detect-secrets/pull/911
This pr is related - currently only scanning things within a string quotation

PR #971, opened 2026-07-01
fix(keyword): detect secrets whose value starts with a symbol

Synvoya:
## Summary

`KeywordDetector` silently misses any secret whose value starts with a non-word symbol (`-`, `+`, `@`, `!`, `$`, …). Because the responsible `SECRET` sub-pattern is shared by every keyword regex, the miss spans every file type (YAML, Terraform, Python, INI/properties, …) and both quoted and unquoted values.

## Symptom

```python
>>> from detect_secrets.plugins.keyword import KeywordDetector
>>> kd = KeywordDetector()
>>> bool(kd.analyze_line(filename='x.properties', line='password = eTxoi3271x', line_number=1))
True
>>> bool(kd.analyze_line(filename='x.properties', line='password = +eTxoi3271x', line_number=1))
False      # identical value with a leading '+' is missed
>>> bool(kd.analyze_line(filename='x.tf', line='password = "-mytokenvalue123"', line_number=1))
False      # quoted, still missed
```

## Root cause

`detect_secrets/plugins/keyword.py`:

```python
SECRET = r'(?=[^\v\'"]*)(?=\w+)[^\v\'"]*[^\v,\'"`]'
```

The `(?=\w+)` is a zero-width lookahead anchored at position 0, so it forces the *first* character of the value to be `\w` (`[A-Za-z0-9_]`). A leading non-word symbol fails it, so the value is never matched. The adjacent comment states the intent is to *"find secrets that start with symbols or alphanumeric characters"* — the opposite of what the anchor does.

Simply removing the anchor over-corrects: it reintroduces two false positives the anchor had been suppressing as a side effect — the delimiter whitespace gets captured (`password: ${link}` → secret `' ${link}'`), and `${...}` template interpolations get treated as secrets.

## Fix

Replace the single position-0 anchor with three explicit guards that isolate each concern:

```python
SECRET = r'(?!\$\{)(?=\S)(?=[^\v\'"]*\w)[^\v\'"]*[^\v,\'"`]'
```

- `(?!\$\{)` — keep excluding `${...}` interpolations.
- `(?=\S)` — the value must start with a non-whitespace character, so the leading delimiter whitespace is consumed by the surrounding `\s*` rather than captured as the secret.
- `(?=[^\v'"]*\w)` — keep the "at least one word character somewhere" guard, so values made purely of symbols/whitespace are still rejected.

This detects symbol-leading secrets while preserving every existing false-positive guard.

## Verification

- Added leading-symbol regression cases (`LEADING_SYMBOL_SECRET`) to both the `CONFIG_TEST_CASES` and `QUOTES_REQUIRED_TEST_CASES` blocks in `tests/plugins/keyword_test.py`; they fail before this change and pass after.
- Full `tests/plugins/keyword_test.py`: **360 passed**. Every existing `None`-expected guard — `${link}`, the pure-symbol `SYMBOL_SECRET`, empty quotes — still passes.
- `flake8` clean on both changed files.


gpflaum:
Fixes #852
I haven't tested the fix, but that issue describes the problem fixed by this PR.

Issue #953, opened 2026-04-01
Still alive?


guettli:
The last release was almost two years ago.

Is this project still alive?

Dear maintainers, it is 100% ok if you say "We have no longer time for this project".

If this is the case, please tell us, so that the community finds a solution.

Thank you for `detect-secrets`!

BTW, I will switch to [gitleaks](https://github.com/gitleaks/gitleaks). It is actively maintained, and has MIT license.

lorenzodb1:
Hello 👋🏼  I'm a former maintainer of this project. I'll see if I can get ahold of someone in the team I was part of and get an answer for this.

Until then: I recall a discussion around this same topic back when I was a _Yelper_ (https://github.com/Yelp/detect-secrets/issues/473) and one of the options proposed by @KevinHock was creating an agnostic organisation ([detect-secrets-oss](https://github.com/detect-secrets-oss)) that would take over the project and maintain it. Obviously, some time has passed, but personally I’d be open to returning as a maintainer in my free time, if that’s acceptable to the current project owners. Even that would still be better than the current complete inactivity.

guettli:
@lorenzodb1 I switched to git-leaks. Does it make sense to keep detect-secrets alive? If "yes", then "why?". Which features does detect-secrets have, which are missing in other tools?

lorenzodb1:
@guettli I'm not familiar with gitleaks so I can't really form an opinion on whether it's worth keeping detect-secrets alive. That being said, I don't see the existence of a "competing" project as a valid reason to stop maintaining this one.

barseghyanartur:
@lorenzodb1 

It's a very useful project. Why is it not maintained? I assume it was created for a business need. Does Yelp switched to something else?

lorenzodb1:
@barseghyanartur as I'm not longer working for Yelp, I have no idea whether they switched to something else or they're still using it.

KevinHock:
Hey @calvinli 👋🏼 hope you're doing well, are you still at Yelp? 

Thoughts on:
> Hello 👋🏼 I'm a former maintainer of this project. I'll see if I can get ahold of someone in the team I was part of and get an answer for this.

[...]

PR #987, opened 2026-07-28
Support scanning directories outside cwd

### LINKED ISSUE #986: scan --all-files <path outside cwd> silently returns an empty result set (exit 0, no diagnostic)
### Summary

When the scan target is a path outside the current working directory, `detect-secrets scan
--all-files` returns `{"results": {}}` and exits 0. No warning, no error, no diagnostic. The output
is indistinguishable from a genuinely clean scan.

The failure direction is the dangerous one: it reports **safe**.

### Reproduction

Self-contained; nothing outside these commands is needed.

```console
$ mkdir -p demo && printf 'DEEPSEEK_API_KEY=PLACEHOLDER-VALUE-NOT-A-KEY\n' > demo/a.md

# 1. Target inside cwd — found.
$ detect-secrets scan --all-files demo | jq -c .results
{"demo/a.md":[{"type":"Secret Keyword","filename":"demo/a.md","hashed_secret":"...","is_verified":false,"line_number":1}]}

# 2. Same bytes, same filename, absolute path outside cwd — NOT found.
$ T=$(mktemp -d) && cp demo/a.md "$T/a.md"
$ detect-secrets scan --all-files "$T" | jq -c .results
{}
$ echo $?
0

# 3. Same directory, but as cwd — found again.
$ cd "$T" && detect-secrets scan --all-files . | jq -c .results
{"a.md":[{"type":"Secret Keyword","filename":"a.md","hashed_secret":"...","is_verified":false,"line_number":1}]}
```

Identical bytes, identical filename, identical plugin set. **The only variable is whether the target
sits inside the working directory**, and the difference between "found" and "silently not found" is
not reported anywhere in the output or the exit code.

Version: `detect-secrets` 1.5.0, Python 3.12, Linux.

### Impact

The natural way to scan anything that is not already a checkout — git history, CI artefacts, an
extracted archive, container layers — is to materialise the content into a temporary directory and
scan that:

```python
with tempfile.TemporaryDirectory() as td:
    # ... write every blob from `git rev-list --objects --all` into td ...
    subprocess.run(["detect-secrets", "scan", "--all-files", td])
```

`tempfile.TemporaryDirectory()` returns a path under `/tmp`, outside cwd. **This pattern reports
clean on every input, always, and gives no indication that it examined nothing.**

We hit it building a git-history scanner. The first version reported clean over 206 blobs while
three deliberately planted credentials sat in the history it was scanning. It was caught only
because those credentials had been planted first, specifically to check that the scanner worked
before its results were believed. Without that step the false clean scan would have been shipped and
trusted.

### Suggested fix

Any of these removes the trap; the first two are the minimum, and the second is worth having on its
own:

1. **Fail loudly.** If the resolved target lies outside cwd and no files are collected, exit
   non-zero with a diagnostic naming the reason.
2. **Report the number of files actually scanned**, so an empty result is distinguishable from an
   empty scan. `0 files scanned` is unambiguous where `{}` is not — and a scanner that cannot say
   how much it examined cannot be audited.
3. Support absolute paths outside cwd, if the restriction is incidental rather than intentional.

**The restriction does not appear to be documented.** `detect-secrets scan --help` describes the
`path` argument only as *"Scans the entire codebase and outputs a snapshot of currently identified
secrets"*, with no mention of the working directory, and the README contains no reference to `cwd`,
the current working directory, absolute paths, or scanning outside a directory. So a user has no way
to learn the constraint short of hitting it — and hitting it produces no signal. **The silence is the
defect regardless of what the docs say**, but as things stand the behaviour is both undocumented and
unreported.

**"Found nothing" and "looked at nothing" are the two results a security tool must never conflate.**

### Related issues

Searched before filing; none of these looks like the same defect, but they are adjacent and a
maintainer may want to link them:

- **#870** (open) — same *symptom*, an empty `results` object, but its comments diagnose a Windows
  UTF-8 encoding problem, and the scan there is of the current repository rather than an outside
  path.
- **#876** (open) — nested directories *inside* a repository, scanned from within it. Different
  situation.
- **#34** (closed, 2018) and **#224** (closed, 2019) — the closest historical matches, both in this
  area and both predating the 1.0 rewrite. #224 in particular reports
  `detect-secrets scan --all-files /tmp/sometemp/*` finding only some files, and was closed after
  the reporter worked around it with `shopt globstar` rather than after a fix. **Neither addresses
  the silence**, which is the part that makes this dangerous rather than merely inconvenient.

Filing as a new issue rather than commenting on any of the above, because the closest two are closed
and pre-1.0 and the two open ones have different causes. Happy to have it merged into an existing
thread if a maintainer sees it differently.

### Workaround

Set `cwd` to the directory being scanned and pass `.` as the target:

```python
subprocess.run(["detect-secrets", "scan", "--all-files", "."], cwd=td)
```

Eljees:
- [ ] Docs have been added / updated (not applicable; this restores the documented `--all-files` behavior)
- [ ] All CI checks are green

Bug fix.

`detect-secrets scan --all-files <path>` silently collects no files when the explicit directory is outside the current working directory, producing an empty result with exit code 0. Fixes #986.

Explicit paths outside the current working directory are resolved and scanned. Files reached through symlinks that escape the explicitly requested directory remain excluded.

No.

* **Other information**:

Targeted regression tests: `python -m pytest tests/core/baseline_test.py tests/core/scan_test.py -q` (18 passed).


Eljees:
Ping - open since 28 July with no review. It makes scanning work for directories outside the current working directory; I can add more test coverage if that would help.

gpflaum:
@Eljees Yelp abandoned this project. See #953.

Eljees:
Thanks for the pointer. #953 is still unanswered and there have been no commits on main since April, so I will close this rather than leave it sitting in the queue. Happy to reopen if the project picks up again.

PR #997, opened 2026-09-15
Detect secrets in type-annotated assignments (Fixes #780)

### LINKED ISSUE #780: Secrets followed by type hint are not detected
* **I'm submitting a ...**
  - [x] bug report
  - [ ] feature request

The current algorithm is not able to detect secrets with type hint. E.g. `my_password: str = "bar"`
 is not detected.

* **What is the expected behavior?**
I expect that `detect_secrets` warns if someone tries to set a secret as a default value.

```python
def my_very_secret_function(my_password: str = "bar"):
    pass
```

MRX-72:
## Summary

The keyword plugin did not detect a secret assigned with a type annotation, for example:

```python
my_password: str = "bar"
api_key: Optional[str] = "sk-abc"
```

Both are missed on `main`, while the un-annotated `my_password = "bar"` is detected. This is common in Python and TypeScript codebases, so real secrets in annotated assignments slip through.

## Cause

For Python (and the other quotes-required file types), `FOLLOWED_BY_EQUAL_SIGNS_QUOTES_REQUIRED_REGEX` expects the assignment operator to follow the keyword directly:

```
{denylist}({closing})?{whitespace}(={1,3}|!==?){whitespace}({quote})({secret})(\4)
```

A `: str` annotation sits between the keyword and the `=`, so the pattern does not match.

## Fix

Add an optional, **non-capturing** type-annotation segment between the keyword and the operator:

```python
OPTIONAL_TYPE_ANNOTATION = r'(?:{whitespace}:{whitespace}[^\v=]+?)?'
```

- Non-capturing (`(?:...)`) so it does not shift the existing regex group numbers or the `..._TO_GROUP` mappings.
- The annotation may not contain `=` or a line break, so a match stays within one annotated assignment.

Only `FOLLOWED_BY_EQUAL_SIGNS_QUOTES_REQUIRED_REGEX` is touched (the regex used for Python, TypeScript/JS, Java, Swift, Terraform, CLS). No other regex, group index, or file type changes.

## Verification

Now detected:

```
my_password: str = "…"          -> secret found
api_key: Optional[str] = "…"     -> secret found
password:str="…"                 -> secret found (no surrounding spaces)
```

Still correctly ignored (no false positives):

```
password: str                    -> no assignment, no match
if password: return "…"          -> not an assignment, no match
```

- New parametrized cases added to `QUOTES_REQUIRED_TEST_CASES` (they run across every quotes-required file type).
- `pytest tests/plugins/` — 770 passed.
- `flake8 --max-line-length 100` clean on both files (matches `.pre-commit-config.yaml`).
- Re-scanning the repo against `.secrets.baseline` surfaces no new findings, so the self-scan hook is unaffected.

Fixes #780


MRX-72:

PR #998, opened 2026-09-16
Detect secrets assigned with the Python walrus operator

MRX-72:
## Summary

The keyword plugin already detects secrets assigned with the walrus operator (`:=`) for Go, but the `:=` regex was never added to the quotes-required group that Python (and Java, JS, Swift, Terraform, CLS) use. As a result a secret assigned in a Python walrus expression is not detected:

```python
if (password := "hunter2secret"):   # missed on main
    ...
```

while the same secret via `=` is detected. The walrus operator is valid Python 3.8+ and appears in conditionals and comprehensions.

## Fix

Add a quotes-required `:=` regex (`FOLLOWED_BY_COLON_EQUAL_SIGNS_QUOTES_REQUIRED_REGEX`) and wire it into `QUOTES_REQUIRED_DENYLIST_REGEX_TO_GROUP`. Quotes-required (rather than reusing the Go quotes-optional `:=` regex) keeps it consistent with the other patterns in that group and avoids introducing false positives on unquoted values.

## Verification

Now detected:

```
if (password := "…"):     -> secret found
api_key:="…"              -> secret found (no surrounding spaces)
while (secret := "…"):     -> secret found
```

Correctly ignored (no false positives):

```
password := unquoted      -> quotes required, no match
secret := ""              -> empty, no match
```

- New parametrized cases added to `QUOTES_REQUIRED_TEST_CASES` (they run across every quotes-required file type).
- `pytest tests/plugins/` — 762 passed.
- `flake8 --max-line-length 100` clean on both files.
- Re-scanning the repo against `.secrets.baseline` surfaces no new findings, so the self-scan hook is unaffected.

Note: this is a sibling to #997 (type-annotated assignments), a different Python assignment form the same plugin was missing. Kept separate so each is independently reviewable.


MRX-72:

PR #999, opened 2026-09-16
List py.typed explicitly in MANIFEST.in

### LINKED ISSUE #968: py.typed is in the published wheel but missing from the sdist — source installs lose PEP 561 typing
The published detect-secrets 1.5.0 wheel contains `detect_secrets/py.typed`, but a wheel built from the published sdist does not — so anyone installing from source (distros, `--no-binary`, platforms without wheels) silently loses type-checking support. (#579 added py.typed to the package, but the sdist doesn't ship it.)

Fix is one line: ensure `py.typed` is included in the sdist (MANIFEST.in or package-data).

Found by a wheel/sdist divergence census of popular PyPI packages ([wheelproof](https://github.com/sethc555/wheelproof), results/divergence.md); self-check with `wheelproof divcheck --package <name>`. Verified twice in clean python:3.12 containers with exact file-list agreement.

MRX-72:
The published 1.5.0 wheel has `detect_secrets/py.typed` (added in #579), but the sdist doesn't — confirmed by downloading `detect-secrets==1.5.0` from PyPI with `pip download --no-binary :all:` and checking the tarball contents. `package_data` in setup.py controls what the wheel build includes, but the sdist command needs the file listed in MANIFEST.in too, same as the existing `rfc.model` entry.

Added `include detect_secrets/py.typed` to MANIFEST.in and rebuilt the sdist locally to confirm the file is now included.

Fixes #968


MRX-72:

PR #1000, opened 2026-09-16
Add pnpm-lock.yaml to is_lock_file

MRX-72:
\`is_lock_file\` already recognizes \`package-lock.json\` (npm) and \`yarn.lock\` (yarn), but not \`pnpm-lock.yaml\`, which is pnpm's equivalent and a very common one in JS/TS repos. That means secrets accidentally captured in a pnpm lockfile aren't filtered the way they would be for the other two package managers.

Added it to the set and a case to \`test_is_lock_file\`.

Fixes #945


MRX-72:

